Issue #23, opened 2016-07-03
Add support for .NET Core


rianjs:
- Get rid of Hashtable usage in ANTLR parser
- Get rid of ArrayList usage in ANTLR parser

rianjs:
Normally I wouldn't merge half-completed work back into `master`, but there was too much value unrealized due to back-porting changes as I prepped the library for .NET Core. It looks like unit test runtimes will be down below 2 seconds after this work.

khalidabuhakmeh:
Any thoughts to migrating to xunit?

rianjs:
To xUnit from nUnit? I haven't thought about it.

kdaveid:
Hi Rian
Great work and thanks for taking over the project. I have the source running for a while. 

I'd like to contribute on ASP.NET Core branch. Is there one? How far are you in porting it?

Thanks and have a great day,
David

rianjs:
@kdaveid - I'd be thrilled for someone else to also work on this. I don't love managing build configurations and pipelines. This PR has some information on where the .NET Core port stands now: https://github.com/rianjs/ical.net/pull/28

The non-Core code has diverged a little bit, as I've been on a deleting/simplification spree, but I haven't written any new code that will be incompatible with .NET Core, AFAIK. I see several possibilities:

Option 1: delete all the directories underneath `net-core/..whatever../src` except for the `xproj` files. Copy and paste the directories containing the unit tests, ical, and ical collections `.cs` files. That'd get you most of the way there. After restoring the required nuget packages, it should build. I'm not wild about this possibility because Microsoft is going to back to XML-based project files rather than `project.json` and (presumably) `global.json`, so I don't have much interest in learning that way of building .NET code.

Option 2: Add another build configuration to the `csproj` files that already exist. According to the blog post linked, targeting PCL profile 259 will create a binary that runs on CoreCLR, which is effectively a port, skipping the bad tooling support and build/configuration management flux that's happening now. I'm assuming ReSharper's test runner and nunit will play nicely, because I assume that it's just a "normal" csproj-based solution. https://oren.codes/2015/07/29/targeting-net-core/

When I got up the energy, I was going to give option 2 a go. My guiding thoughts are these:
- No duplicate code bases. (What's underneath the `net-core` is (and always was) meant to be temporary. Whether it gets hoisted up, and overwrites the non-Core source, or something else... it was always temporary, meant as a "what will this take" experiment.
[...]

PR #28, opened 2016-07-09
Net core

rianjs:
Ical.Net builds for .NET Core, but unit tests are problematic to run. Deferring idiomatic xproj usage until the tooling (ReSharper, Visual Studio) improves. All .Net Core changes were back-ported to the non-Core library, so this should be easy-ish in the future (and I have no reason to believe that the unit tests won't run). Right now, though, we have two side-by-side versions of ical.net in the repo. The .Net Core version is in the `ical-core` directory. #23

Another alternative is to create a PCL targeting .NET 4.5 and PCL Profile 259, which will apparently generate binaries that will run on the CoreCLR:

https://oren.codes/2015/07/29/targeting-net-core/


rianjs:
Normally I wouldn't merge half-completed work back into `master`, but there was too much value unrealized due to back-porting changes as I prepped the library for .NET Core. It looks like unit test runtimes will be down below 2 seconds after this work.

PR #39, opened 2016-07-13
Test Serialize & Deserialize

mcshaz:
This request introduces a few new tests, some of which are failing (as per issue #45) - however the components failing were untested and IMHO fundamental to the functionality of the library. In Detail:
- AttendeeTest is simply the EventTest class replicated for attendees. The remove test currently fails - i.e. if adding and then removing an attendee (with the remove method), the attendee is not removed.
- SerializationTest has been renamed to DeserializationTest, as all these tests are reading a string to create an object model.
- The utility methods to compare 2 calendars remains in SerializationTest
- Several serialization tests are now added. All are failing, as currently very little is being serialized by the serialization engine, as per the description in issue #45. All tests are logging the serialized string, and if you look at the test output you will see why they are failing.

There are still weaknesses with the test - for example the testers computer must have the specified SystemTimeZone, but it is a start.

### DIFF
--- a/ical.NET.UnitTests/AttendeeTest.cs
+++ b/ical.NET.UnitTests/AttendeeTest.cs
@@ -11,7 +11,7 @@ namespace ical.NET.UnitTests
     [TestFixture]
     public class AttendeeTest
     {
-        private static Event EvtFactory()
+        internal static Event EvtFactory()
         {
             return new Event
             {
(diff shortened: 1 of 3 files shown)


rianjs:
Thanks, @mcshaz - This (serialization, as opposed to deserialization) is definitely an area where the unit tests are very weak. I appreciate you taking the time to add them.

rianjs:
(Oh, in the future, you may want to use example.com as the test domain for things. That's its purpose: https://www.iana.org/domains/reserved )

rianjs:
Ah, you forgot to add them to the `csproj` file. I'll fix that momentarily.

mcshaz:
Thanks for fixing and apologies.

Issue #40, opened 2016-07-13
Calendar.LoadFromStream is not thread safe


rianjs:
```
System.ArgumentException occurred
  HResult=-2147024809
  Message=An item with the same key has already been added.
  Source=mscorlib
  StackTrace:
       at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
       at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
       at Ical.Net.Utility.SerializationUtil.GetDeserializingMethods(Type targetType) in C:\dev\ical.net\ical.NET\Utility\SerializationUtil.cs:line 60
  InnerException:
```

Repro case. Code works fine if `.AsParallel()` is removed.

``` csharp
public static IEnumerable<DowntimeWithExactTimes> GetDowntimesInRange(IEnumerable<Downtime> downtimes, DateTime earliestStart, DateTime latestEnd)
{
    var sw = new Stopwatch();
    sw.Start();
    var earliestStartIcal = new CalDateTime(earliestStart, "UTC") {HasTime = true};
    var latestEndIcal = new CalDateTime(latestEnd, "UTC") {HasTime = true};
    var downtimesWithExactTimes = downtimes.AsParallel().SelectMany(downtime =>
    {
        IICalendarCollection calendar;
        using (var reader = new StringReader(downtime.IcalString ?? ""))
        {
            // Originally calendar = iCalendar.LoadFromStream(reader), but that uses a static serializer so it isn't thread safe.
            //calendar = Calendar.LoadFromStream(reader);
            calendar = Calendar.LoadFromStream(reader, new CalendarSerializer { SerializationContext = new SerializationContext() });
        }

        var thisDowntime = downtime;
        var occurrences = calendar.GetOccurrences(earliestStartIcal, latestEndIcal);
        return occurrences.Select(occurrence => new DowntimeWithExactTimes(thisDowntime, occurrence.Period.StartTime.AsUtc, occurrence.Period.EndTime.AsUtc));
    });
    var orderedDowntimes = downtimesWithExactTimes.OrderBy(downtime => downtime.Start).ThenBy(downtime => downtime.End).ToList();
    sw.Stop();
    return orderedDowntimes;
}
```

rianjs:
This is related to the caching I introduced last weekend. Crap.

rianjs:
188889e53028f76a5a2700790641192f5d60e0fe

PR #41, opened 2016-07-13
Fixed Concurrency issue around parallel deserialization #40

rianjs:
The unit test successfully reproduced the failure from https://github.com/rianjs/ical.net/issues/40 with duplicate key exceptions.

Switching to the `ConcurrentDictionary` allowed the test to pass.

### DIFF
--- a/ical.NET/Utility/SerializationUtil.cs
+++ b/ical.NET/Utility/SerializationUtil.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


ical.NET/Utility/SerializationUtil.cs
```diff
@@ -49,11 +50,11 @@ private static List<MethodInfo> GetDeserializingMethods(Type targetType)
                 .Select(t => t.targetTypeMethodInfo)
                 .ToList();
 
-            _onDeserializingMethods.Add(targetType, methodInfo);
+            _onDeserializingMethods.AddOrUpdate(targetType, methodInfo, (type, list) => methodInfo);
```
chescock, line 53: Since you're doing an unconditional update, you could just do `_onDeserializingMethods[targetType] = methodInfo;`.  

I think what you really want here is a single call to `GetOrAdd`, passing a lambda that does the calculation.  That is: 

``` csharp
return _onDeserializingMethods.AddOrUpdate.GetOrAdd(targetType, tt => tt
    .GetMethods(_bindingFlags)
    .Where(targetTypeMethodInfo => targetTypeMethodInfo
        .GetCustomAttributes(typeof(OnDeserializingAttribute), false).Any())
    .ToList());
```

Using the lambda parameter `tt` instead of the method parameter `targetType` means the lambda won't capture an environment, so the compiler can cache it in a static field.

PR #49, opened 2016-07-14
SerializationTests CalDateString bug fixed

mcshaz:
running tests on earlier (partially working serialization) versions helped tweak the tests a little including the stupidly error (on my part) in the CalDateString helper method (requiring leading 0s!).

the GetValues helper method now makes the semicolon optional, which has the advantage of a more useful test fail message (could not find property X) in the case the serialized output is:
`ATTENDEE:MAILTO:brentm@example.com`
without this change, the test fails with the very confusing `could not find an ATTENDEE brentm@example.com`

### DIFF
--- a/ical.NET.UnitTests/SerializationTests.cs
+++ b/ical.NET.UnitTests/SerializationTests.cs
@@ -132,7 +132,7 @@ namespace ical.NET.UnitTests
         //http://www.kanzaki.com/docs/ical/dateTime.html
         static string CalDateString(IDateTime cdt)
         {
-            var returnVar = cdt.Year.ToString() + cdt.Month + cdt.Day + 'T' + cdt.Hour + cdt.Minute + cdt.Second;
+            var returnVar = $"{cdt.Year}{cdt.Month:D2}{cdt.Day:D2}T{cdt.Hour:D2}{cdt.Minute:D2}{cdt.Second:D2}";
             if (cdt.IsUniversalTime)
             {
                 return returnVar + 'Z';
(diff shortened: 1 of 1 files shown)


rianjs:
@mcshaz - I think there's a bug in your regex:

`var match = Regex.Match(lengthened, '^' + Regex.Escape(componentName) + "(;.+)?:" + Regex.Escape(expectedValue) + _newLine, RegexOptions.Multiline);`

I'm not great with regexes, but I think this test is failing because `MAILTO:` can be folded across multiple lines: `MAIL\r\nTO:`. I'm not sure what the termination conditions for the state machine are. How far is the regex supposed to go? I think it's going too far, or not quite far enough.

I'm fairly certain that `AttendeesSerialized` should pass now, but your unit test is overly restrictive. Might it be better to assert that deserialization back into the object representation contains the properties you'd expect, rather than trying to match strings? That seems like a more robust test.

mcshaz:
> I'm fairly certain that AttendeesSerialized should pass now, but your unit test is overly restrictive. Might it be better to assert that deserialization back into the object representation contains the properties you'd expect, rather than trying to match strings? That seems like a more robust test.

Absolutely - the tests I put up were really as a debug tool and to help me understand the scope of the problem. I personally find it easy to narrow in on problems with more focused but less globally applicable tests, but it is getting near time to make them more robust. 

>  MAILTO: can be folded across multiple lines

I had the spec wrong - it states:

> ### 4.1 Content Lines
> 
> That is, a long line can be split between any two characters by inserting a CRLF immediately followed by a single linear white space character

so the line preceding the regex

``` C#
var lengthened = serialized.Replace(_newLine + ' ', string.Empty);
```

would be better replaced with the more descriptive and accurate:

``` C#
string coalescedMultiline = Regex.Replace(serialized,_newLine + @"\s" ,string.Empty);
```

mcshaz:
I just ran the new code, and saw the issue is the lack of a whitespace character at the beginning of the line (that is serialization output still does not quite meet the spec). 

This brings up a second small advantage of keeping some kind of tests of pure serialization rather than just 'round tripping' serialization to deserialization (beyond testing smaller units of code as mentioned in the comment above) - serialization must strictly meet the spec, but during deserialization, it is often an advantage to be able to handle input which technically does not pass validation.

Issue #36, opened 2016-07-13
Enums in capital case, not all upper case


mcshaz:
First - huge thank you and kudos for getting this project under active development once again!

Setting the properties for an event instance, for example as below:

myEvt.Transparency = TransparencyType.Opaque;
myEvt.Status = EventStatus.Confirmed;

results in a serialised string with the following text:
STATUS:Confirmed
TRANSP:Opaque

this does not validate, as the string should be
STATUS:CONFIRMED
TRANSP:OPAQUE

rianjs:
Thanks for the bug report. You seem comfortable making code changes... do you want to tackle this one?

If I had to guess, I would guess it's an easy-ish fix. Probably something like calling a `ToString` on an enum's name instead of using a `const string` inside a `TransparencyType` class. (Admittedly, I haven't looked.)

mcshaz:
happy to - looking at the AttendeeSerializer class in its current form, the staus is not currently being serialized. I believe you might be working on this as per issue #45. ideally I can uncomment the tests in the deserialize test groups checking that these are failing, fix and rerun, but given the current problems with serialization it seems worth postponing issues to do with letter case until serialization of the properties is occurring.

Finding my way to the attendeeSerializer class, I saw a handful of simple/minor optimisations which are included in a pull request.

mcshaz:
fixed d30880a42f43794d5274e90c9d04e918b134b238

PR #53, opened 2016-07-15
minor optimisations to DataTypeMapper

mcshaz:
as per comments in issue #36.

### DIFF
--- a/ical.NET/General/Proxies/UniqueComponentListProxy.cs
+++ b/ical.NET/General/Proxies/UniqueComponentListProxy.cs
@@ -19,9 +19,10 @@ namespace Ical.Net.General.Proxies
 
         private TComponentType Search(string uid)
         {
-            if (_lookup.ContainsKey(uid))
+            TComponentType componentType;
+            if (_lookup.TryGetValue(uid, out componentType))
             {
-                return _lookup[uid];
+                return componentType;
             }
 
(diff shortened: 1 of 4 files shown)


rianjs:
Viewing on mobile but looks plausible. I think going forward, I'd like a unit test demonstrating that caching dictionaries are thread safe in concurrency scenarios. See the bug I fixed 2 days ago for an example test case. Basically create a unit test with .AsParallel call. If the unit test I wrote the other day covers this by invoking that machinery under the hood, that should be ok. (Just let me know that that is the case.)

Edit, pr link: https://github.com/rianjs/ical.net/pull/41

mcshaz:
Apologies if I am mistaken - I  would be lucky to have a tiny fraction of your programming knowledge (and I am not being humble - my day job has very little to do with computers) - I just wanted to make sure you had seen the changes did not introduce the use of dictionaries (which were pre-existing in the code) - the changes were merely:
- replacing the parameter-less constructor for a dictionary with a constructor specifying a StringComparer.OrdinalIgnoreCase, in order to prevent every lookup value for the dictionary being converted to uppercase first.
- replacing the pattern of `if (dictionary.Contains(key)) {return dictionary[key]}` with the dictionary.TryGetValue method - avoiding 2 lookups where 1 would do.

rianjs:
Quite right. Much easier to see on a real computer. :)

PR #54, opened 2016-07-18
Attendees validating

mcshaz:
several commits
- Lines longer than 75 characters now folding as per spec - [issue #49](https://github.com/rianjs/ical.net/pull/49#issuecomment-233226604)
- Enums now converted ToUpperInvariant, thus enum values like OPAQUE will now validate - issue #36. 
- The same dict.TryGetValue pattern as a commit #53, but applied to the GroupedList class. Also removed checks for dict.ContainsKey prior to Remove application wide, as dict.Remove simply returns false if the key is not found. 
  Looking through the GroupedList code, I share (at least a small fraction of) your pain - your ideas of removing event subscription and using inbuilt lists/dictionaries is absolutely spot on - I really hope you can find the fortitude to have another shot.  I couldn't help myself and added a few notes for myself before I realised just how tenuous the design of this class is - simultaneously maintaining a dictionary and list each pointing to the same objects seems a recipe for tangled code. Most of these comments are signed with BM (Brent McSharry). 
- UTC DateTime now meeting spec - issue #56

this leaves only 2 of my tests failing, and failing appropriately - removing an attendee and adding timezone information

If the v3 plan will be to move more properties to enums (which makes assigning properties valid values very easy given intellisense, and does not require knowing string constants are in some namespace somewhere), further changes will need to be made to the EnumSerializer class. The options to serialize hyphenated values like 'REQ-PARTICIPANT' would include (in my voted order of preference):
1.  An AngularJS style PascalCase to hyphenated  e.g. PASCAL-CASE. in that way `evt.Role = Roles.ReqParticipant`. 
2. The EnumSerializer examines the enum Attributes collection for a [DescriptionAttribute](https://msdn.microsoft.com/en-us/library/system.componentmodel.descriptionattribute%28v=vs.110%29.aspx), and uses this in preference if it is present .


rianjs:
_I really hope you can find the fortitude to have another shot. I couldn't help myself and added a few notes for myself before I realised just how tenuous the design of this class is - simultaneously maintaining a dictionary and list each pointing to the same objects seems a recipe for tangled code. Most of these comments are signed with BM (Brent McSharry)._

It'll get done. "When" is just a function of time and energy. :)

rianjs:
Go ahead and merge; I gave you permissions to the repo. Please continue to open pull requests, though; likely I'll leave some feedback and a :+1: 

(It's not reflective of your development skill; all code is reviewed at work before being merged, and I find it's a model that works well for maintaining a high code quality standard.)

mcshaz:
> Go ahead and merge; I gave you permissions to the repo. Please continue to open pull requests, though; likely I'll leave some feedback and a 👍  (It's not reflective of your development skill; all code is reviewed at work before being merged, and I find it's a model that works well for maintaining a high code quality standard.)

Thank you - in truth I really appreciate the review. I learnt to code as an early teenager in C++ and Pascal, and at the end of the day it is not my day job, so I learn a heap - for instance in your changes I learnt about the $"{}" notation in place of string.Format (syntactic sugar, but very nice), and the ReadOnlyList (wrapper).

PR #78, opened 2016-07-25
serialize/deserialize round trip

mcshaz:
as per previous discussions, tests of strings are removed, and the serialize - > deserialize round trip is used. multiple more tests are added, all of them now passing (214 tests in total passing). 

This request includes multiple commits, and you are welcome to tell me which you would/would not like.

I now realize should have merged the remote sooner in order that you didn't have to see your older code (apologies, I am learning git). Needless to say all the changes you have made in the past week have been included.

As per the titles includes:
- posting a web request to an online validator - tenuous, but of some use at the moment (I have reached out to them to provide JSON output, but no response yet). This has helped me find heaps of things not meeting spec.
- the timezone serialization as per discussion/spec, many comments for the developer
- bug fix for datetime UTC midnight only showing date


Ical.Net.nuspec
```diff
@@ -2,16 +2,14 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
     <metadata>
         <id>Ical.Net</id>
-        <version>2.2.5</version>
+        <version>2.1.5</version>
```
rianjs, line 5: This doesn't seem right. I bumped the minor version number from 1 to 2, because I added a couple of public constants a few PRs ago. ical.net is using [semantic versioning](http://semver.org/):

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 1. MAJOR version when you make incompatible API changes,
> 2. MINOR version when you add functionality in a backwards-compatible manner, and
> 3. PATCH version when you make backwards-compatible bug fixes.

Ical.Net.nuspec
```diff
@@ -2,16 +2,14 @@
...
         <title>Ical.Net (formerly dday.ical)</title>
         <authors>Rian Stockbower, Douglas Day, M. David Peterson</authors>
         <owners>Rian Stockbower</owners>
         <requireLicenseAcceptance>false</requireLicenseAcceptance>
         <description>An iCalendar (RFC 5545) control library for .NET. A fork and continuation of the original dday.ical library including bug fixes, and many performance enhancements.</description>
         <projectUrl>https://github.com/rianjs/ical.net</projectUrl>
         <licenseUrl>https://github.com/rianjs/ical.net/blob/master/license.md</licenseUrl>
-        <iconUrl>https://github.com/rianjs/ical.net/raw/master/logo.png</iconUrl>
```
rianjs, line 13: If you merge this PR, you'll merge these deletions.

rianjs:
So I think this is fundamentally a people question before it is a technology question...

I see you're essentially formulating a form-based POST request, and sending it to their endpoint, getting back a response, and then ripping out the bits you care about, and transforming them into `ValidationFailIssue`s. This is certainly a thing you _could_ do, but I feel it's probably not in the spirit of their validator, and could be considered abuse. I would prefer to have a good reputation in the icalendar community, so I think we should try a diplomatic approach, first. If successful, it will also result in more maintainable code.

These are my preferences in order from most to least preferred. Each starts with an email:
1. Do they expose a REST endpoint we can POST to directly and get back an array of errors?
2. If they do, do they mind if we use it? Then all the HTML parsing goes away, and the code becomes ~10 lines, and much easier to maintain.
3. If they don't and would prefer not to, would they mind giving us a copy? I can probably host it myself. (I have a virtual private server I run my blog and several other things on.)
4. If they don't have an endpoint, and they'd rather not give us a copy, do they mind if we use this approach here. If they don't mind, that's fine, we can continue.
5. If they do mind, I think we respect their wishes and not use their service, except as normal consumers when we have a weird edge case, and want a second opinion.

I think it would help to set their expectations with respect to traffic. What does the request traffic pattern look like? Under what circumstances would we be hitting their validator? Is it a human doing it? Part of the unit test suite? Under some other circumstances? Are going to spike their webhosting bill? Let's be good neighbors, and open a dialog.

mcshaz:
Absolutely agree re people issue, and I have put the email response in the code, as well as including a cc to your personal email. If it turns into a big server load, as you state, I could also host it.

Ical.Net.nuspec
mcshaz, line 13: Thanks for your patience & review as I get my head around Github – While I have used it in the past, it has only been for projects I am developing, and this leaves a number of lessons to be learned and improvements to be made. I will make sure deletions are not re-introduced to the project. 

Enjoy your sabbatical.

khalidabuhakmeh:
This issue needs to be rebased, apologies :(

PR #85, opened 2016-08-03
fixes #27: Using Calendar.AddProperty duplicates property name into value

### LINKED ISSUE #27: Using Calendar.AddProperty duplicates property name into value
This may be myself not understanding the spec for ics/ical files however when using `.AddProperty(string, string)` on a calendar the property name gets added twice.

For example using the following:

```
Calendar cal = new Calendar();
cal.AddProperty("test", "123");
```

Will give a serialised result with the following:

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//github.com/rianjs/ical.net//NONSGML ical.net 2.1//EN
test:test:123
```

It also gets prepended to the value of the property on the calendar object, so its not a serialisation issue.

Edit:

I've just tried the same using the `CalanderProperty` class and its `(string, string)` constructor, seem to have the same issue: `cal.AddProperty(new CalendarProperty("123", "test"));` will result in `123:123:test`, so again the name 123 and a colon is being prepended to the value of the property object

This also happens when adding properties to events.

omd:
This simple change consists of adding just the value of the property instead of its .ToString()-serialization to the object value store.

See bug/issue #27 for details.

### DIFF
--- a/ical.NET.UnitTests/CalendarPropertiesTest.cs
+++ b/ical.NET.UnitTests/CalendarPropertiesTest.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Ical.Net.Interfaces.Serialization;
+using Ical.Net.Serialization;
+using Ical.Net.Serialization.iCalendar.Factory;
+using NUnit.Framework;
+
+namespace Ical.Net.UnitTests
+{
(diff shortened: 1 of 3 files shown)


khalidabuhakmeh:
Thanks @omd

rianjs:
Published in 2.2.6: https://www.nuget.org/packages/Ical.Net

PR #87, opened 2016-08-03
Ignore Serialization test (for now)

khalidabuhakmeh:
This is the only failing test in the entire test suite, and the `TODO` indicates that this will be addressed. Ignoring for now to avoid build / run failures.


rianjs:
👍 

Go ahead and merge.

Issue #89, opened 2016-08-04
DTSTAMP should be in UTC without TimeZone property


kdaveid:
As of RFC `4.8.7.2 Date/Time Stamp` [here](http://www.kanzaki.com/docs/ical/dtstamp.html) or [here](https://www.ietf.org/rfc/rfc2445.txt) the DTSTAMP property should be of UTC but without the TimeZone property.

Unfortunately (as I see it) we have no influence on this particular DateTime property during serialization. 

_Or do you have any ideas how to strip the TZID when serialize this property?_

**EnsureProperties() method**
Additionally and in this matter I propose the followin code change for the default value of DTSTAMP: 

```
if (DtStamp == null)
{
var now = DateTime.SpecifyKind(DateTime.Today.Add(DateTime.UtcNow.TimeOfDay), DateTimeKind.Utc);
DtStamp = new CalDateTime(now);
}
```

This way the CalDateTime() variable has the property `IsUniversalTime` set `true`.

Original:

```
if (DtStamp == null)
{
  // Here, we don't simply set to DateTime.Now because DateTime.Now contains milliseconds, and
  // the iCalendar standard doesn't care at all about milliseconds.  Therefore, when comparing
  // two calendars, one generated, and one loaded from file, they may be functionally identical,
  // but be determined to be different due to millisecond differences.
  var now = DateTime.Now;
  DtStamp = new CalDateTime(now.Year, now.Month, now.Day, now.Hour, now.Minute, now.Second);
}
```

From [here](https://github.com/rianjs/ical.net/blob/master/ical.NET/Components/UniqueComponent.cs#L42-L50).

The problem is we get the followin output when serializing:

```
BEGIN:VCALENDAR
PRODID:TESTCAL
VERSION:2.0
X-WR-CALNAME:X-WR-CALNAME:Test
BEGIN:VEVENT
DESCRIPTION:Description
DTEND;TZID=W. Europe Standard Time:20160804T150000
DTSTAMP;TZID=UTC:20160804T110635Z
DTSTART;TZID=W. Europe Standard Time:20160804T140000
SEQUENCE:0
SUMMARY:Test-Event
UID:1234
END:VEVENT
END:VCALENDAR
```

With [ical-validator](http://ical-validator.herokuapp.com/) we get `Parser error at line 8: Cannot set timezone for UTC properties`.

When removed we have no errors.

rianjs:
(Apologies for the delayed response; I have been on a long holiday out of the country.)

[...]

PR #92, opened 2016-08-17
Utc DateTimes Should not have a TZID.

jhsowter:
This is a fix for #89


rianjs:
@jhsowter - I think your changes were incorporated by @kdaveid. Is it OK to close this up?

PR #93, opened 2016-08-17
TZID property parameter MUST NOT be applied to values in UTC.

kdaveid:
As discussed in #89 
All tests are running. @jhsowter has partially created tests.


rianjs:
Published in 2.2.6: https://www.nuget.org/packages/Ical.Net

PR #129, opened 2016-09-03
Fix #124 with InvariantCulture in double.TryParse()

### LINKED ISSUE #124: GeographicLocationSerializer fail parsing Lat and Lon on other culture
```
double.TryParse(values[0], out lat);
double.TryParse(values[1], out lon);
```

Fails to convert values like 12.3456 if the current culture use 12,3456 as decimal seperator. 

Use CultureInfo.InvariantCulture to fix? 

```
double.TryParse(values[0], NumberStyles.Any, CultureInfo.InvariantCulture, out lat);
double.TryParse(values[1], NumberStyles.Any, CultureInfo.InvariantCulture, out lon);
```

dahermansson:
PR for fix #124

### DIFF
--- a/ical.NET/Serialization/iCalendar/Serializers/DataTypes/GeographicLocationSerializer.cs
+++ b/ical.NET/Serialization/iCalendar/Serializers/DataTypes/GeographicLocationSerializer.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using Ical.Net.DataTypes;
 using Ical.Net.Interfaces.DataTypes;
+using System.Globalization;
 
 namespace Ical.Net.Serialization.iCalendar.Serializers.DataTypes
 {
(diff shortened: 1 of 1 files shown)


ical.NET/Serialization/iCalendar/Serializers/DataTypes/GeographicLocationSerializer.cs
```diff
@@ -45,9 +46,9 @@ public GeographicLocation Deserialize(string value)
 
             double lat;
             double lon;
-            double.TryParse(values[0], out lat);
-            double.TryParse(values[1], out lon);
-            g.Latitude = lat;
+			double.TryParse(values[0], NumberStyles.Any, CultureInfo.InvariantCulture, out lat);
```
rianjs, line 49: You have some weird indentation here. Can you line it up with the rest of the lines?

dahermansson: Looks weird. It's correct i Visual Studio and Git don't track changes if I redo the indentation on the rows or the hole method. 
Don't know how to fix that if there is a real problem.

rianjs:
Go ahead and merge it. I made you a contributor. (You might have to accept an invitation or something first.)

ical.NET/Serialization/iCalendar/Serializers/DataTypes/GeographicLocationSerializer.cs
rianjs: It's because they're tabs (`\t`) instead of spaces. ical.net uses 4 spaces for each tab stop, not real tabs. When I bumped the nuget version number, I converted them to spaces.

It would be nice if Visual Studio had a "Figure out what this project is doing, and copy it" option for whitespace.

Issue #150, opened 2016-10-03
Case sensitivity issue when building from mono/linux ...


fvn-linagora:
... using solution file (ical.sln) and xbuild.

```
Warnings:

/home/username/src/ical.net/ical.net.sln:  warning : Project file /home/username/src/ical.net/ical.NET/Ical.Net.csproj referenced in the solution file, not found. Ignoring.
/home/username/src/ical.net/ical.net.sln:  warning : Project file /home/username/src/ical.net/ical.NET.UnitTests/Ical.Net.UnitTests.csproj referenced in the solution file, not found. Ignoring.
/home/username/src/ical.net/ical.net.sln:  warning : Project file /home/username/src/ical.net/ical.NET.Collections/ical.NET.Collections/Ical.Net.Collections.csproj referenced in the solution file, not found. Ignoring.

Errors:

/home/username/src/ical.net/ical.net.sln (default targets) ->
(Build target) ->

        /home/username/src/ical.net/ical.net.sln: error : Could not find the project file '/home/username/src/ical.net/ical.NET/Ical.Net.csproj'
        /home/username/src/ical.net/ical.net.sln: error : Could not find the project file '/home/username/src/ical.net/ical.NET.UnitTests/Ical.Net.UnitTests.csproj'
        /home/username/src/ical.net/ical.net.sln: error : Could not find the project file '/home/username/src/ical.net/ical.NET.Collections/ical.NET.Collections/Ical.Net.Collections.csproj'

         3 Warning(s)
         3 Error(s)
```

rianjs:
1402627cf928f105bcec58866e2fa2e3fdb998cb

PR #151, opened 2016-10-03
Fix case sensitivity issue using mono/linux

fvn-linagora:
Solve issue #150
Note: I've noticed this issue should have already been addressed (along with others) in v3 branch (see https://github.com/rianjs/ical.net/commit/a6c263bb5cb6fa35012ca1994d9bf5b43f4c7c37)


rianjs:
Oh, interesting. I've kind of been wondering what the cross-platform story was going to be for case sensitivity. Non-windows platforms are usually case-sensitive, but pretty much the whole world in .NET is built on CI assumptions. Even namespaces in C# files are case-insensitive. (Probably because VB.NET is case-insensitive.)

Anyway, thanks for the fix!

fvn-linagora:
Indeed, this PR is renaming files to work around the mismatched project references issues.
But then, project filenames, namespaces and folder hierarchy no longer have consistent casing (ical.NET), even though .NET really don't care :)
I guess your patch (on v3 branch, https://github.com/rianjs/ical.net/commit/a6c263bb5cb6fa35012ca1994d9bf5b43f4c7c37) is more consistent in this regard.
Anyway, thank you for merging my pull request !

Issue #152, opened 2016-10-03
Failing unit tests when run from mono/linux (on master)


fvn-linagora:
most failing tests are using Calendar.LoadFromStream() with local datetimes.

see attached files for errors detail.

[TestResult.xml.txt](https://github.com/rianjs/ical.net/files/506192/TestResult.xml.txt)
[ical.failures.txt](https://github.com/rianjs/ical.net/files/506190/ical.failures.txt)

rianjs:
Should be fixed in non-beta versions of 3+

PR #153, opened 2016-10-03
Fix local date time provider on mono

fvn-linagora:
Solve issue #152 

When using Calendar.LoadFromStream on mime content with missing (or local) timezones, default provider (Tzdb) was throwing `NodaTime.TimeZones.DateTimeZoneNotFoundException` at runtime.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: csharp
+
+sudo: false  # use the new container-based Travis infrastructure 
+
+install:
+  - nuget restore v2
+  - nuget restore v3
+  - nuget install NUnit.Runners -Version 3.4.1 -OutputDirectory testrunner
+
+script: 
+  - xbuild v2/ical-v2.sln
+  - mono ./testrunner/NUnit.ConsoleRunner.3.4.1/tools/nunit3-console.exe v2/ical.NET.UnitTests/bin/Debug/net46/Ical.Net.UnitTests.dll
+  - xbuild v3/ical-v3.sln
+  - mono ./testrunner/NUnit.ConsoleRunner.3.4.1/tools/nunit3-console.exe v3/ical.NET.UnitTests/bin/Debug/net46/Ical.Net.UnitTests.dll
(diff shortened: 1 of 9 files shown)


dahermansson:
Is DateTimeZoneProviders.Bcl property available on the PCL build of Noda Time or not?

http://nodatime.org/unstable/api/html/P_NodaTime_DateTimeZoneProviders_Bcl.htm

rianjs:
I'd like to echo @dahermansson's comment. I'd also like to add that I really wish there was more clarity around time zone handling in .NET Core going forward.

I asked a related question on on StackOverflow a month or two ago:
http://stackoverflow.com/questions/38190911/where-is-the-bcl-datetimezoneprovider-in-the-net-core-implementation-of-nodatim

fvn-linagora:
OK, I understand.
Indeed, the provided patch was assuming the library is not targeting other framework profiles.

fvn-linagora:
following-up on your remarks regarding PCL build, the (crude) fix provides a workaround for *nix platforms that "duplicates" behavior from missing Bcl implementation,

Note: only last commit is relevant (was rebased on https://github.com/rianjs/ical.net/pull/192)

fvn-linagora:
FYI, I've also added a  [Travis CI](https://travis-ci.org/) build configuration to prove unit tests pass on mono/linux ([example report](https://travis-ci.org/fvn-linagora/ical.net/builds/181432173)).
Note that preceding PR (https://github.com/rianjs/ical.net/pull/192) [build is failing](https://travis-ci.org/fvn-linagora/ical.net/builds/181431498) as expected, ie compilation passes, but unit tests have failures current PR is trying to fix.
Hope this helps.

fvn-linagora:
Fix is actually trying to provide a fragile (and probably incomplete) workaround for a specific setup (ie mono + NodaTime).
Ideally, the issue should be fixed upstream (cascading NodaTime -> mono).
In the meantime I'll probably try to maintain a separate nuget package without PCL support.

rianjs:
I haven't looked in a while, but the .NET Core version of NodaTime was actively being worked on, which I think may solve the issue. It remains to be seen how support for multiple time zone data providers will be. I think Jon was thinking about doing a separate data package (via nuget) to separate the data and the code. That was several months ago.

PR #164, opened 2016-10-18
V3

rianjs:
Merging into master, so I can more easily have side-by-side versions instead of doing one gigantic merge at the end. This way they can live independent lives.


rianjs:
Apparently I've touched stuff in `v2`. WTF.

PR #196, opened 2016-12-06
Proposed breaking change to Attendee deserialization

dwnichols:
Currently the deserialization of an attendee will return null if the calendar user address is not a valid URI, even after some custom fixes to ensure its a valid mailto URI. I propose we change this to always return an attendee with a null Value property if the URI is malformed. Also, an additive change to include the original calendar user address in a property of Attendee.

The issue with the current approach is that when syncing calendars any malformed attendee will be ignored and the calling code will have no insight into fact. I believe it should the callers responsibility to decide how to handle these cases.

I've split the v2 and v3 changes into separate commits as I understand it may not be possible to merge breaking changes within v2.


dwnichols:
@rianjs : any thoughts on this?

dwnichols:
@rianjs : do you have time to take a look at this PR?

rianjs:
Sorry, my time has been limited, and I've been mainly only working in areas that affect my day job.

Thoughts:

* Thanks for noticing v2 vs v3
* I guess this is "breaking", but it's "breaking" in a way that gets us closer to the spec, [which defines CAL-ADDRESS as a URI](https://tools.ietf.org/html/rfc5545#section-3.3.3). I'm pretty OK with that.
* I think you didn't go far enough...

...if we're gonna break the API, I would rather break it completely:

* If someone wants to try to put something in there that's not a valid URI, we should let the `Uri` object throw an exception. I'm not a big fan of "Postel's law": bad data = exception
* If the typed Uri value is null, we shouldn't return `UserAddress` when serializing/deserializing. So I guess I would get rid of `UserAddress` entirely. If you really want the original string, use `Uri.OriginalString`, which brings me to:
* We should use the `Uri` object, and on serialization/deserialization, we should find some way to express its "Uri-ness" in terms of the C# `Uri` API, rather than doing string manipulation. (`Uri` has a `Scheme` property which handles `mailto:`) (`AttendeeSerializer.Deserialize` was the culprit I saw nearby that got me thinking about this.)
* If someone wants to include non-standard address stuff, let them use the `Properties` key-value pairs, which generates `X-` prefixed fields [indicating application-specific values](https://tools.ietf.org/html/rfc5545#section-3.8.8.2).

SemVer: I think if you get rid of UserAddress, we can leave this as 2.2.*. because while it "breaks" stuff, it's only breaking things that _should_ break, and if we lived in a bug-free world _would always have broken to begin with_.

dwnichols:
Hey Rian,

Thanks for you feedback. My concern with allowing Uri to throw an exception with bad data is that we don't always have control over the data. I'm working on integrations with a number of online calendars and I'm finding that there is actually a large amount of bad data in the iCalendar ics world. Each calendar provider seems to propagate the data generated (good or bad) by calendaring applications. 

Here are a few examples I have run into:

```
ATTENDEE;CUTYPE=INDIVIDUAL;ROLE=REQ-PARTICIPANT;PARTSTAT=ACCEPTED;CN=Test A
 ttendee 1;X-NUM-GUESTS=0:mailto:"test@test.com"
ATTENDEE;CUTYPE=INDIVIDUAL;ROLE=REQ-PARTICIPANT;PARTSTAT=ACCEPTED;CN=Test A
 ttendee 1;X-NUM-GUESTS=0:mailto:test@test.com?
ATTENDEE;CUTYPE=INDIVIDUAL;ROLE=REQ-PARTICIPANT;PARTSTAT=ACCEPTED;CN=Test A
 ttendee 1;X-NUM-GUESTS=0:mailto:test@test.com^v
```

Each of the mailto uri's are invalid, however the calendar providers handle the bad data by cleaning it before presenting it to the user. If my integrations do not have the same behaviour it looks like a bug on my end. Ultimately, this is a question of whether the ical.net library should be strictly adhering to the RFC, or allow the application code leveraging the library make that decision. Alternatively, the library could provide a non-strict setting which allows the user to choose.

Issue #235, opened 2017-02-06
GetOccurrences ignore some day range when I use weekly occurrence


kycuong2208:
Hi all

I got the issue when I use Ical.net to get list Occurrences in calendar. List Occurrences will miss some event in day range such as the event from 1/1/2019 to 1/7/2019. My example below:

```csharp
var calendar = new Calendar();
var rule = new RecurrencePattern("FREQ=WEEKLY;BYDAY=SU,MO,TU,WE,TH,FR,SA");
var startDateTemp=new DateTime(2018,12,30);
var endDateTemp = new DateTime(2018, 12, 31);
var vEvent = new Event
{
    DtStart = new CalDateTime(startDateTemp),
    DtEnd = new CalDateTime(endDateTemp),
    RecurrenceRules = new List<IRecurrencePattern>{ ruleTemp }
};
calendar.Events.Add(vEvent);
var startDate= new DateTime(2019, 1, 1);
var endDate = new DateTime(2019, 1, 6);
var occurrences = calendar.GetOccurrences(startDate, endDate);
```

=> and we will got list occurrences don't have any records.

Note: If I use daily occurrence, this issue not happen. When I tested with another date( not in range 1/1/2019 to 1/6/2019) , the system can execute exactly.

Please give me your advices.

Thanks

rianjs:
@DmitriySelischev's fix published in nuget v2.2.33.

https://www.nuget.org/packages/Ical.Net/2.2.33

PR #244, opened 2017-03-13
Fixed issue with getting occurrences at the border of years

ghost:
Fixed issue [#235](https://github.com/rianjs/ical.net/issues/235)
This issue happened when weekly pattern evaluates on border of 2 years: for example 2018 and 2019. We have 53th week at 31.12.2018(Monday), but 1st week at 01.01.2019(Tuesday). I've added additional check for border situations and covered it with unit test.

### DIFF
--- a/v2/Ical.Net.nuspec
+++ b/v2/Ical.Net.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
     <metadata>
         <id>Ical.Net</id>
-        <version>2.2.32</version>
+        <version>2.2.33</version>
         <title>Ical.Net</title>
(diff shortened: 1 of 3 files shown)


v2/Ical.Net.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
     <metadata>
         <id>Ical.Net</id>
-        <version>2.2.32</version>
+        <version>2.2.33</version>
```
rianjs, line 5: (In future PRs, please don't change the release number)

rianjs (commented):
Thanks for the fix. I'll merge, and copy these changes over to the net-core project.

net-core is a side-by-side copy of ical.net that will become v3 of ical.net once nunit and ReSharper play nicely with .net core. Which apparently should be soon.

PR #247, opened 2017-03-17
Minor performance improvements

chescock:
These are a few other potential performance improvements I noticed while profiling deserialization.

### DIFF
--- a/net-core/Ical.Net/Ical.Net/Serialization/iCalendar/Serializers/Other/StringSerializer.cs
+++ b/net-core/Ical.Net/Ical.Net/Serialization/iCalendar/Serializers/Other/StringSerializer.cs
@@ -126,8 +126,6 @@ namespace Ical.Net.Serialization.iCalendar.Serializers.Other
                 serializeAsList = GetService<IDataTypeMapper>().GetPropertyAllowsMultipleValues(co);
             }
 
-            value = TextUtil.Normalize(value, SerializationContext).ReadToEnd();
-
             // Try to decode the string
             EncodableDataType dt = null;
             if (co != null)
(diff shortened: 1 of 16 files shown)


net-core/Ical.Net/Ical.Net/Serialization/iCalendar/Serializers/Other/StringSerializer.cs
```diff
@@ -102,7 +102,7 @@ public override string SerializeToString(object obj)
             return string.Join(",", values);
         }
 
-        internal static readonly Regex UnescapedCommas = new Regex(@"[^\\](,)", RegexOptions.Compiled);
+        internal static readonly Regex UnescapedCommas = new Regex(@"(?<!\\),", RegexOptions.Compiled);
```
rianjs, line 105: What does this do?

rianjs (commented):
Oh, I specified collection sizes a while ago. 128 is probably unnecessarily large. 4 would probably suffice in 80%+ of cases, and since 4 is effectively "free", I'm OK with the change.

[This was my thinking](http://stackoverflow.com/questions/2247773/is-it-worthwhile-to-initialize-the-collection-size-of-a-listt-if-its-size-rea) when I specified initial sizes.

net-core/Ical.Net/Ical.Net/Serialization/iCalendar/Serializers/Other/StringSerializer.cs
chescock: It matches commas that are not preceded by backslashes.  The `(?<foo)` construct is a zero-width negative lookbehind assertion, which means it ensures the characters *before* that point don't match the given regex.  

Here: https://regex101.com/r/baxiSh/1

PR #251, opened 2017-03-23
Declare dependency on NodaTime in the .nuspec

remyjette:
The Ical.Net .csproj file declares that it references NodaTime, but the .nuspec file does
not declare the dependency. As a result, Ical.Net will throw exceptions unless the NodaTime
NuGet package is installed separately. This change declares the NodaTime dependency in the
.nuspec so NuGet will automatically install NodaTime when Ical.Net is installed.


rianjs:
Ah, I thought its presence is ical.net.deps.json (or whatever the file is called) would be sufficient. Thanks!

rianjs:
https://www.nuget.org/packages/Ical.Net/3.0.1-net-core-alpha

PR #256, opened 2017-03-30
Define IEquitable on Calendar and Occurence classes

ie-zero:
Both Calendar and Occurence classes are already implementing the IEquatable<T> interface. This change simply add the interface at the class definition.

### DIFF
--- a/net-core/Ical.Net/Ical.Net/Calendar.cs
+++ b/net-core/Ical.Net/Ical.Net/Calendar.cs
@@ -17,7 +17,7 @@ using Ical.Net.Utility;
 
 namespace Ical.Net
 {
-    public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, IMergeable
+    public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, IMergeable, IEquatable<Calendar>
     {
         public static CalendarCollection Load(string iCalendarString)
         {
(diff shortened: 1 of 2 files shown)


rianjs:
Good idea. I would take it further, though, because I think there's a lot more opportunity/value to be had. Many (perhaps most) calendar components implement `GetHashCode` and `Equals`. It's reasonable to implement `IEquatable` on those objects as well.

Looking quickly:

- `Event` / `CalendarEvent` (in net-core)
- `Alarm`
- `Journal`
- `GeographicLocation`
- `Attachment`?
- `Period`
- `StatusCode`
- `Organizer`
- `Attendee`
- `Trigger`?
- `UtcOffset`?
- `VTimeZone` / `TimeZone` (I think it's called TimeZone in net-core, I can't remember)
- `Todo`

And there are two versions of ical. They're mirror images of each other (mostly), so changes need to be made in both places. (When nunit works well with .NET Core, I'll retire the v2 version, and just maintain the version in `net-core`.)

* **v2**, located in the `v2` folder -- this is the main version
* **net-core**, located in the `net-core` folder -- being published as an alpha nuget package

The publishing steps:

* In the nuspec file in both of those directories, increment the minor version number
* Add a bullet to `readme.md` adding a brief note about what the minor version changes achieve. Something like "Implement `IEquatable<T>` on `Calendar`, `Occurrence`, `Alarm`, `Event`/`CalendarEvent`, `Journal` <etc>"

ie-zero:
Let me take this work of your hands. I will be happy to do it.

I spotted some of those opportunities as well as I familiarize myself with the codebase but I wanted it see if there was any value of doing so.

rianjs:
> I spotted some of those opportunities as well as I familiarize myself with the codebase but I wanted it see if there was any value of doing so.

Yes, I think implementing IEquatable and IComparable where appropriate is definitely worthwhile.

PR #261, opened 2017-04-06
Ex date fixes

rianjs:


rianjs:
Available in:
* 3.0.10-net-core-beta
* 2.3.2

PR #262, opened 2017-04-08
Define IEquatable when already implemented

ie-zero:
All the classes amended have had implemented the IEquatable<T> interface already. This change simply add the interface at the class definition.

### DIFF
--- a/net-core/Ical.Net/Ical.Net/Components/UniqueComponent.cs
+++ b/net-core/Ical.Net/Ical.Net/Components/UniqueComponent.cs
@@ -11,7 +11,7 @@ namespace Ical.Net
     /// Represents a unique component, a component with a unique UID,
     /// which can be used to uniquely identify the component.    
     /// </summary>
-    public class UniqueComponent : CalendarComponent, IUniqueComponent, IComparable<UniqueComponent>
+    public class UniqueComponent : CalendarComponent, IUniqueComponent, IComparable<UniqueComponent>, IEquatable<UniqueComponent>
     {
         // TODO: Add AddRelationship() public method.
         // This method will add the UID of a related component
@@ -96,8 +96,28 @@ namespace Ical.Net
         public int CompareTo(UniqueComponent other)
             => string.Compare(Uid, other.Uid, StringComparison.OrdinalIgnoreCase);
 
+        public bool Equals(UniqueComponent other)
+        {
+            // TODO: Casting UniqueComponent to RecurringComponent, i.e. a subclass, creates tight coupling of the two classes. 
+            //      This implementation should be moved to RecurringComponent class instead. 
+
+            if (other is RecurringComponent && other != this)
+            {
+                var r = (RecurringComponent)other;
+                if (Uid != null)
+                {
+                    return Uid.Equals(r.Uid);
+                }
+                return Uid == r.Uid;
+            }
+            return base.Equals(other);
+        }
+
         public override bool Equals(object obj)
         {
+            // TODO: Casting UniqueComponent to RecurringComponent, i.e. a subclass, creates tight coupling of the two classes. 
+            //      This implementation should be moved to RecurringComponent class instead. 
+
             if (obj is RecurringComponent && obj != this)
             {
                 var r = (RecurringComponent) obj;
--- a/net-core/Ical.Net/Ical.Net/DataTypes/PeriodList.cs
+++ b/net-core/Ical.Net/Ical.Net/DataTypes/PeriodList.cs
@@ -33,7 +33,7 @@ namespace Ical.Net.DataTypes
 
         public bool Equals(PeriodList other)
         {
-            return string.Equals(TzId, other.TzId) && CollectionHelpers.Equals(Periods, other.Periods);
+            return string.Equals(TzId, other.TzId, StringComparison.OrdinalIgnoreCase) && CollectionHelpers.Equals(Periods, other.Periods);
         }
 
         public override bool Equals(object obj)
--- a/net-core/Ical.Net/Ical.Net/DataTypes/PeriodList.cs
+++ b/net-core/Ical.Net/Ical.Net/DataTypes/PeriodList.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
@@ -12,7 +13,7 @@ namespace Ical.Net.DataTypes
     /// <summary>
     /// An iCalendar list of recurring dates (or date exclusions)
     /// </summary>
-    public class PeriodList : EncodableDataType, IList<Period>
+    public class PeriodList : EncodableDataType, IList<Period>, IEquatable<PeriodList>
     {
         public string TzId { get; set; }
         public int Count => Periods.Count;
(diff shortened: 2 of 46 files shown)


ie-zero:
This is the follow up PR to #256 "Define IEquatable on Calendar and Occurrence classes"

net-core/Ical.Net/Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -30,7 +31,7 @@ public PeriodList(string value) : this()
             CopyFrom(serializer.Deserialize(new StringReader(value)) as ICopyable);
         }
 
-        protected bool Equals(PeriodList other)
+        public bool Equals(PeriodList other)
         {
             return string.Equals(TzId, other.TzId) && CollectionHelpers.Equals(Periods, other.Periods);
```
rianjs: While you're here, let's make all `string.Equals` use `StringComparer.OrdinalIgnoreCase`.

net-core/Ical.Net/Ical.Net/DataTypes/RequestStatus.cs
```diff
@@ -62,7 +63,7 @@ public override string ToString()
             return serializer.SerializeToString(this);
         }
 
-        protected bool Equals(RequestStatus other)
+        public bool Equals(RequestStatus other)
         {
             return string.Equals(_mDescription, other._mDescription) && string.Equals(_mExtraData, other._mExtraData) &&
                    Equals(_mStatusCode, other._mStatusCode);
```
rianjs: Let's use `OrdinalIgnoreCase`, here and elsewhere.

net-core/Ical.Net/Ical.Net/DataTypes/WeekDay.cs
```diff
@@ -64,6 +66,20 @@ public override void CopyFrom(ICopyable obj)
             }
         }
 
+        public int CompareTo(WeekDay other)
+        {
+            if (other == null)
```
rianjs: Throwing an exception in a `CompareTo` method probably isn't a good idea. Consider the usage in a LINQ query:

```csharp
var orderedDays = weekdayCollection
    .OrderBy(w => w)
    .ToList();
```

That could `throw` in the event that WeekDay is null. I think being able to sort `null` is reasonable behavior: any non-null WeekDay has greater weight. Consider instead:

```csharp
public int CompareTo(WeekDay other)
{
    if (ReferenceEquals(this, other)) return 0;
    return ReferenceEquals(null, other)
        ? 1
        : DayOfWeek.CompareTo(other.DayOfWeek);
}
```

I'm not sure using `Offset` as a tie-breaker is necessary or desirable.

net-core/Ical.Net/Ical.Net/DataTypes/WeekDay.cs
```diff
@@ -64,6 +66,20 @@ public override void CopyFrom(ICopyable obj)
...
+            }
+            var compare = DayOfWeek.CompareTo(other.DayOfWeek);
+            if (compare == 0)
+            {
+                compare = Offset.CompareTo(other.Offset);
+            }
+            return compare;
+        }
+
         public int CompareTo(object obj)
```
rianjs: Let's get rid of this `IComparable` implementation in favor of just `IComparable<T>`, so:
* Delete this method
* Delete `IComparable` from `IWeekDay` in v2, and delete `IComparable` from the top here in net-core.

rianjs:
High-level thoughts:

* Let's delete all the non-generic `IComparable` and `IEquatable`s that exist, and use their generic equivalents instead: `IComparable<T>` and `IEquatable<T>`
* When you delete the non-generic interfaces, place the generic replacement on the object itself instead of the interface like you have done in this pull request. For example: `WeekDay` should implement `IComparable<WeekDay>`, and delete `IComparable` from `IWeekDay`. (I'm trying to simplify interface dependencies, which are out of control.)
* `IComparable<T>` implementations should allow null values without throwing exceptions. [This StackOverflow thread](http://stackoverflow.com/questions/8642080/icomparable-behaviour-for-null-arguments) explains what I mean.
* Where there are `string.Equals` calls, let's make sure they all specify `StringComparer.OrdinalIgnoreCase`. [Case-insensitive is OK for properties and most other component values, because RFC-5545 doesn't specify case sensitivity in most cases](https://tools.ietf.org/html/rfc5545#section-2).

ie-zero:
Thank you Rian (@rianjs) for the code review and helpful feedback. The additional PRs, I believe, implement all the above points.

Please let me know about any further improvements that you might see fit.

net-core/Ical.Net/Ical.Net/Components/UniqueComponent.cs
```diff
@@ -96,8 +96,28 @@ protected override void OnDeserialized(StreamingContext context)
...
+
+            if (other is RecurringComponent && other != this)
+            {
+                var r = (RecurringComponent)other;
+                if (Uid != null)
+                {
+                    return Uid.Equals(r.Uid);
+                }
+                return Uid == r.Uid;
+            }
```
rianjs, line 112: * Using the `Uid` as the basis for equality isn't consistent with the rest of the library. (I know that's the current implementation -- which is wrong, so not your fault.)
* This method should delegate to `RecurringComponent.Equals()` in the event that the element is recurring, and call its base class in the event that the element is NOT recurring. I don't think there's anything inside `UniqueComponent` that's useful for determining equality or hashing. 

`RecurringComponent` should know how to do a reasonable comparison and hash, but I think it's missing the `IEquatable<T>`, and isn't defining an `Equals` method, so we have nothing to delegate to. That will need to be implemented.

If you want, you can add the `IEquatable<T>` to the `RecurringComponent`, and define an `Equals` function. I think it will require some thought for both `GetHashCode` and `Equals`. If you don't want to do that, or would prefer to do it as a separate pull request, you can remove the `IEquatable` from this class, and add it back in with the changes to RecurringComponent. (Or I'll do it myself.)

Actually, I think that work should be done as a separate ticket, because it's a little more significant, and needs to be done carefully.

v2/ical.NET/Components/UniqueComponent.cs
```diff
@@ -95,8 +95,26 @@ protected override void OnDeserialized(StreamingContext context)
...
+            if (other is RecurringComponent && other != this) {
+                var r = (RecurringComponent)other;
+                if (Uid != null) {
+                    return Uid.Equals(r.Uid);
+                }
+                return Uid == r.Uid;
+            }
+            return base.Equals(other);
+        }
+
```
rianjs, line 112: (Same comment here,)

ie-zero:
I also think it is better for this work to be done under a different PR. 

With this PR we are in at the excact same state as before (no better or worst) so, if you are happy, we can merge it in and open a ticket to rework the Equal() in RecurringComponent and UniqueComponent classes.

rianjs (approved):
Can you merge this after I submit my approval, or do I need to tweak the repo settings?

rianjs:
Oh, two last, easy things to tack onto this PR:

* Please open the two nuspec files, and bump the minor revision number
* Please update `readme.md` to summarize the changes you made. Something like "All calendar components now explicitly implement `IEquatable<T>`", should be sufficient.

After that, I'll build and publish an updated nuget package.

ie-zero:
I never pushed changes directly into the original repository. Sorry to be a pain but would it possible to provide me with some guidance or merge it yourself please?

PR #265, opened 2017-04-11
Add unit tests for RequestStatus class

ie-zero:
* Unit tests for RequestStatus class
* Fix: The CopyFrom(ICopyable) method in RequestStatus class never copied the value for the parameter object for ExtraData property.

The changes are affecting both the v2 and .NET Core projects.

### DIFF
--- a/net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
+++ b/net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using Ical.Net.DataTypes;
 using Ical.Net.Interfaces.General;
 using NUnit.Framework;
@@ -8,15 +9,10 @@ namespace Ical.Net.UnitTests
 {
     public class RequestStatusTest
     {
-        private static IEnumerable<ITestCaseData> ConstructorTestCases()
+        [Test]
+        public void NullStringConstructorThrowsException()
         {
-            var emptyRequestStatus = new RequestStatus();
-            yield return new TestCaseData(emptyRequestStatus, null, null, null)
-                .SetName("Default request status constructor has null properties");
-
-            var invalidSerializedValue = new RequestStatus("ABC");
-            yield return new TestCaseData(invalidSerializedValue, null, null, null)
-                .SetName("Constructor with invalid serialised text has null properties");
+            Assert.Throws<ArgumentNullException>(() => new StatusCode((string)null));
         }
 
         [Test, TestCaseSource(nameof(ConstructorTestCases))]
@@ -27,31 +23,19 @@ namespace Ical.Net.UnitTests
             Assert.AreEqual(incoming.ExtraData, expectedExtraData);
         }
 
+        private static IEnumerable<ITestCaseData> ConstructorTestCases()
+        {
+            var empty = new RequestStatus();
+            yield return new TestCaseData(empty, null, null, null)
+                .SetName("Default request status constructor has null properties");
+
+            var invalidSerializedValue = new RequestStatus("ABC");
+            yield return new TestCaseData(invalidSerializedValue, null, null, null)
+                .SetName("Constructor invoked with invalid serialised text has null properties");
+        }
+
         [Test, Ignore("Understand what 'valid' values RequestStatusSerializer expects")]
         public void InitialObjectState_Constructor_ValidSerialisedValue() { }
-
-        private static IEnumerable<ITestCaseData> ToStringTestCases()
-        {
-            var emptyRequestStatus = new RequestStatus();
-            yield return new TestCaseData(emptyRequestStatus, ";")
-                .SetName("ToString() for empty request status should return semicolon");
-
-            var emptyStringPropertiesRequestStatus = new RequestStatus()
-            {
-                Description = "",
-                ExtraData = ""
-            };
-            yield return new TestCaseData(emptyStringPropertiesRequestStatus, ";")
-                .SetName("ToString() for request status with empty string properties should return semicolon");
-
-            var nonEmptyStringPropertiesRequestStatus = new RequestStatus()
-            {
-                Description = "ABC Description",
-                ExtraData = "ABC ExtraData"
-            };
-            yield return new TestCaseData(nonEmptyStringPropertiesRequestStatus, ";ABC Description;ABC ExtraData")
-                .SetName("ToString() for request status should return semicolon separated values");
-        }
 
         [Test, TestCaseSource(nameof(ToStringTestCases))]
         public void ToStringTests(RequestStatus incoming, string expected)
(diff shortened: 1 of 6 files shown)


rianjs:
* First off: looks good, thank you.
* Second: I don't think the `[TestFixture]` attribute is required anymore? That could just be ReSharper being smart -- maybe other test runners require it; I don't know.
* Third: Are you familiar with `TestCaseSource`? It will let you reuse a lot of the code, which will make your test cases smaller, and allows you to give them real names. It allows you to be more expressive while being less verbose. There are a bunch examples in `SymmetricSerializationTests`, but here're two examples:

```csharp
[Test, TestCaseSource(nameof(ConstructorTestCases))]
public void ConstructorTests(RequestStatus incoming, string expectedDescription, string expectedStatusCode, string expectedExtraData)
{
    Assert.AreEqual(incoming.StatusCode, expectedStatusCode);
    Assert.AreEqual(incoming.Description, expectedDescription);
    Assert.AreEqual(incoming.ExtraData, expectedExtraData);
}

public static IEnumerable<ITestCaseData> ConstructorTestCases()
{
    var emptyRequestStatus = new RequestStatus();
    yield return new TestCaseData(emptyRequestStatus, null, null, null)
        .SetName("Empty request status has null properties")
    
    var invalidSerializedValue = new RequestStatus("ABC");
    yield return new TestCaseData(invalidSerializedValue, null, null, null)
        .SetName("Constructor with invalid text has null properties");
}
```

You can also return values from a `[Test]`, and check them in the `TestCaseSource`. That syntax looks like this:

```csharp
[Test, TestCaseSource(nameof(IdentityTestCases))]
public int IdentityTests(int i) => return i;

public static IEnumerable<ITestCaseData>(IdentityTestCases)
{
    foreach (var i in Enumerable.Range(0, 10))
    {
        yield return new TestCaseData(i)
            .Returns(i)
            .SetName($"{i} returns {i}");
    }
}
```

ie-zero:
I never worked with nunit before but from what I can see it is the same as MemberData and Theory in xUnit. 

Thanks for the examples. I will have a look.

ie-zero:
Rian you are absolutely right about the ```[TestFixture]```. It is not required anymore except in specific cases,

> Beginning with NUnit 2.5, the TestFixture attribute is optional for non-parameterized, non-generic fixtures, so long as the class contains at least one method marked with the Test, TestCase or TestCaseSource attribute. (http://www.nunit.org/index.php?p=testFixture&r=2.5)

net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
```diff
@@ -1,122 +1,126 @@
...
-            Assert.IsFalse(requestStatusA.Equals(null));
-            Assert.IsFalse(requestStatusA.Equals("Invalid type value"));
-            Assert.IsTrue(requestStatusA.Equals(requestStatusA));
-            Assert.IsTrue(requestStatusA.Equals(requestStatusB));
-            Assert.IsTrue(requestStatusA.Equals(requestStatusC));
-            Assert.IsFalse(requestStatusA.Equals(requestStatusD));
+        [Test, TestCaseSource(nameof(EqualsTestCases))]
+        public void EqualsTests(RequestStatus incoming, object other, bool expected)
+        {
+            Assert.AreEqual(expected, incoming.Equals((object)other));
```
rianjs: I think this cast is redundant.

net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
```diff
@@ -1,122 +1,126 @@
...
         }
```
rianjs: Please put the TestCaseSource after the `[Test]` method.

net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
```diff
@@ -1,122 +1,126 @@
...
+        [Test, TestCaseSource(nameof(GetHashCodeTestCases))]
+        public void GetHashCodeTests(RequestStatus incoming, int expected)
         {
-            var requestStatus = new RequestStatus();
-            requestStatus.CopyFrom(null);
-
-            Assert.IsNull(requestStatus.StatusCode);
-            Assert.IsNull(requestStatus.Description);
-            Assert.IsNull(requestStatus.ExtraData);
+            Assert.AreEqual(expected, incoming.GetHashCode());
```
rianjs: The primary value of testing hash codes is not to assert what the value should be, but rather to make sure the hash is stable and deterministic. To that end, I would prefer to do a side by side comparison of hash values, instead of hard-coding a specific hash value. You did this already with `Equals()`. In most respect, `GetHashCode` and `Equals` are two sides of the same coin, and are amenable to similar test strategies. Inside the function, the properties and fields that are considered should be the same between the two.

Example: consider a more complicated object like `CalendarEvent`. What if we changed the `Summary` string's hashing to be culture-specific? That would break a test that uses a hard-coded constant, even though the change isn't wrong. But if we compare the output of the two `GetHashCode()` calls, they'd better match, otherwise our hash function is bad. 

```csharp
var a = new RequestStatus();
var b = new RequestStatus();
Assert.AreEqual(a.GetHashCode(), b.GetHashCode());
```

net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
```diff
@@ -126,56 +130,63 @@ private class Copyable : ICopyable
...
-            var requestStatus = new RequestStatus();
-            requestStatus.CopyFrom(new Copyable());
+            yield return new TestCaseData(new RequestStatus(), null).SetName("Request status CopyFrom() null");
+            yield return new TestCaseData(new RequestStatus(), new Copyable()).SetName("Request status CopyFrom() non-IRequestStatus type");
+        }
+
+        [Test, TestCaseSource(nameof(CopyFrom_InvalidTestCases))]
+        public void CopyFromTests(RequestStatus incoming, ICopyable copyable)
+        {
+            incoming.CopyFrom(copyable);
```
rianjs, line 175: I haven't evaluated the `CopyFrom` implementations, so I don't trust them. Leave it for now, though.

v2/ical.NET.UnitTests/DataTypes/RequestStatusTest.cs
```diff
@@ -1,123 +1,127 @@
...
 
-        [Test]
-        public void ToString_ValidInputWithNullStatusCode()
+        private static IEnumerable<ITestCaseData> ToStringTestCases() 
         {
-            var requestStatusA = new RequestStatus();
-            var requestStatusB = new RequestStatus()
+            var emptyRequestStatus = new RequestStatus();
+            yield return new TestCaseData(emptyRequestStatus, ";")
+                .SetName("ToString() for empty request status should return semicolon");
```
rianjs: I think you've written unit tests to assert the _current_ behavior instead of the _correct_ behavior. Why would an empty `RequestStatus.ToString()` return anything other than null or `string.Empty`?

I think this test should assert that an empty request status returns `null`. I know it won't pass, so let's mark it `Ignore`d.

net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
ie-zero: I definitely see your point. I obviously have not thought it through.

v2/ical.NET.UnitTests/DataTypes/RequestStatusTest.cs
ie-zero: You are 100% Rian. I wrote the unit tests to validate the current behavior assuming it is the correct one as I am still not familiar enough with the iCalendar RFC specification to know differently. 

By adding the unit tests we _might_ be able to bring into light omissions or errors which can be otherwise hard to spot. 

I will be happy to change the unit tests as you see fit.

net-core/Ical.Net/Ical.Net.UnitTests/DataTypes/RequestStatusTest.cs
ie-zero: I will do.

ie-zero:
@rianjs: Would you like to keep the ```CopyFrom``` tests or you would rather remove them until you validate the functionality?

At the moment, the ToString() is expected to return null for an empty ```RequestStatus```. As iCalendar is mostly about strings I would propose to return string.Empty instead as it is makes chaining cleaner in my opinion. 

It really does not matter which one you decide to be used as long as we keep it consistent across throughout the library.

PR #267, opened 2017-04-17
Add unit tests for StatusCode class

ie-zero:
* Add unit tests for StatusCode class
* Fix: Ensure Parts property is initialised in all cases otherwise some of the methods throw unexpected exceptions.


ie-zero:
I did not include any tests for negative integer numbers for Parts property. At the moment, are accepted as normal.

@rianjs: What would be your preference ...
1. accept negative part numbers and handle them as normal?
2. reject them and throw exceptions in the constructor?
3. treat them as zeros?
4. do nothing? 

I vote for option 3. It is relatively easy to implement but as far as I can see in the standard there is no need to handle negatives specifically.

rianjs:
Just as a general note, @ie-zero -- I have a lot on my plate right now, but I'll get to some of your more involved PRs this week, hopefully.

ie-zero:
I completely understand. We all have other work or family priorities.

PR #269, opened 2017-04-17
Move images and update README

ie-zero:
* Link to RFC 5545 specification.
* Add a short description about iCalendar specification.
* Move images into assets/ folder reduce the number of files on the main directory.


rianjs:
I'm OK with the asset reorganization. I'm not OK with the change from `ical.net` to `iCal.NET` in the readme.

I may make that change in the future, but I want to think about it a bit first. (I really hate the lowercase `i` prefixed nonsense that Apple started. It also makes following the Framework Design Guidelines' naming conventions ugly. `ICalendar` that's not an interface. Ugh. `IICalendarInterfaceThing`? Terrible.)

ie-zero:
I completely agree with you that the iThing is ugly for the reasons you mentioned but all I did was make uniform across the front page (i.e. README) on how we refer to the product. 

One proposal is to rename the software cal.NET or ical.NET or even a random name

It goes without saying that I will be happy to revert back the name formatting until there is another plan.

Issue #253, opened 2017-03-23
[.NET Core] NullReferenceException during Event serialization


remyjette:
(Forked from #23)

Creating a simple calendar with a single calendar event (basically the sample in https://github.com/rianjs/ical.net/wiki without the recurrence rules) in a .NET Core console app fails to serialize.

I did some debugging, looks like the `actualEvent.Properties.Remove("DURATION")` in `EventSerializer.cs` removes the DURATION from `_dictionary`, but not from `_lists`, causing the `_lists` entry to be set to null.
Since `_lists` is what is used to create the enumerator, this null entry is included when using LINQ extension methods.

This means that during serialization, in `ComponentSerializer.SerializeToString()` the following throws a NullReferenceException because p is null:

    var properties = c.Properties.OrderBy(p => p.Name).ToList();

Sample code to reproduce is below, using Microsot.NETCoreApp 1.1.1:

    var now = DateTime.Now;
    var later = now.AddHours(1);
    var e = new CalendarEvent
    {
        DtStart = new CalDateTime(now),
        DtEnd = new CalDateTime(later)
    };
    var calendar = new Calendar();
    calendar.Events.Add(e);
    var serializer = new CalendarSerializer(new SerializationContext());
    var serializedCalendar = serializer.SerializeToString(calendar);

Mfron-42:
@remyjette  what is your solution ?

remyjette:
I mean, a quick hack would be to change

    var properties = c.Properties.OrderBy(p => p.Name).ToList();

to

    var properties = c.Properties.Where(p => p != null).OrderBy(p => p.Name).ToList();

But I'm not planning on submitting a PR to change that line because it's just that - a hack. The underlying issue is that when `GroupedListEnumerator` is enumerating the values of the `CalendarPropertyList` that contains the properties of the event, it's including a `null` for the removed `DURATION` property rather than skipping it. I haven't yet had time to dig into this further, but that's my findings so far. Hopefully I'll have soon have some more time to investigate this so I can submit a PR with an actual fix rather than the hack I mentioned above.

Mfron-42:
Ok perfect thx, im waiting for the next release ;)

Ps : i suggest maybe : var properties = c.Properties.OrderBy(p => p?.Name).ToList ();
 
If the problem is just this line ? Just because im loving this operator <3

Mfron-42:
Ps : we can 

var properties = c.Properties.OrderBy(p => p?.Name).ToList ();

[...]

PR #270, opened 2017-04-22
Keep data structures in sync in GroupedList.Remove() #253

remyjette:
`GroupedList.Remove()` only removes the `TGroup` from `_dictionary`, not `_lists`.
This causes a problem during enumeration because the `GroupedValueEnumerator`
is created from the `_lists` object. Depending on the method of enumeration
used, it might enumerate the removed item (meaning it's still in the results
despite the call to `Remove()`) or it might be null in the enumerated list,
either way causing a problem.

Since `GroupedList.EnsureList()` adds to both `_lists` and `_dictionary`, this
change causes `GroupedList.Remove()` to mirror this behavior by removing
from both instead of just `_dictionary`.

### DIFF
--- a/net-core/Ical.Net/Ical.Net.Collections/GroupedList.cs
+++ b/net-core/Ical.Net/Ical.Net.Collections/GroupedList.cs
@@ -154,6 +154,11 @@ namespace Ical.Net.Collections
                 return false;
             }
 
+            if (!_lists.Remove(_dictionary[group]))
+            {
+                return false;
+            }
+
             return _dictionary.Remove(group);
 
             //var list = _dictionary[group];


rianjs:
* Does a symmetrical change need to be made in `v2`, or is it only `net-core` that's affected?
* Can you write a unit test that fails pre-change, and succeeds post-change?

jordanbrowneb:
Any way we can get this PR merged in? This fixes a show-stopping bug for .NET core users.

jchannon:
I've forked and applied this fix and released a nupkg to our internal nuget server because I need it now 😄  Will obviously use the public nuget when this is fixed

rianjs:
OK, wow, this file has _way_ diverged from `v2`, and has a bunch of commented-out code. (Ew.) I'll merge it as-is. I have no idea what unit tests will have broken as a result of my apparently-incomplete refactoring.

rianjs (approved):

rianjs:

jordanbrowneb:
Thanks for the timely response @rianjs

Issue #280, opened 2017-05-02
PeriodList should implement IList


rianjs:
The .NET Core version of PeriodList already implements IList<Period>. We should do the same thing in v2, because it's a List<IPeriod> under the hood already anyway, and we get additional power by doing so.

rianjs:
Published in Ical.Net 3.0.9-net-core-beta:
https://www.nuget.org/packages/Ical.Net/

PR #281, opened 2017-05-02
PeriodList now implements IList<IPeriod> #280

rianjs:


rianjs:

PR #286, opened 2017-05-29
Deserialize Google Calendar comment for Attendee

kajaj:
I am using iCal.Net for deserializing .ics files sent across email. However I have noticed, that even though the RFC 2445 has a field Comment (which Outlook uses), Google decided to add a special "X-RESPONSE-COMMENT" field to the attendee. Since I need that comment, I have added it in the deserializator.


rianjs:
@kajaj - I appreciate your enthusiasm for closing a gap that you found. Unfortunately this isn't the right way to handle non-standard icalendar properties. RFC-5545 [defines non-standard properties](https://tools.ietf.org/html/rfc5545#section-3.8.8.2) as being prefixed with `X-`. This allows application developers to insert their own properties without each one having to be approved, and so library developers don't have to individually support every property some developer could imagine.

You can already make use of non-standard properties with ical.net. Here's an example:

```csharp
var calendarEvent = new Event
{
    Start = new CalDateTime(_now),
    End = new CalDateTime(_now.AddHours(1)),
    Properties =
    {
        new CalendarProperty("X-RESPONSE-COMMENT", "Some value")
    },
};

var retrievedProperty = calendarEvent.Properties
    .FirstOrDefault(p => p.Name.Equals("X-RESPONSE-COMMENT"));
```

rianjs:
I've created a wiki page about non-standard properties:
https://github.com/rianjs/ical.net/wiki/Working-with-non-standard-(X-prefixed)-properties

PR #290, opened 2017-06-08
Make NetCore PeriodList implement IsReadOnly

jchannon:
NetCore PeriodList does not compile as it doesn't implement IsReadOnly property.


rianjs (approved):
Thanks! No idea how this snuck through!

rianjs:

Issue #293, opened 2017-06-08
Combine Framework and Core versions of the library into a single project


rianjs:

rianjs:
Published in 3.0.11-net-core-beta

https://www.nuget.org/packages/Ical.Net/

PR #294, opened 2017-06-08
netstandard1.3 and net46 #293

rianjs:
Per the API surface area, net46 and netstandard1.3 overlap. This is a change from netstandard1.6 which I thought was a little aggressive.

https://docs.microsoft.com/en-us/dotnet/standard/library

I also updated some dependencies (NodaTime, etc.), and I did remove one section of unreachable code.

### DIFF
--- a/net-core/Ical.Net/Ical.Net.Collections/GroupedList.cs
+++ b/net-core/Ical.Net/Ical.Net.Collections/GroupedList.cs
@@ -153,13 +153,6 @@ namespace Ical.Net.Collections
                 list.RemoveAt(i);
             }
             return true;
-
-            if (!_lists.Remove(_dictionary[group]))
-            {
-                return false;
-            }
-
-            return _dictionary.Remove(group);
         }
 
         public virtual bool Contains(TItem item)
--- a/net-core/Ical.Net/Ical.Net.Collections/Ical.Net.Collections.csproj
+++ b/net-core/Ical.Net/Ical.Net.Collections/Ical.Net.Collections.csproj
@@ -1,6 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
-    <TargetFramework>netstandard1.4</TargetFramework>
+    <TargetFrameworks>netstandard1.3;net46</TargetFrameworks>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <OutputType>library</OutputType>
+  </PropertyGroup>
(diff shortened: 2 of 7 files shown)


rianjs:
I would appreciate @thoemy and @jchannon taking a look as well.

net-core/Ical.Net/Ical.Net.Collections/GroupedList.cs
```diff
@@ -159,13 +159,6 @@ public virtual bool Remove(TGroup group)
                 list.RemoveAt(i);
             }
             return true;
-
```
jchannon, line 156: Didn't we just add that yesterday?

net-core/Ical.Net/Ical.Net.Collections/Ical.Net.Collections.csproj
```diff
@@ -1,6 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
-    <TargetFramework>netstandard1.4</TargetFramework>
+    <TargetFrameworks>netstandard1.3;net46</TargetFrameworks>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <OutputType>library</OutputType>
```
jchannon: This can be deleted, don't need library defined

rianjs: Thanks, I'm not very familiar with having a single project that builds core and framework versions.

I think it's also possible to move the nuspec contents into the csproj. A task for another day, perhaps.

jchannon: Yup you can do that however bear in mind the default behaviour when you run `dotnet pack` is to list the dependencies as separate packages so you'd need to push the collections etc libraries separately. That is one large change compared to full .net fwk stuff. There is a way around it by including dlls in the nupkg like you currently do with the nuspec but i'm not sure its advised.

thoemy:
I'm not a packaging expert. But wouldn't it be enough to just target netstandard1.3 for the antlr and Collections projects?

jchannon:
Yup good point net46 ical could target netstandard1.3 packages such as the collections

rianjs:
Published in 3.0.11-net-core-beta

net-core/Ical.Net/Ical.Net.Collections/GroupedList.cs
jchannon: Can you confirm this removal hasn't broken serialization?

PR #296, opened 2017-06-12
Reverted TZID parameter in UTC Dates. 

TwistyPop:
DateTime Serializer now places a Zulu suffix at the end of the string when dates are UTC.


TwistyPop:
Side note, after I pulled your branch several classes had the namespace casing convention messed up. I had to fix that for it to build properly.

rianjs (approved):

PR #302, opened 2017-06-21
Update Constants.cs

jgshumate1:
Added Quarterly (4 times per year)


rianjs:
Quarterly isn't part of the spec:
https://tools.ietf.org/html/rfc5545#section-3.3.10

PR #306, opened 2017-07-13
Enhanced Occurence Comparison for GetOccurences

maxmanzero:
Updated the IRecurrable interface with IUniqueComponent. Update Occurence objects Equals and GetHashCode methods to leverage the UID property exposed on the Source property so the comparison routine in GetOccurences can distinguish between Occurences that look equal but have different underlying event data.


andrea-spotsoftware:
I've found the same problem with GetOccurrences.
I was writing some unit test and i found out that 2 events with the same attributes (startDate, endDate, recurrenceRule) but different Uid, are treated as duplicates by the getOccurrences method.
As a temp workaround, i've also specified the Uid in the summary 😞

minichma:
I assume the problem has been fixed in the meantime. If you think it hasn't, please feel free to reopen this. The project is being more vital these days.

PR #307, opened 2017-07-26
Two dates (at the same moment) with different time zones should be equal

rbuehlma:


rianjs:
`CalDateTime`s with different time zones are NOT equal, even if their _occurrence_ results at the same UTC instant.

`CalDateTime` expresses a time _in a place_, which is important because recurrences based off of them will have different rules, and the UTC instants will not always overlap. Time zones are absolutely significant when doing these comparisons. If you want to compare _instants_, you'll need to do it in your application layer.

rbuehlma:
The issue was inside the library itself. If I remember correctly, it was with exceptions having a different zone and were therefore not applied as the dates where not "equal" even though the same UTC instant was used.

Do you have an other way to handle that?

rianjs:
There are some edge cases where equality isn't perfect. Basically I would like:

* Equality expressed in terms of the type hierarchy (CalDateTimes derive from other types, and should call the parent's `Equals` method instead of re-implementing the hierarchy internally)
* Sort `RDATE` and `EXDATE` `CalDateTime`s by time zone, both internally and when serializing and deserializing. When the recurrence set is computed (`GetOccurrences()`), it should drop `Occurrence`s out (if excluded via EXDATE), or add them in (if included via RDATE).

Right now, UTC is sometimes used as shorthand for asking about equality, and for most use cases, this works OK, because people are rarely crossing time zones.

I think your idea here is covered by the second bullet: you want to drop `Occurrence`s when `Europe/Berlin` overlaps with `America/New_York` (or whatever) for those times when there's overlap. And those overlapping instants are easiest to express as UTC because they're an apples-to-apples comparison, even though the underlying rules that generated those `Occurrence`s are different.

I think this idea should live in the `Occurrence` object somewhere, but doing so isn't straightforward given the complex type hierarchy and sometimes-weird, often-circular type dependencies.

PR #309, opened 2017-07-31
Equals on the CalendarEvent should only be true if summary and description match

rbuehlma:


rianjs (approved):
In the future, can you make symmetrical changes to both `net-core` and `v2`? I'll add the change to the `v2` release after I merge this.

rianjs:
Oh, and you should never change `Equals` without also making the same change in `GetHashCode` and vice-versa.

That's true in _all_ of the code you write.

rbuehlma:
ok, will update both versions of the library in the future.

for the GetHashCode: I made the Equals more strict. GetHashCode needs to be the same if Equals true. If Equals is false, GetHashCode may still be the same. This should still be the case. But yes, the HashCode could be improved by also adding the fields there.

Oh and thanks for merging :)

PR #310, opened 2017-08-19
Implemented missing Journal constructor so Initialize method is called.

HitTheDrum:
If not intended, the constructor of the Journal class was missing and its Initialize() Method was not called which led to "Component: null" object when instantiated.

### DIFF
--- a/v2/ical.NET/Components/Event.cs
+++ b/v2/ical.NET/Components/Event.cs
@@ -68,7 +68,7 @@ namespace Ical.Net
                 if (!Equals(DtEnd, value))
                 {
                     Properties.Set("DTEND", value);
-                    ExtrapolateTimes(nameof(DtEnd));
+                    ExtrapolateTimes();
                 }
             }
         }
@@ -102,7 +102,7 @@ namespace Ical.Net
                 if (!Equals(Duration, value))
                 {
                     Properties.Set("DURATION", value);
-                    ExtrapolateTimes(nameof(Duration));
+                    ExtrapolateTimes();
                 }
             }
         }
@@ -266,36 +266,20 @@ namespace Ical.Net
             ExtrapolateTimes();
         }
 
-        /// <summary>
-        /// Flag to check if ExtrapolaTimes is currently invoking to prevent multi extrapolating
-        /// </summary>
-        private bool _isExtrapolating;
-        private void ExtrapolateTimes(string propertyName = null)
+        private void ExtrapolateTimes()
         {
-            if (_isExtrapolating) return;
-            _isExtrapolating = true;
-
             if (DtEnd == null && DtStart != null && Duration != default(TimeSpan))
             {
-                if (propertyName == nameof(DtEnd))
-                    Duration = TimeSpan.Zero;
-
-                if (propertyName != nameof(DtEnd))
-                    DtEnd = DtStart.Add(Duration);
+                DtEnd = DtStart.Add(Duration);
+            }
+            else if (Duration == default(TimeSpan) && DtStart != null && DtEnd != null)
+            {
+                Duration = DtEnd.Subtract(DtStart);
             }
             else if (DtStart == null && Duration != default(TimeSpan) && DtEnd != null)
             {
                 DtStart = DtEnd.Subtract(Duration);
             }
-            else if (/*Duration == default(TimeSpan) && */ DtStart != null && DtEnd != null)
-            {
-                if (propertyName == nameof(Duration))
-                    DtEnd = DtStart.Add(Duration);
-                else
-                    Duration = DtEnd.Subtract(DtStart);
-            }
-
-            _isExtrapolating = false;
         }
 
         protected bool Equals(Event other)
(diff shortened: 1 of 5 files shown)


rianjs:
Can you get rid of Initialize(), and move that logic to the constructor?

And then implement the change in net-core as well?

HitTheDrum:
Initialize(): Initialize() is also called in OnDeserializing. However atm Initialize() is a one-liner (Name = Components.Journal;). Would you still prefer to move the logic to both (constructor and OnDeserializing) or keep it in the existing Initialize() method? Or just remove from OnDeserializing as this should not be necessary?

net-core: no problem! I will adjust it then too.

rianjs:
> Initialize(): Initialize() is also called in OnDeserializing. However atm Initialize() is a one-liner (Name = Components.Journal;). Would you still prefer to move the logic to both (constructor and OnDeserializing) or keep it in the existing Initialize() method? Or just remove from OnDeserializing as this should not be necessary?

I think it's safe to move it to the constructor. It's just setting properties, and by rights, this is a constant property for the type. In fact, it could probably be real `const`. In the meantime, though, the constructor is fine.

v2/ical.NET/Components/Event.cs
```diff
@@ -266,20 +266,36 @@ protected override void OnDeserialized(StreamingContext context)
...
+            _isExtrapolating = true;
+
             if (DtEnd == null && DtStart != null && Duration != default(TimeSpan))
             {
-                DtEnd = DtStart.Add(Duration);
-            }
-            else if (Duration == default(TimeSpan) && DtStart != null && DtEnd != null)
-            {
-                Duration = DtEnd.Subtract(DtStart);
+                if (propertyName == nameof(DtEnd))
```
rianjs: Please use braces for conditionals.

v2/ical.NET/Components/Event.cs
```diff
@@ -266,20 +266,36 @@ protected override void OnDeserialized(StreamingContext context)
...
+                    Duration = TimeSpan.Zero;
+
+                if (propertyName != nameof(DtEnd))
+                    DtEnd = DtStart.Add(Duration);
             }
             else if (DtStart == null && Duration != default(TimeSpan) && DtEnd != null)
             {
                 DtStart = DtEnd.Subtract(Duration);
             }
+            else if (/*Duration == default(TimeSpan) && */ DtStart != null && DtEnd != null)
```
rianjs: Please don't commit commented-out code.

v2/ical.NET/Components/Event.cs
```diff
@@ -266,20 +266,36 @@ protected override void OnDeserialized(StreamingContext context)
             ExtrapolateTimes();
         }
 
-        private void ExtrapolateTimes()
+        /// <summary>
+        /// Flag to check if ExtrapolaTimes is currently invoking to prevent multi extrapolating
+        /// </summary>
+        private bool _isExtrapolating;
```
rianjs: Does this solve a problem that you have a test case for?

HitTheDrum:
Moved Initialize() logic to constructor and removed reduntant OnDeserializing() in
- .\ical.net-master\net-core\Ical.Net\Ical.Net\Components\Journal.cs
- .\ical.net-master\v2\ical.NET\Components\Journal.cs
(Please just ignore rev 2971e35 (reverted with 3aaf539) as it was not intended for this pull request)

v2/ical.NET/Components/Journal.cs
```diff
@@ -14,19 +14,17 @@ public JournalStatus Status
...
+        /// (usually an iCalendar object) as its parent.
+        /// </summary>
+        public Journal()
         {
             Name = Components.Journal;
         }
 
         protected override bool EvaluationIncludesReferenceDate => true;
 
-        protected override void OnDeserializing(StreamingContext context)
```
rianjs: I seem to recall that deleting these overrides can cause subtle problems, so I think these should be put back.

I'm sorry my other comment was unclear: I meant delete the call to `Initialize()` from this method, moving that logic to the constructor, which you've done. Can you put back this override? Then it will be ready to merge.

rianjs (commented):
Does this solve a problem you have? I.e., do you need me to publish a nuget package version that incorporates these change?

If so, can you:
* Update the `readme.md` file, and explain quickly what problem was solved with these changes.
* Bump the minor version in the respective nuspec files in `v2` and `net-core`?

v2/ical.NET/Components/Event.cs
HitTheDrum: Understood!

v2/ical.NET/Components/Event.cs
HitTheDrum: Ok!

v2/ical.NET/Components/Event.cs
HitTheDrum: I will open another Pull Request incl. description to separate this from the current request.

v2/ical.NET/Components/Journal.cs
HitTheDrum: Reverted removal of OnDeserializing()
(see 7a3f471)

rianjs (approved):

Issue #311, opened 2017-08-23
SerializeToString throw InvalidOperationException without PRODID


ewanzd:
I programming a CalDAV server. There is a case, where it is conform to remove the PRODID and then serialize to string: [rfc4791 section-7.8.1](https://tools.ietf.org/html/rfc4791#section-7.8.1)
But `SerializeToString` of `CalendarSerializer` throw a `InvalidOperationException`.

I wrote a little function to show this case:

`var calendar = new Calendar();
calendar.ProductId = "-//MyProduct";
var prodId = calendar.Properties["PRODID"];
calendar.Properties.Remove(prodId);

var serializer = new CalendarSerializer(calendar);
var output = serializer.SerializeToString();`

poulerik-dk:
@https://tools.ietf.org/html/rfc5545#section-3.7.3 says that this would not be a valid iCalendar object. 

If you want to produce an invalid iCalendar you could implement it as:

    var serializer = new ComponentSerializer();
    var output = serializer.SerializeToString(calendar);

ewanzd:
Then do you assume rfc4791 section-7.8.1 is wrong? In a perspective the property is available in iCalendar object, but not included in the limited response.

Thanks for the code snippet

poulerik-dk:
I don't know to be honest. But I see the point of not relaying PRODID, when only partial data is requested. Don't think of me as an authority on this matter though ;)

rianjs:
[`PRODID` is a required property per the spec.](https://tools.ietf.org/html/rfc5545#section-3.7.3)

PR #312, opened 2017-08-23
Check if collection is empty before trying to manipulate first element.

poulerik-dk:
Fixes the InvalidOperationException mentioned in #311.

I assume that it is the correct behaviour, that it simply reappends the product id property.


rianjs (approved):
In the future, can you make a parallel change to `v2`?

Also, you can express this more concisely:

```csharp
if (ContainsKey(group))
{
    AllOf(group)?.FirstOrDefault()?.SetValue(values);
    return;
}
```

I'll merge and make the parallel change in v2, and tighten up the expression of the idea.

Thanks for the fix!

PR #316, opened 2017-09-19
Roll up prs

rianjs:


rianjs:
Published in nuget package versions `2.3.3` and `2.3.3`

PR #322, opened 2017-10-17
Catch case where recurring item has no Start

rbuehlma:
In my case, the object without a Start was a TODO. The backend is Zimbra. I'm not sure if it is a Bug in Zimbra or if it is actually a valid item but the fix is simple in the library.

### DIFF
--- a/net-core/Ical.Net/Ical.Net.nuspec
+++ b/net-core/Ical.Net/Ical.Net.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
     <metadata>
         <id>Ical.Net</id>
-        <version>3.0.12</version>
+        <version>3.0.13</version>
         <title>Ical.Net</title>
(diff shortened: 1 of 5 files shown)


rianjs (commented):
Yeah, it looks like the only required property in `VTODO`s is the `DTSTAMP`:

https://tools.ietf.org/html/rfc5545#section-3.6.2

Can you make the same change in the .NET 2.x version as well in the `v2` folder? Should be in the same place.

rbuehlma:
Sorry, missed the 2.x version again :-S done now.

rianjs (commented):
Cool. Two last things:
* Can you edit the nuspec files in both the `v2` and `net-core` versions and bump the patch version by 1?
* Then add a short blurb to `release-notes.md` explaining the change with the associated major.minor.patch version. (You'll see what I mean when you open the file.) Something like "`DTSTART` is not required for `VTODO` components." would be sufficient.

Then I'll merge and publish new versions to nuget.

rbuehlma:
Done as well.

rianjs (approved):

PR #328, opened 2017-10-23
Ignore timezone if two periods are compared for equality

rbuehlma:
This solves the case where exception dates are ignored due to different time zone than the recurring event.

This PR is a follow up to https://github.com/rianjs/ical.net/pull/307.

On IDateTime there were cases where it was required to honour the timezone in equals.
Therefore, this PR changes the Equals on "Period" which also solves the issue. Are there also cases where the Timezone for a Period is relevant in case of comparing them? I currently cant think of one and therefore think this could be a nice fix for my issue.

What is your opinion on that?

If its ok, I will also update the v2-version and do the package updates.

### DIFF
--- a/net-core/Ical.Net/Ical.Net/DataTypes/Period.cs
+++ b/net-core/Ical.Net/Ical.Net/DataTypes/Period.cs
@@ -60,7 +60,7 @@ namespace Ical.Net.DataTypes
             Duration = p.Duration;
         }
 
-        protected bool Equals(Period other) => Equals(StartTime, other.StartTime) && Equals(EndTime, other.EndTime) && Duration.Equals(other.Duration);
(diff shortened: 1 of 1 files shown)


minichma:
I think two Periods shouldn't be treated as equal if their tzIds don't match. The tzId is clearly part of the object's identity. Moreover, the instant to which a Period's startTime resolves is connot be determined by the Period's state alone. It rather requires evaluation of the timezone itself, which in turn can change over time.

If the intention is to receive instants only once when returned by `Calendar.GetOccurrences()`, then the right approach would probably be to return all occurrenes in the same timezone, i.e. in the one of DTSTART or maybe UTC. Then equality comparison would work as expected.

PR #333, opened 2017-11-13
Until serialization

rianjs:


rianjs:
Published in nuget version 4.0.0: https://www.nuget.org/packages/Ical.Net/4.0.0

PR #364, opened 2018-02-23
Infiinte Loop with MaxDate for GetOccurrences

dsbegnoche:
Fixes bug when GetOccurrences ends on MaxDate and only Count is set, an
error is thrown inside an infinite loop

### DIFF
--- a/net-core/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
+++ b/net-core/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


net-core/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -259,11 +259,21 @@ private void EnforceEvaluationRestrictions(RecurrencePattern pattern)
                     break;
                 }
 
+                if (pattern.Until != DateTime.MinValue && candidate != DateTime.MinValue && candidate > pattern.Until)
```
pomoika: This condition is exact duplicate of the above statement, therefore it should be removed.

dsbegnoche: Good catch.  This was an accident.  Removed the duplication.  @pomoika

mscrivo:
Running into this as well after upgrading from 3.0.6 to 4.1.2 .. any reason this hasn't been merged in yet?

dsbegnoche:
@rianjs Any chance we can get this reviewed and merged?  This bug is preventing us from upgrading to latest version.

rianjs:

Issue #330, opened 2017-11-02
CalDateTime.ToTimeZone produces UTC time instead of target time zone


pomoika:
It looks like CalDateTime.ToTimeZone always creates CalDateTime in UTC even though internal conversion between source and destination time zones has been done properly (e.g. from "Central Standard Time" to "Eastern Standard Time"). The result is represented in UTC, whereas I would expect it in desired time zone. The same issue occurs in DateUtil.MatchTimeZone method.

rianjs:
Could you provide a code example demonstrating the problem?

pomoika:
Here is a simple sample:
```
            const string source = @"
BEGIN:VCALENDAR
PRODID:-//github.com/rianjs/ical.net//NONSGML ical.net 2.2//EN
VERSION:2.0
BEGIN:VEVENT
DTEND;TZID=Central Standard Time:20171102T093000
DTSTART;TZID=Central Standard Time:20171102T090000
RRULE:FREQ=WEEKLY;COUNT=4;BYDAY=MO
SEQUENCE:0
END:VEVENT
END:VCALENDAR";
            ICalendar cal = Calendar.LoadFromStream(new StringReader(source)).First();
            IDateTime start = cal.Events[0].Start;
            Console.WriteLine(start.ToString());
            IDateTime eastern = start.ToTimeZone("Eastern Standard Time");
            Console.WriteLine(eastern.ToString());
```
11/2/2017 9:00:00 Central Standard Time
11/2/2017 14:00:00 UTC

rianjs:
Fixed in nuget version 4.0.0: https://www.nuget.org/packages/Ical.Net/4.0.0

afterlogic-support:
Those who are not on .NET Core are out of luck to get this fix?

rianjs:
v4 is .NetStandard 1.3, which means you can use it with .NET 4.6 and higher.

afterlogic-support:
My target is .NET 4.5. Anyway, I took the fixed code from net-core branch and made it work as a separate util method (so I'm not using ToTimeZone at all now). Can live with that for now. Thanks!

rianjs:
In a perfect world, I would support .NET 4.5+, but there were reasons (that I can't recall now) why .NetStandard < 1.3 wouldn't work.

PR #368, opened 2018-03-10
V2 fixes for RecurrenceUtil.GetOccurrences and CalDateTime.ToTimeZone

pomoika:
This pull request is intended to resolve several Ical.Net v2 issues:
* CalDateTime.ToTimeZone produces UTC time instead of target time zone (bug #330)
* Fix issues with UNTIL recurrence rule that caused RecurrenceUtil.GetOccurrences not returning the last occurrence from the series in v2.
* Update NodaTime to v1.3.6

@rianjs, even though Ical.Net v2 is no longer officially supported, I would be grateful if you could review and merge this pull request and then release new version 2.3.6 if there are no objections. We still use .NET 4.0 and can't migrate to new Ical.Net v4.

Thanks.

### DIFF
--- a/v2/ical.NET.UnitTests/EqualityAndHashingTests.cs
+++ b/v2/ical.NET.UnitTests/EqualityAndHashingTests.cs
@@ -259,12 +259,12 @@ namespace Ical.Net.UnitTests
             Assert.IsFalse(serialized.Contains("Hello"));
         }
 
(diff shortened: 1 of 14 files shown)


minichma:
Cannot comment, why this hasn't been merged back then, but probably isn't relevant anymore.

Issue #372, opened 2018-03-30
CalDateTime equality operators handle null-values incorrectly.


minichma:
`CalDateTime.operator ==` and `CalDateTime.operator !=` fail when comparing against null. This implies, that natural expressions like `((x == null) || (x <= y))` fail. This certainly violates the intended contract for == and !=. It probably should be implemented similar to:

```cs
public static bool operator ==(CalDateTime left, IDateTime right) => 
    (ReferenceEquals(left, null) || ReferenceEquals(right, null))
    ? ReferenceEquals(left, right):
    left.Equals(right);

public static bool operator !=(CalDateTime left, IDateTime right) =>
    (ReferenceEquals(left, null) || ReferenceEquals(right, null))
    ? !ReferenceEquals(left, right) :
    !left.Equals(right);
```

Other comparison operators should probably handle null values gracefully too. Here is, how [Nullable\<T\>](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/using-nullable-types) does it:
> When you perform comparisons with nullable types, if the value of one of the nullable types is null and the other is not, all comparisons evaluate to false except for != (not equal).

rianjs:
Looks reasonable. I’ll merge a stylistically tweaked version of this this week.

rianjs:
Available in nuget version 4.1.5:
https://www.nuget.org/packages/Ical.Net/4.1.5

PR #373, opened 2018-04-02
Fix for #372: Make CalDateTime comparison operators null-safe

minichma:
This PR makes the CalDateTime comparison operators null-safe. This is a fix for issue #372.


rianjs:
Available in nuget version 4.1.5:
https://www.nuget.org/packages/Ical.Net/4.1.5

Issue #378, opened 2018-05-03
Wrong offset selected for timezones which don't observe DST and have had their UTC offset increased since the beginning of time


mfields3:
When a timezone is added which does not observe Daylight Savings Time, the minimum UTC offset which was ever used for the timezone is used which can be incorrect if the timezone has ever increased the UTC offset it uses.

For example:
- A calendar with a timezone of "America/Cayman" reports the offset as -05:19:36 (which was from 01/01/1890 - 22/04/1908). Since 22/04/1908, the correct offset is -05:00:00.
- A calendar with a timezone of "Asia/Singapore" reports the offset as 06:55:25 (which was from the beginning of time - 31/05/1905). Since 31/12/19818, the correct offset is 08:00:00.

```
string timezone = "Asia/Singapore";
var iCal = new Ical.Net.Calendar();
iCal.AddTimeZone(timezone);
Console.WriteLine("The reported offset is: " + ((VTimeZoneInfo)iCal.TimeZones[0].Children[0]).OffsetTo);
```

The issue is because NodaTime treats the current timezone interval as not having an end time. Code at `VTimeZone.cs:47` specifically excludes timezone intervals which don't have an end time incorrectly.

rianjs:
Will be available in nuget version 4.1.5 in the next hour or so:

https://www.nuget.org/packages/Ical.Net/4.1.5

PR #379, opened 2018-05-03
Fix for #378 - handle timezones which do not observe DST

mfields3:
Noda Time doesn't specify End values for the current timezone interval for timezones which do not observe DST.
This causes no current interval to be returned, and the existing code to create at least one standard interval incorrectly - it selects the minimum offset from the timezone on any interval which may not be correct for today.
This fix allows the correct current interval to be returned, rather than relying on fallback code.


rianjs:
Looks good. Will merge this week.

rianjs (approved):

Issue #390, opened 2018-05-21
CalDateTime.Initialize overrides DateTimeKind property on incoming DateTimes when it should not


rianjs:
- Specifying a tzid of "UTC" should ensure that `Value.Kind` is `DateTimeKind.Utc`
- Specifying a non-UTC tzid should ensure that `Value.Kind` is `DateTimeKind.Local`
- Everything else should leave the `Kind` property alone

Right now, the bug is that Anything that doesn't have an explicit, real, time zone is marked as `Unspecified` which is wrong. We shouldn't touch it in that case.

```csharp
[Test, TestCaseSource(nameof(DateTimeKindOverrideTestCases))]
public DateTimeKind DateTimeKindOverrideTests(DateTime dateTime, string tzId)
    => new CalDateTime(dateTime, tzId).Value.Kind;

public static IEnumerable<ITestCaseData> DateTimeKindOverrideTestCases()
{
    const string localTz = "America/New_York";
    var localDt = DateTime.SpecifyKind(DateTime.Parse("2018-05-21T11:35:33"), DateTimeKind.Local);

    yield return new TestCaseData(localDt, "UTC")
        .Returns(DateTimeKind.Utc)
        .SetName("Explicit tzid = UTC time zone returns DateTimeKind.Utc");

    yield return new TestCaseData(DateTime.SpecifyKind(localDt, DateTimeKind.Utc), null)
        .Returns(DateTimeKind.Utc)
        .SetName("DateTime with Kind = Utc and no tzid returns DateTimeKind.Utc");

    yield return new TestCaseData(localDt, localTz)
        .Returns(DateTimeKind.Local)
        .SetName("Local datetime with local tzid returns DateTimeKind.Local");

    yield return new TestCaseData(DateTime.SpecifyKind(localDt, DateTimeKind.Utc), localTz)
        .Returns(DateTimeKind.Local)
        .SetName("DateTime with Kind = Utc with explicit local tzid returns DateTimeKind.Local");

    yield return new TestCaseData(DateTime.SpecifyKind(localDt, DateTimeKind.Unspecified), localTz)
        .Returns(DateTimeKind.Local)
        .SetName("DateTime with Kind = Unspecified with explicit local tzid returns DateTimeKind.Local");

    yield return new TestCaseData(localDt, null)
        .Returns(DateTimeKind.Local)
        .SetName("DateTime with Kind = Local with null tzid returns DateTimeKind.Local");

[...]

PR #391, opened 2018-05-21
Only override DateTimeKind when we know for sure what it should be #390

rianjs:


rianjs:
https://www.nuget.org/packages/Ical.Net/4.1.2

PR #398, opened 2018-05-29
Net46

rianjs:


rianjs:
Available in nuget version 4.1.5:
https://www.nuget.org/packages/Ical.Net/4.1.5

Issue #408, opened 2018-07-05
PRODID not being respected after upgrade from v2 to v4


gregveres:
Hi, 

I just upgraded from v2 to v4. After getting the code to recompile and changing all the lower case values to upper case, I am left with one difference that is showing up in my unit tests. 

The ProductId of the calendar isn't being respected when serializing calendar events. I used to set the productId and serialize an event like this:
```
           var calendar = new Ical.Net.Calendar
            {
                ProductId = "-//www.skycourt.ca//RemoteAPI/EN"
            };
            calendar.Events.Add(ev);
            var serializer = new CalendarSerializer(new SerializationContext());
```

now that ProductId string comes out like this:
`PRODID:-//github.com/rianjs/ical.net//NONSGML ical.net 4.0//EN`

Is there a way to get the prodid to be set to the value I want?

gregveres:
Hmm I see that this is forced / hard coded in the serialization of a calendar on line 31 of https://github.com/rianjs/ical.net/blob/master/net-core/Ical.Net/Serialization/CalendarSerializer.cs

I am not sure what I think of that. The product is really our product, not the library that it uses. But I guess I can also see the other point of view....

rianjs:
See release-notes.md it was a very recent change.

That’s a perspective I hadn’t considered: that it might be an
implementation detail of a larger product...

gregveres:
I would probably recommend that if ProductId isn't set, then set it to your string. But if it is set, then the user wants to set it something that indicates their product. 

Is there a parsing benefit to having it set to your string?

tech-nik89:
Hi,

is this issue still being discussed? I've got the same problem here and would like to identify the output as my software instead of just the library.

Thanks a lot!

Franklin89:
Any news on this?

gonace:
Hi
Any news on this? is this something that are going to be supported? Otherwise, we'll be stuck on an old version forever.

Samelandslaget:
@rianjs You write in the release notes:

> 4.1.8 - (2018-06-12) - PRODID and VERSION are unconditionally set when serializing a calendar so as not to mislead consumers about where the serialized output came from. Previous behavior would use the values that came from deserialization. #403

This is all well and good as the default behavior, but what about the suggestion from @gregveres ?

> I would probably recommend that if ProductId isn't set, then set it to your string. But if it is set, then the user wants to set it something that indicates their product.

[...]

PR #410, opened 2018-07-17
Only override ProductId/Version if they arent already provided

CreepyGnome:
Fix for https://github.com/rianjs/ical.net/issues/408

### DIFF
--- a/net-core/Ical.Net.CoreUnitTests/SerializationTests.cs
+++ b/net-core/Ical.Net.CoreUnitTests/SerializationTests.cs
@@ -471,14 +471,27 @@ END:VEVENT";
             Assert.IsTrue(!until.EndsWith("Z"));
         }
 
-        [Test(Description = "PRODID and VERSION should use ical.net values instead of preserving deserialized values")]
-        public void LibraryMetadataTests()
+        [Test(Description = "PRODID and VERSION should retain ical.net values if they are provided")]
+        public void CalendarProductIdVersionTests()
         {
             var c = new Calendar
             {
                 ProductId = "FOO",
                 Version = "BAR"
             };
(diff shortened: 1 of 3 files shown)


jonathanbrecher (changes requested):
This is not a good change. Or not a complete change, anyway.

This change will satisfy the case where a calendar is created from scratch in code.

Consider the alternate case where you read an existing iCal with PRODID:BuggySoftware1.0.

Right after you imported that iCal, you should be able to look at the ProductId and see "BuggySoftware1.0" so you can know who created the original version. 

But ICal.Net does a bunch of normalizations internally, so if you immediately export the iCal, the newly exported version _must not_ be written with PRODID:BuggySoftware1.0 since it most likely will not be byte-for-byte identical to the original version written by that buggy software.

There needs to be some sort of notion of "ProductId/Version were set via code" versus "were imported from a file" versus "were unset". Only the "set via code" case would preserve the values as set. The other two cases would use the hardcoded ICal.Net identifier.

CreepyGnome:
This change is safe. I am not sure I understand your concerns.,

The issue is that ProdId and Version where being overwritten and forced to the iCal.net values. The only change here is if they are not set it will continue to work as it has been. The allows people to set the values without having to worry about them being overwritten. Which is the source of the issue for us, and how I understood the bug #408 have you read that?

This serialization is serializing from an object to a string.  Also, not sure why it matters if it came from code or was deserialized from another source. 

Anyone who is using the Internet Calendar format as specified by its RFC should know how and why to set ProdId and Version and that they should ensure it is set before serializing to a string. If they don't then they get the Ical.net defaults. 

I will leave it to @rianjs to review as it does resolve the bug that is currently open and which is one I have had to work around and would like to remove my workaround.

randbrown:
Any news on this?

PR #411, opened 2018-07-18
Associate attachments with their events when VALUE is BINARY.

chescock:
Without the association, parameters such as FMTTYPE would be lost.


rianjs (approved):

rianjs:
- bump the nuspec version in net-core directory
- build in release mode in visual studio
- update readme.md
- nuget pack in the net-core directory

(I’m in an Uber approving a pull request! We are truly living in the future!)

PR #420, opened 2018-10-13
deserialize STANDARD and DAYLIGHT timezone infos

dtrefilov-piterbyte:
The opposite to the https://github.com/rianjs/ical.net/pull/304
Provides the ability to populate the daylight and standard time zone information from the serialized ics.


rianjs:

Issue #159, opened 2016-10-17
Strong named assemblies


vijaydeep:
I am working with SharePoint and my project dll is deployed to GAC. And this requires all the dependencies, including Ical.Net.dll and Ical.net.collections.dll to be strongly named. Do you have strong named assemblies that I can use. 

I am using the Ical.net Nuget package version 2.3.1.

rianjs:
I do not have a strong named version, no. It IS a reasonable feature, though.

fvn-linagora:
There have been previous [discussions](https://github.com/miensol/SimpleConfigSections/issues/3#issuecomment-219826659) for similar strong naming requests in other OSS projects
[dsplaisted/strongnamer](https://github.com/dsplaisted/strongnamer) is one way to deal with this.

dolunay:
You must create a key.snk file from "Project Properties"->"Signing" page. check "Sign The Assembly" box and set password. Save the project and build it.

josteink:
@rianjs This is *really easy* to implement. If someone submitted a PR for this, would you be willing to publish this new, strong-named version?

rianjs:
https://www.nuget.org/packages/Ical.Net/2.3.6

rianjs:
https://www.nuget.org/packages/Ical.Net/4.1.10

josteink:
That's absolutely great and I'm sure will be appreciated.

Many thanks!

PR #435, opened 2019-01-30
Add strong-name signing to ical.NET assemblies.

josteink:
This closes the long standing issue #159 and enables use of ical.NET in enterprise deployments where strong-name signing is a requirement.


rianjs:
https://www.nuget.org/packages/Ical.Net/2.3.6

PR #437, opened 2019-01-31
Remove section about plans for v3 from readme

josteink:
V4.x is already current. Surely this is no longer needed?


rianjs (approved):

rianjs:
Indeed. Thanks!

PR #442, opened 2019-03-14
Update pdb format for framework and VSTS

eriknuds:
The VSTS didn't like the framework .pdb file, giving an error, similar to this command line test of validity:
---
# "C:\Program Files (x86)\Windows Kits\10\Debuggers\x86\srcsrv\pdbstr.exe" -r -s:ssrc -p:D:\src\packages\Ical.Net\lib\net46\Ical.Net.pdb
error 0xb opening D:\src\packages\Ical.Net\lib\net46\Ical.Net.pdb
---

This is fixed by adding some conditional debug settings in the project file.

Without this fix one must remember to delete the pdb file whenever one updates packages to avoid build errors, which is a nuisance.

### DIFF
--- a/net-core/Ical.Net/Ical.Net.csproj
+++ b/net-core/Ical.Net/Ical.Net.csproj
@@ -6,6 +6,8 @@
     <Company />
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\IcalNetStrongnameKey.snk</AssemblyOriginatorKeyFile>
+    <DebugSymbols Condition=" '$(TargetFramework)' == 'net46' ">true</DebugSymbols>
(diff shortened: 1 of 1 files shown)


eriknuds:
Please merge this, as this causes problems for people using VS Teams Services and .net framework...

The problem is similar to:
https://github.com/sendgrid/sendgrid-csharp/issues/494
...and the fix is similar to:
https://github.com/sendgrid/sendgrid-csharp/commit/c90111f3bdb62c239b38615e92583357f046a90d

Until merged, larger systems using .net framework and VSTS might experience issues with other teams or team members updating nuget packages and getting the incompatible .pdb file and thus breaking build and automatic deployments - quite annoying.

rianjs (approved):

rianjs:
https://www.nuget.org/packages/Ical.Net/4.1.11

Issue #432, opened 2019-01-04
GetOccurrences inconsistent results when end time is on start time for RRule.


jviau:
iCal.net version: 4.1.9

When you have a recurring vEvent, `GetOccurrences` on the first day of the event has an _inclusive_ end time. On subsequent days, the end time is _exclusive_. The code below highlights the issue. Checking on the first day yields 1 result when the end time is exactly on the schedule time. Adding 1 day to the checked time and checking again yields 0 results.

Code to reproduce issue:
``` CSharp
var rrule = new RecurrencePattern
{
    Frequency = FrequencyType.Daily,
    Until = DateTime.Today.AddMonths(4),
};

var vEvent = new CalendarEvent
{
    Start = new CalDateTime(DateTime.Parse("2019-01-04T08:00Z")),
};

vEvent.RecurrenceRules.Add(rrule);

var checkTime = DateTime.Parse("2019-01-04T08:00Z");
var occurrences = vEvent.GetOccurrences(checkTime.AddSeconds(-1), checkTime);
Console.WriteLine(occurrences.Count); // "1"

checkTime = checkTime.AddDays(1);
occurrences = vEvent.GetOccurrences(checkTime.AddSeconds(-1), checkTime);
Console.WriteLine(occurrences.Count); // "0" - Should be 1, it is dairly recurrence

occurrences = vEvent.GetOccurrences(checkTime.AddSeconds(-1), checkTime.AddSeconds(1));
Console.WriteLine(occurrences.Count); // "1"
```

PR #443, opened 2019-03-19
FIX BUG: Stop adding referenceDate to results when at least one RRule is defined 

ITESMEDIA-Devs:
We changed a few things based on the following premises:

- The DTSTART of a recurring calendar component must not be included the recurrence set if it does not match the RRule
- Calling RecurrenceUtil.GetOccurrences with the same periodStart and periodEnd must include events that starts at that specific moment
- Calling RecurrenceUtil.GetOccurrences with the same periodStart and periodEnd must not include events that ends at that specific moment

Not adding referenceDate to the results set at the start allows the rest of the logic to run normally and only add that date if it's part of the reccurenceRule. It will still be added to the results if the element has no recurrenceRule defined.

Theses modifications make issue #432 more predictable by returning the same results on the first day and subsequent days which are in the RRule. A slightly modified version of the code provided from that issue has been added to the unit tests.


BrettBurbidge:
Please approve this change. I am doing some strange (if date != today) stuff to get around this now.  Thank you for your contribution.

bkadhim:
I don't understand this change. I am experimenting with migrating from the Aspose.Email library and this is creating an incompatibility. My reading of the iCal standard is that DTSTART always represents the first occurrence (section 3.3.10):

The ["DTSTART"](https://icalendar.org/iCalendar-RFC-5545/3-8-2-4-date-time-start.html) property value always counts as the first occurrence.

Not including makes it more complicated to deal with exclusion rules, because if the exclusion rule happens to include DTSTART, one would expect it to be removed.

minichma:
Actually dtstart MUST match the first occurrence according to the RFC. If it doesn't, the behavior is undefined. So this change should only cause incompatibilities with calendars that violate the standard.

PR #449, opened 2019-05-16
Fix inverted IsActive property

thoemy:
Fixes a bug that was introduced in commit a3c3025 when IsActive was converted to a property.


rianjs:
The EventStatus.ics file is missing, so there's an exception thrown. I'm going to commit the changes, because it doesn't break other unit tests, and comment out the line. If you add the ics file back and uncomment it, I'll merge it in.

rianjs (approved):

rianjs:

PR #450, opened 2019-06-05
BUG FIX: Mutation of property can output as an array

TheeJamesLee:
Bug found where changing a property that was already set either by deserialization or property set.

On serialization, will be output as an array of values.

Corrects the SetValue property and adds test in serialization exercising a property change.


rianjs:

PR #463, opened 2019-09-12
BYWEEKNO & BYMONTH fix & ISO 8601 week number fix

NorekZ:
BYWEEKNO was not working completely (week numbers was not considered when generating dates)
BYMONTH was not working completely (month numbers was not considered when generating dates)

...see tests for more details

Week numbers was retrieved from .NET's Calendar which behaves differently than ISO 8601 (see https://blogs.msdn.microsoft.com/shawnste/2006/01/24/iso-8601-week-of-year-format-in-microsoft-net/). iCalendar RFC says that ISO 8601 should be used.


rianjs:

PR #470, opened 2019-11-04
Checking obj2 with MinValue of type obj2, not type obj1.

renefloor:
Found this code error while searching for an other bug.

It currently checks the value of obj2 against the MinValue of obj1.

`var fi2 = t1.GetField("MinValue");` doesn't look like it should be that way. I fixed this by checking against a new t2.


minichma (approved):

minichma:
Thank your and I apologize in the project's name for the delay. The project is more vital these days. Would you rebase that? Or alternatively give me Push permission?

axunonb (approved):

PR #471, opened 2019-11-04
Checking for int.MinValue separately.

renefloor:
We use this library with Xamarin. When updating Visual Studio this library broke, but only when building for a release package. When debugging everything worked fine. 

After adding extra logging we found that `t1.GetField("MinValue")` resulting in `null` and thus `isMin1` was false. When setting an EndDate it complained that "Both COUNT and UNTIL cannot be supplied together; they are mutually exclusive."

As `CheckMutuallyExclusive` is only used once and count is always obj1 I just added a check for int.MinValue if obj1 is an int.

I cannot find why this behaviour changed. I also don't know whether it is the xamarin compiler, the .net framework or the c# language. (We forced c# to 7.0, so the last one shouldn't be the problem.) 

For now this fixes our problem, but I can understand it when you don't really like the solution and I'm willing to discuss a better solution.

### DIFF
--- a/net-core/Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
+++ b/net-core/Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
@@ -75,6 +75,14 @@ namespace Ical.Net.Serialization.DataTypes
             {
                 return;
             }
+
+            // For some reason GetField("MinValue") sometimes returns 'null' when the obj is an int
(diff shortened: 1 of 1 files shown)


winstonpang:
We got this exact same problem as well in an app that's deployed, I've logged an issue on the xamarin repo as well:

https://github.com/xamarin/xamarin-macios/issues/12508

minichma:
This will be fixed soon by separate efforts making `COUNT` and `UNTIL` nullable rather than working with magic numbers (MinValue).

PR #484, opened 2020-03-06
Bugfix: UNTIL now works correctly for positive TZ offset

pierluca:
In our application, using the RecurrencePatternEvaluator, we noticed a bug which boils down to what was reported in https://github.com/rianjs/ical.net/issues/406 .

This is our simple fix to this bug, all unit tests remain functional.

We did not work on the .Net Core implementation, which seems quite different. Is this by design?

### DIFF
--- a/v2/ical.NET/Evaluation/RecurrencePatternEvaluator.cs
+++ b/v2/ical.NET/Evaluation/RecurrencePatternEvaluator.cs
@@ -299,7 +299,7 @@ namespace Ical.Net.Evaluation
                         {
                             break;
                         }
-                        else if (pattern.Until == DateTime.MinValue || candidate <= pattern.Until)
+                        else
                         {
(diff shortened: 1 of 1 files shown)


v2/ical.NET/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -299,7 +299,7 @@ private void EnforceEvaluationRestrictions(IRecurrencePattern pattern)
                         {
                             break;
                         }
-                        else if (pattern.Until == DateTime.MinValue || candidate <= pattern.Until)
```
pierluca, line 302: as you can see, these conditions are redundant with line 305 anyways, but here the candidate is not treated in the UTC timezone, hence the bug, as we're comparing apples to oranges.

pierluca:
Just realized about the mistake, in fixing v2 rather than v4.
The utc conversion problem seems to persist in v4 though.

pierluca:
Mistaken debugging on our side. Never mind. Sorry !

PR #490, opened 2020-04-16
Jenkins integration

brianhy:
Preliminary Jenkins Integration for this multibranch

Right now Jenkins won't build this, so we'll need to check this in, then follow-up with changes to make things work.


brianhy:
I didn't realize I was in the original repo when I made the above changes.  Closing this out.

PR #513, opened 2021-04-06
Ordering of week numbers should not matter

gabet1337:
Fixes a bug where the order of the byweekno had an effect on the outcome


rianjs (approved):

rianjs:

Issue #448, opened 2019-05-08
Issue parsing ics from Beeline, Inc


lloiodice:
We run into an ICS which has the following line:

SUMMARY; ENCODING=QUOTED-PRINTABLE:Interview:=20Kalim,=20Usraa=20(24524-1)

iCal.Net library version 4.1.11 throws an exception:
Could not parse line: 'SUMMARY; ENCODING=QUOTED-PRINTABLE:Interview:=20Kalim,=20Usraa=20(24524-1)'

   at Ical.Net.Serialization.SimpleDeserializer.ParseContentLine(SerializationContext context, String input) in C:\Gradifi\ical.net\net-core\Ical.Net\Serialization\SimpleDeserializer.cs:line 126
   at Ical.Net.Serialization.SimpleDeserializer.Deserialize(TextReader reader)+MoveNext() in C:\Gradifi\ical.net\net-core\Ical.Net\Serialization\SimpleDeserializer.cs:line 78
   at System.Linq.Enumerable.OfTypeIterator[TResult](IEnumerable source)+MoveNext()
   at System.Collections.Generic.List`1.AddEnumerable(IEnumerable`1 enumerable)
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at Ical.Net.CalendarCollection.Load(TextReader tr) in C:\Gradifi\ical.net\net-core\Ical.Net\CalendarCollection.cs:line 34
   at Ical.Net.CalendarCollection.Load(Stream s) in C:\Gradifi\ical.net\net-core\Ical.Net\CalendarCollection.cs:line 27

Would you know why this is not supported?

Thanks in advance

petertsu:
I have the same issue

JamesMenzies:
I have a similar issue with fields with `QUOTED-PRINTABLE` Encoding.

I can see that problems with parsing of are already associated with test data `/src/Ical.Net.CoreUnitTests/Calendars/Serialization/Parse1.ics`.

The associated unit test asserts that the result is either a not null or a `SerializationException`.

However the field I have in source data look more like this which isn't even real `QUOTED-PRINTABLE`:
```
DESCRIPTION;ENCODING=QUOTED-PRINTABLE;CHARSET=UTF-8:=0D=0A=2D=2D=20=44=6F=20=6E=6F=74=20=64=65=6C=65=74=65=20=6F=72=20=63=
[...]

PR #523, opened 2021-06-10
Quoted Printable field support

JamesMenzies:
# Adding support for Quoted Printable fields

Related to Issue #448

* `GetContentLines` string parsers to support line continuation when Encoding is quoted printable AND line ends with =
* handle fields with encoding quoted printable string fields.
* support utf8, utf16, and SJIS encoding in quoted printable encoding.

I am far from convinced that this is a good solution. There are a number of places where this pull seems inelegant.

1. The implementation I have made in of `GetContentLines` that tests for the encoding by a string match is ugly. I would prefer to, get the field name, read the properties, and then use that information while parsing the value but that is structural change.
2. I think I've made an untidy thing in the `DataTypes/QuotedPrintableString.cs` I think I should be wrapping the value in a strongly typed object. The way I am using it as something like a Type flag  feels more like a hack.
3. encoding name matching added to `GetEncoderFor` isn't pretty. 
4. The trick to support SJIS encoding is secret magic. I think you just have to live with in .netstandard it but in is a bear trap that will eat someone.


axunonb:
@ical-org/maintainers 
To my understanding, the presence of `ENCODING=QUOTED-PRINTABLE` and `CHARSET=UTF-8` as property parameters for the `DESCRIPTION` property makes that specific content line **non-compliant** with the iCalendar specification.
The **only allowed values** for the `ENCODING` parameter are "8BIT" and "BASE64" (RFC 5545, Section 3.2.7)
There is also **no property parameter** to declare the **charset** used in a property value within the iCalendar object itself.

Microsoft Outlook partly sets `ENCODING=QUOTED-PRINTABLE` (incorrectly) in the `.ics`, but also uses e.g. `X-ALT-DESC;FMTTYPE=text/html` to transport rich text of the description.

Should we continue to follow-up on this PR or rather close it?

Issue #524, opened 2021-07-27
Multiple Recurrence Patterns (RecurrenceRules & ExceptionRules) are ignored


OlegBarkov:
**Nuget version**: 4.2.0
**Observed behavior**:  If multiple RecurrenceRules and/or ExceptionRules are added to the Event, only the first rule is taken into account. Others are ignored.
**Expected behavio**r:  All RecurrenceRules and ExceptionRules  should be applied.

**Example 1:**

```
var rrule = new RecurrencePattern("FREQ=DAILY;INTERVAL=1"); // Event occurs every day
var exception1 = new RecurrencePattern("FREQ=WEEKLY;INTERVAL=1;BYDAY=MO"); // Except Mondays
var exception2 = new RecurrencePattern("FREQ=MONTHLY;INTERVAL=1;BYDAY=1TU"); // Except the first Tuesday of each month
var startDate = DateTime.Parse("07/30/2021");
var endDate = DateTime.Parse("08/07/2021");

var iCalevent = new CalendarEvent
{
    Start = new CalDateTime(2021, 8, 1),
    RecurrenceRules = new List<RecurrencePattern> { rrule },
    ExceptionRules = new List<RecurrencePattern>() { exception1, exception2 }
};
var calendar = new Calendar();
calendar.Events.Add(iCalevent);

var occurrences = calendar.GetOccurrences(startDate, endDate).AsEnumerable();
foreach (var occurrence in occurrences)
{
    Console.WriteLine($"{occurrence.Period.StartTime:dddd, MMMM dd, yyyy h:mm:ss tt}");
}
   /*
    * Actual output:
    * Sunday, August 01, 2021 12:00:00 AM
    *                                       <-- No Monday is expected due to exception1
    * Tuesday, August 03, 2021 12:00:00 AM  <-- this is UNEXPECTED since it is the first Tuesday of the month (see exception2)
    * Wednesday, August 04, 2021 12:00:00 AM
    * Thursday, August 05, 2021 12:00:00 AM
    * Friday, August 06, 2021 12:00:00 AM             
    */
```
**Example 2:**

```
var rrule1 = new RecurrencePattern("FREQ=WEEKLY;INTERVAL=1;BYDAY=MO"); // Event occurs on each Monday
var rrule2 = new RecurrencePattern("FREQ=MONTHLY;INTERVAL=1;BYDAY=1TU"); // Event occurs on the first Tuesday of each month
var startDate1 = DateTime.Parse("08/01/2021");
[...]

PR #525, opened 2021-07-27
BUG FIX: Multiple Recurrence Patterns (RecurrenceRules & ExceptionRules) are ignored

OlegBarkov:
Related to the issue [#524](https://github.com/rianjs/ical.net/issues/524)

Instead of using the Evaluator object from the first object select many results from each Evaluator and then combine them into the Hashset.


kibblewhite (commented):
Looking forwards to seeing if this one fixes the issue of multiple recurrence patterns. Thanks for the fix, it looks promising.

chunkychode (approved):

Issue #491, opened 2020-04-23
Period in RDate not fully supported in serialization


matzeboe:
The serializer does not fully support RDates with periods.
When parsing an RDate with a given period everything is fine but wehn it is serialized only the date is present. Information about duration etc. is lost.

E.g.:
RDATE;VALUE=PERIOD:20170103T020000Z/20170103T021500Z,20170215T010000Z/PT3H
becomes
RDATE;VALUE=PERIOD:20170103T020000Z,20170215T010000Z

This leads to anther problem: If we parse a VCard in order to make changes, serialize it for storage and parse it again when we use it the next time, the information about periods (duration etc.) in rdates is completely lost now.

```
public void RDatePeriodNotSupported()
        {
            const string ical =
        @"BEGIN:VCALENDAR
PRODID:-//ddaysoftware.com//NONSGML DDay.iCal 1.0//EN
VERSION:2.0
BEGIN:VEVENT
DTSTART:20160829T080000
DTEND:20160829T090000
RDATE;VALUE=PERIOD:20170103T020000Z/20170103T021500Z,20170215T010000Z/PT3H
SEQUENCE:0
SUMMARY:Do your cardio
UID:abab717c-1786-4efc-87dd-6859c2b48eb6
END:VEVENT
END:VCALENDAR";

            var collection = Ical.Net.Calendar.Load(ical);
            //var firstEvent = collection.Events.First();
            var serializer = new Ical.Net.Serialization.CalendarSerializer(new Ical.Net.Serialization.SerializationContext());
            var serialized = serializer.SerializeToString(collection);

            Console.WriteLine(serialized);
            Console.WriteLine("Problem: The serialized calendar does not contain the proper period information from RDate (Only the date is present now)).");
            Console.WriteLine("Due to examples in RFC-5545 this is allowed (https://tools.ietf.org/html/rfc5545#section-3.8.5.2)");

            //The following part shows that information is lost:
            var collection2 = Ical.Net.Calendar.Load(serialized);
            var durationFromFirstParsing = collection.Events.First().RecurrenceDates.First().First().Duration;
            var durationFromSecondParsing = collection2.Events.First().RecurrenceDates.First().First().Duration;
            System.Diagnostics.Debug.Assert(durationFromFirstParsing == durationFromSecondParsing);

            Console.ReadKey();
        }
```

s3mic0lin:
Created pull request #528 to fix this issue.

@matzeboe  Can you confirm this fixes your issue?

PR #528, opened 2021-08-28
Fixed issue#491

s3mic0lin:
Fixed issue #491 
Duration and end time was not appended to period. By checking for p.Endtime.HasTime adding end time or duration was ensured.

Unit tests passed.


DemeSzabolcs:
@s3mic0lin I had the same problem as you described, I tried your fix and it worked for me!
@rianjs Have you seen this one?

kevinsmgov:
I could use this fix in the NuGet version of this library as well.

kevinsmgov:
Also line 41-47 in EventEvaluator reset rdate EndTime values even if they are supplied (and appear to be redundant to 50-56).

i.e. if I add periods with different durations to RecurrenceDates of an event, EventEvaluator resets them.

PR #560, opened 2022-10-23
Use specified culture/format provider when provided to CalDateTime.ToString

nzbart:
The ToString method on CalDateTime has useful logic to render the date and/or time components, but this rendering ignores the format provider and always uses the current culture. The format provider is currently used when the format string is specified, but not otherwise.

This PR adds some characterisation tests and alters ToString to use the format provider for date formatting.


axunonb:
The PR is included in #631. Thanks for posting!
Sorry for the confusion when pushing mistakenly to your PR.

Issue #564, opened 2023-01-10
PeriodSerializer not working with midnight DATE-TIME.


minichma:
PeriodSerializer doesn't deserialize a period's duration correctly if the start IDateTime's time is set to midnight.

E.g. the following period string should be serialized to a period with a duration of 6H but actually has 1D:
`20180219T000000Z/PT6H`

Can be reproduced via the following test case:

```CS
[Test]
public void TestPeriods()
{
    var serializer = new PeriodSerializer();
    var res = serializer.Deserialize(new System.IO.StringReader("20180219T000000Z/PT6H")) as Period;

    Assert.AreEqual(6.0, res.Duration.Hours);
    Assert.IsTrue(res.StartTime.HasTime);
 }
```

The problem boils down to `Period` [automatically assuming a duration of 1D if the time component is zero](https://github.com/rianjs/ical.net/blob/1ad17e51a0e71b30aabf88d95c16066f4db6e08b/src/Ical.Net/DataTypes/Period.cs#L142).

Also see [similar code in CalDateTime.cs](https://github.com/ical-org/ical.net/blob/35b506c479d16102b5c22f9d210c853ed1698c8d/Ical.Net/DataTypes/CalDateTime.cs#L93) and a related issue in #602.

axunonb:
Please double-check the new version merged with #595

minichma:
@axunonb: Yes, the problem still exists.

minichma:
@axunonb I think we should reopen this. Don't think I can do that myself. Started working on a PR but could take a while until I find enough time, as all these little issues are interconnected.

minichma:
@axunonb Thanks for the invitation! Reopened the relevant issues.

minichma:
Fixed with #638

PR #598, opened 2024-10-14
Keep apart `VEVENT.DTEND` and `.DURATION`

### LINKED ISSUE #574: CalendarEvent mixes up `DURATION` and `DTEND` on deserialization
In Ical.Net `VEVENT` components are deserialized into `CalendarEvent` objects. During the deserialization process the `DURATION` property is extrapolated into `DTEND` and vice versa. I.e. one is calculated out of the other, so after deserialziation both are present. The information, which of both was contained in the original `VEVENT` component is lost during deserialization.

This is a problem when it comes to recurring events, because according to RFC 5545 recurrence sets are generated differently depending on which of both was used to specify the duration.

See https://www.rfc-editor.org/rfc/rfc5545#section-3.8.5.3:

```
      If the duration of the recurring component is specified with the
      "DTEND" or "DUE" property, then the same exact duration will apply
      to all the members of the generated recurrence set.  Else, if the
      duration of the recurring component is specified with the
      "DURATION" property, then the same nominal duration will apply to
      all the members of the generated recurrence set and the exact
      duration of each recurrence instance will depend on its specific
      start time.
```

Suggestion: During deserialization retain the information, which property (`DURATION`, `DTEND`, `DUE`) was used to specify an VEVENT's duration.

minichma:
Fixes https://github.com/ical-org/ical.net/issues/574
Fixes https://github.com/ical-org/ical.net/issues/629

According to RFC 5545 `VEVENT`'s properties `DTEND` and `DURATION` cannot be calculated from each other, because the evaluation rules are slightly different. The previous implementation of `CalendarEvent` extrapolated both values based on each other, and as a consequence remove the information, which of both was originally specified.

With this PR the extrapolation only happens on the fly when needed but only the originally specified property is stored. The additional information is not used for any calculations so far, but can in the future.

DISCUSS: This is a breaking change, as uses might expect DtEnd to be calculated from Duration and vice versa. An alternative would be to keep the behavior of the properties as is but calculate fallback on the fly when needed rather than upfront. That way the `CalendarEvent.DtEnd` and `.Duration` properties would behave as before but the serialized versions might still be slightly different. However, the behavior would be less transparent and potentially confuse new users. TBD

### DIFF
--- a/Ical.Net.Tests/CalendarEventTest.cs
+++ b/Ical.Net.Tests/CalendarEventTest.cs
@@ -469,4 +469,81 @@ END:VCALENDAR";
         var expectedNegative = TimeSpan.FromMinutes(-17.5);
         Assert.That(negativeOffset?.Offset, Is.EqualTo(expectedNegative));
     }
+
+
+    [Test, Category("CalendarEvent")]
+    public void TestGetEffectiveDuration()
+    {
+        var now = _now.Subtract(TimeSpan.FromTicks(_now.Ticks % TimeSpan.TicksPerSecond));
+
+        var evt = new CalendarEvent()
+        {
+            DtStart = new CalDateTime(now) { HasTime = true },
+            DtEnd = new CalDateTime(now.AddHours(1)) { HasTime = true },
+        };
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(evt.DtStart.Value, Is.EqualTo(now));
+            Assert.That(evt.DtEnd.Value, Is.EqualTo(now.AddHours(1)));
+            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
+        });
+
+        evt = new CalendarEvent()
+        {
+            DtStart = new CalDateTime(now.Date) { HasTime = true },
+            DtEnd = new CalDateTime(now.Date.AddHours(1)) { HasTime = true },
+        };
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(evt.DtStart.Value, Is.EqualTo(now.Date));
+            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
+        });
+
+        evt = new CalendarEvent()
+        {
+            DtStart = new CalDateTime(now.Date) { HasTime = false },
+        };
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(evt.DtStart.Value, Is.EqualTo(now.Date));
+            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromDays(1)));
+        });
+
+        evt = new CalendarEvent()
+        {
+            DtStart = new CalDateTime(now) { HasTime = true },
+            Duration = TimeSpan.FromHours(2),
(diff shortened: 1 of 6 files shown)


axunonb:
Yes, I saw these calculations when reviewing the code for the v4.3 release. Following RFC 5545 is clearly the preferred way to go. We can expect more breaking changes, so we could also consider a temporary flag like v4-compatibility-enabled. This could be introduced at any time, and all new code could have RFC compliance in focus. v5 could then switch compatibility == false by default. What do you think?

There will be a new major version anyway, as soon as the project is migrated to NRT enabled. I can't foresee yet, how much time this will require and how much risks this will bring because of low code coverage in some areas.

minichma:
Yes, a temporary flag would certainly be an option. Where would we put it? But on the other hand, this would introduce quite some additional complexity to the code and require a considerable amount of additional testing. As you intend to introduce a v5 soon anyways, it might also be a good idea to start collecting those breaking changes from now on. [Similar code is contained in Period.cs](https://github.com/ical-org/ical.net/blob/35b506c479d16102b5c22f9d210c853ed1698c8d/Ical.Net/DataTypes/Period.cs#L87-L101), which seems to be related to #564 and should be improved as well. Keeping everything backwards-compatible might not be worth the effort.

axunonb:
To setup a project and a new v5 branch looks like the way to go, I agree.

minichma:
@axunonb While rebasing this PR I stumbled across following change, because the affected test fails now

https://github.com/ical-org/ical.net/pull/620/files#r1807942277

The problem is that cloning of the calendar events normalizes the `ORGANIZER` uri. As a consequence, in the affected test, if the event is cloned, it will be serialized to lower-case `mailto:...`. However, if it is not cloned, it will NOT be converted to lower-case. The reason for that behavior is, that in

https://github.com/axunonb/ical.net/blob/469c767a00d6e0d1c88bda8dd647ad8487f8f050/Ical.Net/DataTypes/Attendee.cs#L251

we create a copy of the Uri by applying a pattern of the form `newUri = new Uri(oldUri.ToString())`. The copy will not have the same `OriginalString` as the old uri.

Previously the calendar has always been copied during serialization due to the fact, that `DTEND` and `DURATION` have always both been present and due to this:

https://github.com/axunonb/ical.net/blob/fedbd4f10de072b69662fec40b14ec05681a9a61/Ical.Net/Serialization/EventSerializer.cs#L26

The question is now, what the expected behavior would be. From my perspective copying should not have any side effects, so we should just copy the reference to the URI rather than cloning the URI itself. Copying the reference is safe, as Uri is immutable. The question is, whether an implicit conversion from lower to upper-case should happen or not. In this case I'd say its the responsibility of the caller to provide clean input. I'd generally recommend keeping the original input rather than applying sanitization.

axunonb:
> copying should not have any side effects, so we should just copy the reference to the URI rather than cloning the URI itself

agree

> I'd generally recommend keeping the original input rather than applying sanitization

agree

I remember this failing test, and obviously I was not diving deep enough. Thanks!

axunonb (approved):
Love it, no more `ExtrapolateTimes`, thanks!
There is a new branch `version/v5` where this PR should get merged.

PR #616, opened 2024-10-19
Define behavior for time-based FREQ and date-only DTSTART

### LINKED ISSUE #602: Test 'Evaluate1' broken - should fail but succeeds
The test 'Evaluate1' is broken, as it tests for occurrences' properties but doesn't produce any. This is a improved version of the test. The relevant difference is the start date passed to `GetOccurrences` is fixed and `Assert.AreEqual(5, occurrences.Count)` is added. The fixed version fails on current main branch.

However, I'm not sure, whether the case covered in that test is allowed at all but I assume it isn't. I.e. the `MINUTELY` freq requires a time component, which doesn't exist in the given case, because `evt.Start` is date-only. The comment says 'Start at midnight, UTC time', but without a time component, there is no real start date. Maybe it wasn't intended to have a date-only start, so then the problem would be caused [by this](https://github.com/ical-org/ical.net/blob/35b506c479d16102b5c22f9d210c853ed1698c8d/Ical.Net/DataTypes/CalDateTime.cs#L93), which is a problem closely related to #564.

```cs
        /// <summary>
        /// Ensures that the StartTime and EndTime of periods have
        /// HasTime set to true if the beginning time had HasTime set
        /// to false.
        /// </summary>
        [Test, Category("Recurrence")]
        public void Evaluate1()
        {
            Calendar cal = new Calendar();
            CalendarEvent evt = cal.Create<CalendarEvent>();
            evt.Summary = "Event summary";

            // Start at midnight, UTC time
            evt.Start = new CalDateTime(DateTime.SpecifyKind(DateTime.Today, DateTimeKind.Utc));

            evt.RecurrenceRules.Add(new RecurrencePattern("FREQ=MINUTELY;INTERVAL=10;COUNT=5"));
            var occurrences = evt.GetOccurrences(CalDateTime.Today.AddDays(-1), CalDateTime.Today.AddDays(1));

            Assert.AreEqual(5, occurrences.Count);
            Assert.IsTrue(occurrences.All(x => x.Period.StartTime.HasTime), "All recurrences of this event should have a time set.");
            Assert.IsTrue(occurrences.Skip(1).All(x => x.Period.StartTime.Value.TimeOfDay != TimeSpan.Zero));
        }
```

minichma:
Fixes https://github.com/ical-org/ical.net/issues/602.

Implement defined behavior of recurrence evaluation in case of a time-based FREQ (sec/min/hour) and a DTSTART of type DATE (i.e. date-only). With this PR such scenarios are handled as if DTSTART had a time set to midnight.

The case was previously covered by test `Evaluate1`, which was broken and is fixed with this PR.

### DIFF
--- a/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
@@ -900,6 +900,15 @@ namespace Ical.Net.Evaluation
 
         public override HashSet<Period> Evaluate(IDateTime referenceDate, DateTime periodStart, DateTime periodEnd, bool includeReferenceDateInResults)
         {
+            if ((this.Pattern.Frequency != FrequencyType.None) && (this.Pattern.Frequency < FrequencyType.Daily) && !referenceDate.HasTime)
+            {
+                // This case is not defined by RFC 5545. We handle it by evaluating the rule
+                // as if referenceDate had a time (i.e. set to midnight).
+
+                referenceDate = referenceDate.Copy<IDateTime>();
+                referenceDate.HasTime = true;
+            }
+
             // Create a recurrence pattern suitable for use during evaluation.
             var pattern = ProcessRecurrencePattern(referenceDate);
 
--- a/Ical.Net.Tests/RecurrenceTests.cs
+++ b/Ical.Net.Tests/RecurrenceTests.cs
@@ -2823,21 +2823,43 @@ namespace Ical.Net.Tests
         /// HasTime set to true if the beginning time had HasTime set
         /// to false.
         /// </summary>
-        [Test, Category("Recurrence")]
-        public void Evaluate1()
+        [Category("Recurrence")]
+        [TestCase("SECONDLY", 1, true)]
+        [TestCase("MINUTELY", 60, true)]
+        [TestCase("HOURLY", 3600, true)]
+        [TestCase("DAILY", 24*3600, false)]
+        public void Evaluate1(string freq, int secsPerInterval, bool hasTime)
         {
             Calendar cal = new Calendar();
+
             CalendarEvent evt = cal.Create<CalendarEvent>();
             evt.Summary = "Event summary";
 
             // Start at midnight, UTC time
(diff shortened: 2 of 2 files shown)


Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -900,6 +900,15 @@ private Period CreatePeriod(DateTime dt, IDateTime referenceDate)
 
         public override HashSet<Period> Evaluate(IDateTime referenceDate, DateTime periodStart, DateTime periodEnd, bool includeReferenceDateInResults)
```
axunonb, line 901: While we update the code, let's also add missing or update existing XmlDoc (for us and for others).

axunonb (dismissed):
Nice, thanks!

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
minichma: Basically a great idea, but also quite challenging. In some cases we can work on details of the code without having a comprehensive understanding of the code as a whole. But the latter is needed for writing a meaningful doc comment. In the given case I wouldn't feel comfortable with writing a doc comment, as I don't have sufficient understanding of the code as a whole. And a bad comment could be misleading and therefore counterproductive.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -900,6 +900,15 @@ private Period CreatePeriod(DateTime dt, IDateTime referenceDate)
...
         {
+            if ((this.Pattern.Frequency != FrequencyType.None) && (this.Pattern.Frequency <= FrequencyType.Daily) && !referenceDate.HasTime)
```
minichma: oh, this should be Frequency < Daily (not <=). Will fix

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: Let's say, add docs as far as we believe to understand :)
No need to be complete in all aspects, but should contain what we found out.

axunonb (approved):
Free free to merge if this PR is complete now

PR #617, opened 2024-10-19
Test: Run recurrence test cases from the libical project and introduce generic file-based test cases.

minichma:
This PR is for discussion, not to be merged (at least for now).

The [libical project](https://github.com/libical/libical) maintains a [list of test cases](https://github.com/libical/libical/blob/cfd401b9d043214395888de1d9daf52263e3245b/src/test/icalrecur_test.out
) for their recurrence evaluation code. The PR adds a copy of the test case definitions and runs them on Ical.Net. 8/70 test cases fail, but I didn't have the time to dig deeper for now. However, the failing tests are probably worth an investigation.

Licensing: If it turns out that the test cases make sense and we think this should be merged, we'd have to check the licensing. The file used from the libical project is available under a dual license, either the MPL v2.0 or the LGPL v2.1. I assume the MPL v2 should be compatible with this project, but at this time that's just a guess.


minichma:
@axunonb what do you think?

axunonb:
If it's possible to use existing wheels, reinventing is a waste of time.
So yes, let's move on.

> Licensing

Using LGPL-licensed Code in MIT Projects: brings - in general - some complexity. If we include LGPL-licensed code in ICal.Net that is licensed under the MIT license, we must comply with the LGPL's requirements. 
This typically means that any modifications to the LGPL-licensed code must be made available under the same LGPL license, and we must provide a way to relink the application with a modified version of the LGPL library.

At the same time we'd like to use what they're publishing "as is", and also only for the test project. We can add a note to contributors like "Do not modify code in the contrib folder".
This will reduce the license complexity close to zero. The files included in this PR are sufficient re licensing.

Just one more thing: Tt would be a matter of courtesy to inform [libical project](https://github.com/libical/libical) owners that we are using parts of their code for testing, and ask for their formal okay.

minichma:
The project is published under a dual license. We can choose between the LGPL and the MPL. Maybe the MPL is more suitable for us. What do you think?

minichma:
@winterz We consider reusing test definitions from the libical project in Ical.Net by copying the `icalrecur_test.out` file. :-) Do you think license comments introduced in this PR are sufficient? Do you have any thoughts on compatibility between libical's LGPL/MPL and this project's MIT license?

winterz:
> @winterz We consider reusing test definitions from the libical project in Ical.Net by copying the `icalrecur_test.out` file. :-) Do you think license comments introduced in this PR are sufficient? Do you have any thoughts on compatibility between libical's LGPL/MPL and this project's MIT license?

Yes, you can add icalrecur_test.out to ical.net project. mixing MPL with MIT is ok. you need to mention somewhere that this file is MPL licensed and copyrighted by "the libical developers".

minichma:
@winterz Thank you!

minichma:
@axunonb Please have a closer look especially at the licensing text.

license.md
```diff
@@ -24,3 +24,12 @@ WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 OTHER DEALINGS IN THE SOFTWARE.
 
+=====================
+
+
+The following directories and their subdirectories contain code
+thatw as developed by different projects and is therefore available
```
axunonb: ... that **was** developed ...

Ical.Net.Tests/contrib/libical/LICENSE.txt
```diff
@@ -0,0 +1,34 @@
...
+     The Initial Developer of the Original Code is Eric Busboom
+
+     All Rights Reserved.
+
+     Contributor(s): See individual source files.
+
+     Copyright: The libical developers
+
+
+This (the Ical.Net) project uses the code under the
```
axunonb: Maybe more clear:
```The project that contains this LICENSE.txt file (ICal.Net) uses the code under the```

File: Ical.Net.Tests/contrib/libical/LICENSE.txt [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+
+     All Rights Reserved.
+
+     Contributor(s): See individual source files.
+
+     Copyright: The libical developers
+
+
+This (the Ical.Net) project uses the code under the 
+Mozilla Public License (MPL) v2.0.
```
axunonb, line 33: Yep, that's compatible with MIT

axunonb (dismissed):
Just noticed that this
https://github.com/ical-org/ical.net/blob/2146ea076dcbe090b7decb7da7b4ea4b329fe06f/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L346
is the only reference to https://github.com/ical-org/ical.net/blob/2146ea076dcbe090b7decb7da7b4ea4b329fe06f/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L472
and the ```bool? expand``` arg is always `true` in unit tests meaning this block
https://github.com/ical-org/ical.net/blob/2146ea076dcbe090b7decb7da7b4ea4b329fe06f/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L493-L514
never gets called. A test with `false` as arg would be good, if applicable.
Must not necessarily go into this PR

minichma:
@axunonb Thanks for the review. Please approve once again!

> A test with false as arg would be good, if applicable. Must not necessarily go into this PR

Good finding! Agree that this PR should just take care about adding the infrastructure and the existing libical test set, not go deep with additional tests. Coincident (more or less): Exactly this test case will probably be added to the libical test cases too, see https://github.com/libical/libical/issues/791. I'm currently also doing some reviews and testing there, which leads to a number of additional test cases. Will try to add them here as well over time. Unfortunately my time is quite limited (investing way more than I should already), so I don't have an ETA yet.

The test case would probably have to be of the form `FREQ=[HOURLY|MINUTELY|SECONDLY];BYYEARDAY=...`, because `BYYEARDAY` is limiting only in the case of these frequencies.

axunonb (approved):
Thanks a lot the time and dedication. It's a lot of work and amazing what you're doing!
Regarding ETA: we can't catch up 3 years in 2 weeks time - I'm ware.

minichma:
Thanks for the kind words. Its really you who deserves the kudus. Great to see, you're taking care of this important project.

Issue #618, opened 2024-10-20
Evaluate some tests failing from the libical test suite.


minichma:
When running a [list of test cases](https://github.com/libical/libical/blob/cfd401b9d043214395888de1d9daf52263e3245b/src/test/icalrecur_test.out) from [the libical project](https://github.com/libical/libical), the following tests fail. The reason should be investigated.

- [x] `DTSTART:20120101T120000`, `RRULE:FREQ=YEARLY;BYYEARDAY=-366;COUNT=3`
  Values differ at index [1]
  Expected: <01.01.2016 12:00:00 UTC>
  But was:  <31.12.2012 12:00:00 UTC>

- [x] `DTSTART:20121231T120000`, `RRULE:FREQ=YEARLY;BYYEARDAY=366;COUNT=3`
    Values differ at index [1]
    Expected: <31.12.2016 12:00:00 UTC>
    But was:  <01.01.2014 12:00:00 UTC>

- [x] `DTSTART:20121231T120000`, `RRULE:FREQ=YEARLY;BYYEARDAY=366;UNTIL=20200101T000000`
   Values differ at index [1]
   Expected: <31.12.2016 12:00:00 UTC>
   But was:  <01.01.2014 12:00:00 UTC>

- [x] `DTSTART:20130101T000000`, `RRULE:FREQ=YEARLY;BYWEEKNO=1,2,-1,-2;BYDAY=TU;UNTIL=20170101T000000Z`
    Fixed in #654
    System.ArgumentOutOfRangeException : The added or subtracted value results in an un-representable DateTime. (Parameter 'value')
    DateTime.ThrowDateArithmetic(Int32 param)
    DateTime.AddTicks(Int64 value)
    RecurrencePatternEvaluator.GetWeekNoVariants(List`1 dates, RecurrencePattern pattern, Nullable`1 expand) line 425

- [x] `DTSTART:20141006T090000`, `RRULE:FREQ=WEEKLY;INTERVAL=1;COUNT=2;BYDAY=MO;`
    Fixed in #645
    System.IndexOutOfRangeException : Index was outside the bounds of the array.
    RecurrencePatternSerializer.Deserialize(TextReader tr) line 244

- [x] `DTSTART:20161229T090000`, `RRULE:FREQ=WEEKLY;UNTIL=20170127T000000Z;WKST=MO;BYDAY=SU,TU,TH;INTERVAL=2`
    Fixed in #645
[...]

PR #619, opened 2024-10-20
RRULE evaluation: Ignore BYtime if DTSTART is date-only

minichma:
According to RFC 5545, when evaluating an RRULE based on a DTSTRT of type DATE (i.e. date-only), any BY[SECOND|MINUTE|HOUR] MUST be ignored. This is implemented by this PR.

> The BYSECOND, BYMINUTE and BYHOUR rule parts MUST NOT be specified
> when the associated "DTSTART" property has a DATE value type.
> These rule parts MUST be ignored in RECUR value that violate the
> above requirement(e.g., generated by applications that pre - date
> this revision of iCalendar).

This fixes the case `DTSTART:20241018, RRULE:FREQ=DAILY;BYMINUTE=1,2,3,4;INTERVAL=2;COUNT=3` of #618.


axunonb (approved):
Nicely commented,, important fix, tnx

axunonb:
@minichma Feel free to merge the 2 approved PRs after considering whether the remarks are useful or not.

The only thing I'm pretty much convinced by now is: we need more coverage of use cases from the wild, which results in more effort than code coverage. Thanks for the structured approach you already started.
BTW: Would CI based code coverage integration be useful at this time, and if so: What should be the minimum coverage and code quality benchmark for PRs?

minichma:
Regarding test cases for this fix I was running [this](https://github.com/libical/libical/blob/39e7cc0a0340c828663d5c0ea59a74666ba6121f/src/test/icalrecur_test.out#L364-L366) based on #617. I think we should consider merging #617 and introducing our own test files in the same format. This would allow us adding new tests very easily and also to update test cases from the libical project. Then we'd also have specific coverage for this PR.

minichma:
> Would CI based code coverage integration be useful at this time, and if so: What should be the minimum coverage and code quality benchmark for PRs?

I'm pretty sure it would. Most important would be that the overall coverage would increase over time. Imposing high quality figures would only be helpful if we also have a reasonable path to get there.

Issue #614, opened 2024-10-18
EXDATE should serialize TZID


RemcoBlok:
I have an EXDATE at a specific time with a specific TZID. Presently ical.net does not serialize the TZID with the EXDATE. When deserializing the calendar this results in `GetOccurrences()` returning an occurrence on what was supposed to be the EXDATE had it not lost its TZID.

**_I first saw this issue in ical.net 4.2.0. It is also present in 4.3.1._**

First of all, without serialization and deserialization, this passes:

Example 1
```cs
        [Fact]
        public void ShouldNotOccurOnExceptionDate()
        {
            // Arrange
            Guid id = Guid.NewGuid();
            string timeZoneId = TimeZoneInfo.Local.Id; // GMT Standard Time
            DateTime start = new DateTime(2024, 10, 19, 18, 0, 0, DateTimeKind.Local);
            DateTime end = new DateTime(2024, 10, 19, 19, 0, 0, DateTimeKind.Local);
            DateTime exceptionDate = new DateTime(2024, 10, 19, 21, 0, 0, DateTimeKind.Local);

            RecurrencePattern recurrencePattern = new RecurrencePattern(FrequencyType.Hourly);
            recurrencePattern.Count = 2;
            recurrencePattern.Interval = 3;

            IDateTime exceptionOccurrsOn = new CalDateTime(exceptionDate, timeZoneId);

            PeriodList exceptionDates = new PeriodList();
            exceptionDates.Add(exceptionOccurrsOn);

            CalendarEvent myRecurringEvent = new CalendarEvent();
            myRecurringEvent.Summary = "My Recurring Event";
            myRecurringEvent.Uid = id.ToString();
            myRecurringEvent.Start = new CalDateTime(start, timeZoneId);
            myRecurringEvent.End = new CalDateTime(end, timeZoneId);
            myRecurringEvent.RecurrenceRules.Add(recurrencePattern);
            myRecurringEvent.ExceptionDates.Add(exceptionDates);

            Calendar calendar = new Calendar();
            calendar.Events.Add(myRecurringEvent);

            // Act
            DateTime date = new DateTime(2024, 10, 19);
            HashSet<Occurrence> occurrences = calendar.GetOccurrences<CalendarEvent>(date);

            // Assert
            Assert.Single(occurrences);
        }
```

This correctly returns a single occurrence. It does not return an occurrence on the EXDATE.

Add serialization and deserializtion, and this fails:

Example 2
```cs
        [Fact]
[...]

Issue #591, opened 2024-09-01
Serialized EXDATE is unpermitted PERIOD value type when originated as System.DateTime


jvraines:
This code
```
Calendar myCal = new Calendar();
CalendarEvent myEvent = new CalendarEvent {
    Start = new CalDateTime(2024, 12, 1, 12, 0, 0),
    End = new CalDateTime(2025, 1, 1, 18, 0, 0),
    RecurrenceRules = new List<RecurrencePattern> {
                            new RecurrencePattern {
                                Frequency = FrequencyType.Daily
                            }
                      }
};
myEvent.ExceptionDates.Add(new PeriodList { 
    new CalDateTime(2024, 12, 25)
    new CalDateTime(new DateTime(2024, 12, 26))
});
myCal.Events.Add(myEvent);
Console.WriteLine(new CalendarSerializer().SerializeToString(myCal));
```
produces

BEGIN:VCALENDAR
PRODID:-//github.com/rianjs/ical.net//NONSGML ical.net 4.0//EN
VERSION:2.0
BEGIN:VEVENT
DTEND:20250101T180000
DTSTAMP:20240901T232258Z
DTSTART:20241201T120000
EXDATE:20241225T000000,<mark>20241226/P1D</mark>
RRULE:FREQ=DAILY
SEQUENCE:0
UID:d65129f0-d00c-46d1-a7d2-81a73a41990d
END:VEVENT
END:VCALENDAR
 
The first excluded date is an acceptable DATE-TIME value type. The second, however, is a PERIOD value type which is not permitted by the [EXDATE spec](https://icalendar.org/iCalendar-RFC-5545/3-8-5-1-exception-date-times.html).

RemcoBlok:
Another (presumably related) issue is that if you create the CalDateTime with a TZID (when it is an exception date for an event that is not an all-day event, but for a specific time on the day), iCal.NET does not serialize the TZID with the EXDATE. When deserializing this may result in GetOccurrences() producing an occurrence on the EXDATE, or what was supposed to be the EXDATE if it did not loose the TZID.

What can I do as a temporary workaround here? Fix the EXDATE after deserializing? Or remove the occurrence from the result of GetOccurrences()?

axunonb:
[...]

Issue #590, opened 2024-08-29
Exdate not taken correctly into account when getting the FreeBusyStatus? 


GillesDV:
Hello,

We currently have a problem and I'm wondering if we parse our Calendar string wrong somewhere or if there's a bug in the code. 

Background: In this scenario we have a Calendar that starts at 1st of August. Fills the whole day, and repeats all day. Basically always occupied. The 29th of August is an exception, then we're free. I did a GetFreeBusyStatus check on that day, but rather than Free, I get Busy. 

This happens on ical.net nuget 4.2.0

Reproducable with the following bit of code: 

```
void Main() {
	var timestamp = new DateTimeOffset(2024, 8, 29, 11, 00, 00, 0, TimeSpan.FromHours(2));
	string timeZoneId = "Europe/Paris";

	// All-day, repeating every day, starting from 1st of August. Except 29th of August
	var scheduleEverythingExcept29th = new Schedule2($@"BEGIN:VCALENDAR
BEGIN:VEVENT
DTSTART;TZID=Europe/Paris:20240801T000000
DTEND;TZID=Europe/Paris:20240802T000000
RRULE:FREQ=DAILY
EXDATE:20240828T220000Z
END:VEVENT
END:VCALENDAR");

	var calStatus = scheduleEverythingExcept29th.GetStatus(timestamp, timeZoneId);
	Console.WriteLine("calStatus: " + calStatus);
	// calStatus should be Free. 
        // Since we check in the middle of the 29th and the 29th is an exception of the daily-repetition of always occupied / busy
}
```

```
public sealed class Schedule2
{
	private readonly Calendar m_calendar;

	public Schedule2(string iCalendar)
	{

		Calendar calendar;

		using (var stringReader = new StringReader(iCalendar))
		{
			calendar = Calendar.Load(stringReader);

			if (calendar == null)
				throw new ArgumentException($"Unable to build a calendar out of the specified {nameof(iCalendar)} instance.");
		}

		m_calendar = calendar;
	}

	public FreeBusyStatus GetStatus(DateTimeOffset timestamp, string timeZoneId)
	{
		DateTime timestampLocal = DateTime.SpecifyKind(timestamp.DateTime, DateTimeKind.Local);

		SetCalendarTimeZone(timeZoneId);

		//  Computes the schedule of the day.
		var fromBeginningOfDay = new CalDateTime(timestampLocal.Date, timeZoneId);
[...]

PR #620, opened 2024-10-20
Implement deep copy methods for components

### LINKED ISSUE #149: Event.Copy should do a deep copy
If I do
var tmpEvent = permanentEvent.Copy<IEvent>();
var newStart = new CalDateTime(blabla);
tmpEvent.Start = newStart;

then permanentEvent.Start has also changed to newStart. Shouldn't IEvent.Copy<T>() do a deep copy and return a new instance with all things copied over, not just references?

I am doing all this because permanentEvent.getOccurrences(startTime, endTime) is returning occurences with just the start time set correctly and duration 0 (I see no end time anyway) and I need to calculate  end and duration for each occurrence manually by creating a new event and copying stuff from permanentEvent. If getOccurrences() worked fine then I wouldn't have spotted this oddity.
Or maybe it's the lack of documentation that is confusing me. Can we have that documentation please?

axunonb:
- Implemented and completed `CopyFrom` method from `ICopyable` interface in multiple classes for deep copying.
- Removed questions in code and implemented solution
- Updated `ICopyable` interface documentation.
- Introduced pattern matching for better readability
- Added new test methods in `CopyComponentTests` class for deep copying various calendar components.
- Fixed broken SerializationTests.AttendeesSerialized() (Corrected attendee names and fixed typos in assertions.)
- Refactored `CalendarObjectBase` (should eventually become abstract)
- Added `ExcludeFromCodeCoverage` attribute to `CalendarObjectList`.
- Updated `Ical.Net.csproj` to latest C# version.

Fixes #149


axunonb:
@minichma 

#149 focuses on a special case where deep copy did not work, It would have been simple to fix.
This PR is an effort for a generic approach to make `ICopyable.Copy()` work for all components.

Not sure, whether I discovered all components where an individual `CopyFrom` is needed.
The implementation of the `ICopyable` interface is a bit tricky: If `CopyFrom` is not overridden in a component, it falls back to parents' implementation. That, however, may or may not copy all members. 
At the same time `T Copy<T>()` is only implemented in the base class `CalendarDataType`. From here, all the calls to `CopyFrom` are invoked.

Any comments and further hints are welcome.

Ical.Net.Tests/SerializationTests.cs
```diff
@@ -325,7 +326,7 @@ public void AttendeesSerialized()
 
             foreach (var a in evt.Attendees)
             {
-                var vals = GetValues(vEvt, "ATTENDEE", a.Value.OriginalString);
```
axunonb, line 328: This was the reason for the test to fail.

Ical.Net/CalendarObjectList.cs
```diff
@@ -1,12 +1,13 @@
...
 
 namespace Ical.Net
 {
     /// <summary>
     /// A collection of calendar objects.
     /// </summary>
+    [ExcludeFromCodeCoverage]
     public class CalendarObjectList : GroupedList<string, ICalendarObject>, ICalendarObjectList<ICalendarObject>
     {
-        public CalendarObjectList(ICalendarObject parent) {}
```
axunonb, line 10: useless

minichma:
@axunonb I will try to find some time for a review, but this could take a few days.

axunonb:
Yes, of course, the task required more changes than I expected due to incomplete implementation.
`EXDATE` looks like another hotspot (#614 #590 #591) to be fixed

Ical.Net/DataTypes/Attachment.cs
```diff
@@ -71,8 +71,22 @@ public override string ToString()
...
-        //ToDo: See if this can be deleted
-        public override void CopyFrom(ICopyable obj) { }
+        /// <inheritdoc/>
+        public override void CopyFrom(ICopyable obj)
+        {
+            if (obj is not Attachment att) return;
+            base.CopyFrom(obj);
+
+            Uri = att.Uri != null ? new Uri(att.Uri.ToString()) : null;
+            if (att.Data != null && att.Data.Length != 0)
```
minichma: If Data was set to an empty array, we might want to preserve that. null != []

Ical.Net/DataTypes/Period.cs
```diff
@@ -44,15 +44,13 @@ public Period(IDateTime start, TimeSpan duration)
...
             base.CopyFrom(obj);
 
-            var p = obj as Period;
-            if (p == null)
-            {
-                return;
-            }
+            if (obj is not Period p) return;
+
             StartTime = p.StartTime;
```
minichma: StartTime/EndTime are ref types (ICalDateTimes), so they should probably be copied too.

Ical.Net/DataTypes/StatusCode.cs
```diff
@@ -45,15 +45,14 @@ public StatusCode(string value) : this()
...
             base.CopyFrom(obj);
-            if (obj is StatusCode)
-            {
-                var sc = (StatusCode)obj;
-                Parts = new int[sc.Parts.Length];
-                sc.Parts.CopyTo(Parts, 0);
-            }
+            if (obj is not StatusCode statusCode) return;
+
+            Parts = new int[statusCode.Parts.Length];
```
minichma, line 54: At some point it might be worth the effort to modernize the syntax to something like (but prob. not in this PR)
`Parts = statusCode.Parts.ToArray();`
or
`Parts = [..Parts];`

Ical.Net/DataTypes/Trigger.cs
```diff
@@ -73,15 +73,15 @@ public Trigger(string value) : this()
...
             base.CopyFrom(obj);
-            if (!(obj is Trigger))
+            if (obj is not Trigger t)
             {
                 return;
             }
-
-            var t = (Trigger)obj;
+            
             DateTime = t.DateTime;
```
minichma: Clone DateTime as well?

minichma (dismissed):
LGTM, just minor comments.

The following classes don't have `CopyFrom` methods, because they are expected to have all their data stored in the `Properties` collection. Did only a very rough check to verify, this actually is the case.

```
Ical.Net.Calendar
Ical.Net.VTimeZoneInfo
Ical.Net.DataTypes.EncodableDataType
Ical.Net.CalendarComponents.Alarm
Ical.Net.CalendarComponents.CalendarEvent
Ical.Net.CalendarComponents.FreeBusy
Ical.Net.CalendarComponents.Journal
Ical.Net.CalendarComponents.RecurringComponent
Ical.Net.CalendarComponents.Todo
Ical.Net.CalendarComponents.UniqueComponent
Ical.Net.CalendarComponents.VTimeZone
Ical.Net.CalendarComponents.VTimeZone.IntervalRecurrencePattern
```

Ical.Net/DataTypes/Attachment.cs
axunonb: Sure, good point

Ical.Net/DataTypes/Period.cs
axunonb: Yes oversaw to Copy<IDateTime>()

Ical.Net/DataTypes/StatusCode.cs
axunonb: Absolutely!

Ical.Net/DataTypes/Trigger.cs
axunonb: Yep

PR #621, opened 2024-10-20
RRULE Evaluation: Apply `BYYEARDAY=366` only in leap years.

minichma:
Fix for the following cases of #618:
* `RRULE:FREQ=YEARLY;BYYEARDAY=-366;COUNT=3`
* `RRULE:FREQ=YEARLY;BYYEARDAY=366;COUNT=3`

I.e. `BYYEARDAY=366` should only exist if day 366 (or -366) exists, which is only the case in leap years.

### DIFF
--- a/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
@@ -467,7 +467,9 @@ namespace Ical.Net.Evaluation
                     var date1 = date;
                     yearDayDates.AddRange(pattern.ByYearDay.Select(yearDay => yearDay > 0
                         ? date1.AddDays(-date1.DayOfYear + yearDay)
-                        : date1.AddDays(-date1.DayOfYear + 1).AddYears(1).AddDays(yearDay)));
+                        : date1.AddDays(-date1.DayOfYear + 1).AddYears(1).AddDays(yearDay))
+                        // Ignore the BY values that don't fit into the current year (i.e. +-366 in non-leap-years).
+                        .Where(d => d.Year == date1.Year));
                 }
                 return yearDayDates;
             }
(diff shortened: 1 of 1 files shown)


Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -467,7 +467,8 @@ private List<DateTime> GetYearDayVariants(List<DateTime> dates, RecurrencePatter
                     var date1 = date;
                     yearDayDates.AddRange(pattern.ByYearDay.Select(yearDay => yearDay > 0
                         ? date1.AddDays(-date1.DayOfYear + yearDay)
-                        : date1.AddDays(-date1.DayOfYear + 1).AddYears(1).AddDays(yearDay)));
+                        : date1.AddDays(-date1.DayOfYear + 1).AddYears(1).AddDays(yearDay))
```
axunonb, line 470: Maybe this method in general and the changes deserves some comments?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -483,7 +484,7 @@ private List<DateTime> GetYearDayVariants(List<DateTime> dates, RecurrencePatter
                         ? date.AddDays(-date.DayOfYear + yearDay)
                         : date.AddDays(-date.DayOfYear + 1).AddYears(1).AddDays(yearDay);
 
-                    if (newDate.DayOfYear == date.DayOfYear)
+                    if (newDate.Date == date.Date)
```
axunonb, line 488: The outer loop has no code coverage right now, assume this will follow after merging #617?

axunonb (dismissed):
Very good, thanks

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
minichma: Yes, this was the idea :-)

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
minichma: Absolutely. Added a comment to explain my changes for now.

minichma:
@axunonb I think you need to approve again. Thanks!

axunonb (approved):
Re-approved, thanks

PR #624, opened 2024-10-25
Add GitHub workflow for code coverage calculation

axunonb:
The coverage workflow triggers for pull request and push to main.
It will not stop CI in case of failure for now. Coverage is processed on [Codecov](https://app.codecov.io/github/ical-org/ical.net/tree/main/Ical.Net).
Should be integrated into the test workflow, if everything works fine.
[That's the calculation](https://app.codecov.io/github/axunonb/ical.net/tree/main/Ical.Net) from my ical.net fork.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -3,14 +3,13 @@
 # The report is stored as artifact, and uploaded to Codecov.
 
 on:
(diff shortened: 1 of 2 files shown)


minichma (dismissed):
Very nice!

PR #625, opened 2024-10-26
Fix BOM causing error on line 1 in coverage.yml

axunonb:


axunonb:
Okay, after the fix it works https://app.codecov.io/github/ical-org/ical.net/tree/main

Issue #623, opened 2024-10-24
Daylight saving transition bugs


RemcoBlok:
In Ical.Net 4.3.1 I see bugs when getting occurrences for an event around the daylight saving transition when the clock
goes an hour forward or back.

For instance the following test currently fails when the clock goes an hour forward so that between 00:00 and 02:00 a single hour elapsed.

```cs
        [Fact]
        public void ClockGoingForwardTest()
        {
            // Arrange
            string tzId = "Europe/London";

            CalendarEvent myEvent = new CalendarEvent();
            myEvent.Start = new CalDateTime(2025, 3, 30, 0, 0, 0, tzId);
            myEvent.End = new CalDateTime(2025, 3, 30, 2, 0, 0, tzId);

            Calendar calendar = new Calendar();
            calendar.Events.Add(myEvent);

            // Act
            IDateTime start = new CalDateTime(2025, 3, 30, 0, 0, 0, tzId);
            IDateTime end = new CalDateTime(2025, 3, 31, 0, 0, 0, tzId);

            ICollection<Occurrence> occurrences = calendar.GetOccurrences<CalendarEvent>(start, end);

            // Assert
            Assert.Equal(1, myEvent.Duration.TotalHours);

            Assert.Single(occurrences);
            Occurrence occurrence = occurrences.Single();

            Assert.Same(myEvent, occurrence.Source);

            Assert.Equal(myEvent.Duration, occurrence.Period.Duration);
            Assert.Equal(myEvent.Start, occurrence.Period.StartTime);
            Assert.Equal(myEvent.End, occurrence.Period.EndTime);
        }
```

The test fails when asserting

```cs
            Assert.Equal(myEvent.End, occurrence.Period.EndTime);
```

with

```
Assert.Equal() Failure: Values differ
Expected: 30/03/2025 02:00:00 Europe/London
Actual:   30/03/2025 01:00:00 Europe/London
```

Similarly, the following test currently fails when the clock goes back an hour so that between 01:00 and 02:00 two hours elapsed. Note that 01:00 is an ambiguous time, but iCal.Net should use the first occurrence of 01:00 (as NodaTime does by default), therefore two hours elapse between 01:00 and 02:00.

```cs

        [Fact]
        public void ClockGoingBackTest()
        {
            // Arrange
            string tzId = "Europe/London";
            
            CalendarEvent myEvent = new CalendarEvent();
[...]

Issue #588, opened 2024-07-02
RecurrencePattern.Until DateTime property should be a CalDateTime


rmtexas:
iCalendar spec says that the UNTIL format should match the same format as DTSTART.  And if DTSTART specifies a TZID then the UNTIL should as well.
Because UNTIL is just a DateTime, there is no way to specify the timezone that should be applied the property.
So serialization of a recurrence comes out incorrectly.
Example:
BEGIN:VEVENT
DTEND;TZID=Central Standard Time:20190208T133000
DTSTAMP:20240702T173339Z
DTSTART;TZID=Central Standard Time:20190208T130000
RRULE:FREQ=DAILY;UNTIL=20190215T130000;WKST=SU
SEQUENCE:0
SUMMARY:Intro 1 : Recurring Appointment
UID:1
END:VEVENT

Notice that UNTIL has no TZID identifier applied. This is against spec.

RemcoBlok:
Similarly the EXDATE also does not serialize a TZID. See https://github.com/ical-org/ical.net/issues/614

RemcoBlok:
The way I read the standard, UNTIL must have the same value type as DTSTART.  If DTSTART is a DATE,  then UNTIL must be a DATE. If DTSTART is a DATE-TIME,  then UNTIL must be a DATE-TIME. Oddly, when DTSTART specifies a TZID, UNTIL must be in UTC, and not in the TZID of DTSTART. See https://icalendar.org/iCalendar-RFC-5545/3-3-10-recurrence-rule.html.

"If the DTSTART property is specified as a date with UTC time **or a date with local time and time zone reference**, then the UNTIL rule part MUST be specified as a date with UTC time"

It does not say the same thing about EXDATE having to be in UTC by the way, which is nice and consistent. See https://icalendar.org/iCalendar-RFC-5545/3-8-5-1-exception-date-times.html. For EXDATE you must be able to specify a TZID.

minichma:
I agree that `Until` should not be specified as `DateTime`, as it is rather ambiguous (see #638). However, I fully agree with @RemcoBlok that according to the RFC, `UNTIL` MUST be specified either in UTC or as floating time (without TZ).

axunonb:
Agree - now what does it mean in code? Only use UTC without implementing a floating time representation?
For `CalDateTime` in #638 a date/time with TZ `null` finally resolves to system's local time. Should TZ `null` be a floating time rather?
[...]

PR #627, opened 2024-10-28
Improve recurrence handling, remove restrictions

### LINKED ISSUE #622: Get Occurrences fails on a RRULE FREQ=SECONDLY
### Discussed in https://github.com/ical-org/ical.net/discussions/615

<div type='discussions-op-text'>

<sup>Originally posted by **DaleyKD** October 18, 2024</sup>
Does this library support getting occurrences if we use a secondly recurrence pattern?

Our project doesn't use this for calendar events, but more to help us determine when to run the next job (as part of our job scheduler). And SOME jobs need to run every 45 seconds or so.

I've been trying to get it to work, but have had no luck. It seems to work for FREQ=MINUTELY.</div>

axunonb:
- Deprecated `RecurrenceRestrictionType` and `RecurrenceEvaluationModeType` enums.

Updated RecurrencePattern.cs:
- Deprecated `RestrictionType` and `EvaluationMode` properties.
- Set default for `RestrictionType`.

Fix RecurrenceUtil:
- None of `GetOccurrences()` overloads convert `IDateTime` arguments with `AsSystemLocal`. This partial conversion is considered a bug. **The change may break user code that contains a fix for the bug.**

Refined RecurrencePatternEvaluator:
- Convert existing Java style comments for methods to XML documentation comments.

Updated Calendar.cs:
- Deprecated properties and methods associated with `RestrictionType` and `EvaluationMode`
- None of `GetOccurrences()` overloads convert `IDateTime` arguments with `AsSystemLocal`. This partial conversion is considered a bug. **The change may break user code that contains a fix for the bug.**
- Removed obsolete `Evaluate` method (just threw when called)

Enhanced RecurrenceTests.cs with more robust test cases:
- Replaced `Assert.That` with `Assert.Multiple`.
- Changed assertion for occurrences count.
- Renamed test methods for clarity.

Corrected typos and improved formatting for better readability.

Fixes #622

### DIFF
--- a/Ical.Net/Constants.cs
+++ b/Ical.Net/Constants.cs
@@ -219,6 +219,7 @@ namespace Ical.Net
         Fifth = 5
     }
 
+    [Obsolete("Usage may cause undesired results or exceptions. Will be removed.", false)]
     public enum RecurrenceRestrictionType
     {
         /// <summary>
@@ -247,6 +248,7 @@ namespace Ical.Net
         RestrictHourly
     }
 
+    [Obsolete("Usage may cause undesired results or exceptions. Will be removed.", false)]
     public enum RecurrenceEvaluationModeType
     {
         /// <summary>
--- a/Ical.Net/DataTypes/RecurrencePattern.cs
+++ b/Ical.Net/DataTypes/RecurrencePattern.cs
@@ -84,6 +84,11 @@ namespace Ical.Net.DataTypes
 
         public DayOfWeek FirstDayOfWeek { get; set; } = DayOfWeek.Monday;
 
+        /// <summary>
+        /// The type of restriction to apply to the evaluation of this recurrence pattern.
+        /// Returns <see cref="RecurrenceRestrictionType.NoRestriction"/> if not set.
+        /// </summary>
+        [Obsolete("Usage may cause undesired results or exceptions. Will be removed.", false)]
         public RecurrenceRestrictionType RestrictionType
         {
             get
@@ -93,11 +98,12 @@ namespace Ical.Net.DataTypes
                 {
                     return _restrictionType.Value;
                 }
-                return Calendar?.RecurrenceRestriction ?? RecurrenceRestrictionType.Default;
+                return RecurrenceRestrictionType.NoRestriction;
             }
             set => _restrictionType = value;
         }
 
+        [Obsolete("Usage may cause undesired results or exceptions. Will be removed.", false)]
         public RecurrenceEvaluationModeType EvaluationMode
         {
             get
--- a/Ical.Net/DataTypes/RecurrencePattern.cs
+++ b/Ical.Net/DataTypes/RecurrencePattern.cs
@@ -15,9 +15,10 @@ namespace Ical.Net.DataTypes
     public class RecurrencePattern : EncodableDataType
     {
         private int _interval = int.MinValue;
+#pragma warning disable 0618
         private RecurrenceRestrictionType? _restrictionType;
         private RecurrenceEvaluationModeType? _evaluationMode;
-
+#pragma warning restore 0618
         public FrequencyType Frequency { get; set; }
 
         private DateTime _until = DateTime.MinValue;
@@ -84,6 +85,7 @@ namespace Ical.Net.DataTypes
 
         public DayOfWeek FirstDayOfWeek { get; set; } = DayOfWeek.Monday;
 
+#pragma warning disable 0618
         /// <summary>
         /// The type of restriction to apply to the evaluation of this recurrence pattern.
         /// Returns <see cref="RecurrenceRestrictionType.NoRestriction"/> if not set.
@@ -98,7 +100,7 @@ namespace Ical.Net.DataTypes
                 {
                     return _restrictionType.Value;
                 }
-                return RecurrenceRestrictionType.NoRestriction;
+                return Calendar?.RecurrenceRestriction ?? RecurrenceRestrictionType.Default;
             }
             set => _restrictionType = value;
         }
@@ -117,6 +119,7 @@ namespace Ical.Net.DataTypes
             }
             set => _evaluationMode = value;
         }
+#pragma warning restore 0618
 
         public RecurrencePattern()
         {
@@ -148,8 +151,10 @@ namespace Ical.Net.DataTypes
         }
 
         protected bool Equals(RecurrencePattern other) => (Interval == other.Interval)
+#pragma warning disable 0618
             && RestrictionType == other.RestrictionType
             && EvaluationMode == other.EvaluationMode
+#pragma warning restore 0618
             && Frequency == other.Frequency
             && Until.Equals(other.Until)
             && Count == other.Count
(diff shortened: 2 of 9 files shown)


axunonb:
@minichma 
Sorry for trashing your inbox with this PR.
I couldn't reproduce the timeouts locally, so it was a trial and error work to get them fixed.
The `RecurrenceUtil.RecurrenceTimeout` now has a default of `RecurrenceUtil.DefaultTimeout` (1000ms), which is increased to 1500ms for some test. This is really a lot for calculating recurrences, isn't it?

minichma:
@axunonb no worries 😄 Didn't have time to go into detail so far, just my two cents on timeouts in general. I think it would be a good idea to keep the behavior of the algorithms deterministic. By introducing a timeout, we lose determinism, which can cause quite some issues of different kind. E.g. debugging is getting more challenging, as the timeout will expire quickly after a breakpoint was hit. More importantly the libs will behave differently depending on the system they are running on and the load that system currently experiences. So on server systems people might start seeing exceptions at peak hours in code that normally runs just fine. So all in all I would rather discourage the implementation of time-based behavior. I think the limitation based on increments was not too bad, because its fully deterministic, but certainly should be configurable. An alternative could be to allow users to configure a callback that's invoked frequently (probably on every increment). Users can then implement a timeout on their own. Or maybe pass a CancellationToken to the evaluation function that allows the caller to cancel based on a timeout, though this the callback would be preferable, as the CancellationToken wouldn't allow implementing deterministic behavior either.

axunonb:
Thanks for the quick feedback!

"Deterministic algorithms" is a valid point, agree.
After disabling `RecurrenceRestrictionType` and `RecurrenceEvaluationModeType` there were several tests running for ever. This was resolved with the timeout. But of course these unpredictable results are the downside that I dislike.

1. **Configurable number of increments:**
   * Con:`GetOccurrences` will block for an unpredictable time, if not configured properly (forever, with default "infinite"). 
   * Con: With recurrences not known at compile time, configuration is hard, may lead to unexpected results.
   * Pro: simple.
   * How to implement: An optional "max increments" argument to `GetOccurrences`, with "infinite" as the default.
2. **Callback:** 
    * Con: GetOccurrences` will block for an unpredictable time, maybe forever, if not configured properly.
    * Con: A bit more complex for users.
    * Pro: On top of option 1, it gives the user the chance to cancel. 
    * How to implement: An optional callback argument to `GetOccurrences`, meaning "infinite" if not set differently by the user.
3. Not to forget: Return `IEnumberable<Occurrence>` instead of `HashSet<Occurrence>`

Really hard to decide. What would you prefer?

minichma:
> Not to forget: Return IEnumberable<Occurrence> instead of HashSet<Occurrence>

This should be implemented anyways, because there's really no need (from a user's perspective) to iterate over all occurrences if just a few are needed. And more importantly, AFAIK the order in which HashSet returns items is not defined, so the user would have to sort them before using. But this would just be an addition, not an alternative to 1/2, right? Because if no occurrences would be returned, it could still run forever.

From the other two options I would probably prefer 1, because its very straightforward. Question would be, how to do the configuration. I'd suggest to pass something like `EvaluationOptions` to `GetOccurrences` (probably default to `null`), which can contain all the params that should be adjustable, also for the future.

That GetOccurrences could block for an indefinite amount of time is not something I see critical. To some degree we could say, this is how computers work. The time processing takes is most often dependent on the CPU and system load, so this is not something special here. Its not on us to decide how long a task is supposed to take. And generally the duration should be low. If its noticable, there's something wrong anyways.

One other thought I have is whether it should be possible that the iteration runs for a noticable amount of time at all. It should probably be possible to simplify the whole algorithm in a way, that it will only process time values that will eventually be returned to the user (exception: very large `COUNT` or `SETPOS`, but those should be forbidden anyways), which would prevent the whole thing from blocking. But that's probably a bigger story.

axunonb:
I'm fine going this way, thanks. `EvaluationOptions` will give us the flexibility for any future extensions. I'll work on an update. Except for the last paragraph, which is a bigger story indeed.

axunonb:
@minichma 
The PR now serves the following targets:
1. Mark all code related to `RecurrenceRestrictionType` and `RecurrenceEvaluationModeType` enums as obsolete. 
2. Update test methods accordingly
3. Close #622

What if we merged it and published a new minor version?
This would be a bug fix release for 2 issues.
Using it in existing code has no effect and is better than before, because restrictions were widely broken. We explain details in the release notes.

Ical.Net/Evaluation/RecurrenceUtil.cs
```diff
@@ -15,7 +15,7 @@ public static void ClearEvaluation(IRecurrable recurrable)
         }
 
         public static HashSet<Occurrence> GetOccurrences(IRecurrable recurrable, IDateTime dt, bool includeReferenceDateInResults) => GetOccurrences(recurrable,
-            new CalDateTime(dt.AsSystemLocal.Date), new CalDateTime(dt.AsSystemLocal.Date.AddDays(1).AddSeconds(-1)), includeReferenceDateInResults);
```
axunonb, line 18: Converting to local time would only be correct, when dealing with Floating Time: If DTSTART does not include a TZID parameter and does not end with a Z, the event is defined in the local time of the calendar user.

axunonb:
@minichma Do you think this PR needs some more tasks to be completed? Otherwise I'd like to proceed [as described](https://github.com/ical-org/ical.net/pull/627#issuecomment-2442804417).

Ical.Net/Constants.cs
```diff
@@ -219,6 +219,7 @@ public enum FrequencyOccurrence
         Fifth = 5
     }
 
+    [Obsolete("Usage may cause undesired results or exceptions. Will be removed.", false)]
```
minichma, line 222: We receive warnings at usage sites now (e.g. RecurrrencePatternEvaluator). We should suppress those warnings there.

Ical.Net/DataTypes/RecurrencePattern.cs
```diff
@@ -93,11 +98,12 @@ public RecurrenceRestrictionType RestrictionType
                 {
                     return _restrictionType.Value;
                 }
-                return Calendar?.RecurrenceRestriction ?? RecurrenceRestrictionType.Default;
+                return RecurrenceRestrictionType.NoRestriction;
```
minichma: That's a breaking change I assume, right?

Ical.Net/Evaluation/RecurrenceUtil.cs
minichma: Even then `AsSystemLocal` might not be correct, because the end user's time zone might come from a completely different context than the system's time zone, in particular when using the lib on a web server. We might be processing calendars received via the web and in this case the systems local time would easily be different from the user's, depending on whether it is/can be derived form the request.

Anyhow, this seems to be a breaking change. I think it's a bugfix, therefore it should be fine. Just to be aware, that some users might have adapted to that behavior.

Related to #623, #588, #19, https://github.com/orgs/ical-org/projects/3?pane=issue&itemId=85730174

minichma (dismissed):

minichma:
> Do you think this PR needs some more tasks to be completed? Otherwise I'd like to proceed https://github.com/ical-org/ical.net/pull/627#issuecomment-2442804417.

I think it's good as is. Left just a few minor comments.

Ical.Net/Evaluation/RecurrenceUtil.cs
axunonb: Yes, this is potentially a breaking change, even when it's rather a fix. This must go into release notes for sure.

Ical.Net/Constants.cs
axunonb: right, will add #pragma warning disable 0618 where needed

Ical.Net/DataTypes/RecurrencePattern.cs
axunonb: That should **not** have been left in the code. Thanks!
I will revert the change and set `RecurrenceRestrictionType.NoRestriction` in the tests that would fail otherwise.

PR #638, opened 2024-11-09
Enhance CalDateTime and DateTimeSerializer

### LINKED ISSUE #197: GetOccurrences should not assume the system local date time
RecurrenceUtil uses `AsSystemLocal` when taking in an `IDateTime`. (line 18) That's bad, because it loses information, and we have NodaTime backing the search, which should make working with time zones easy. In order, these should be the rules for time zone ambiguity:

1. The `CalDateTime`'s tzid, if specified
2. Otherwise the `Event`'s tzid, if specified
3. Otherwise the local system's time zone

## Original ticket:

```
   /// <summary>
      /// New zealand time zone should return occurrence but fails for tomorrow.
      /// </summary>
      [Test]
      public void NewZealandTimeZoneShouldReturnOneOccurrenceForTomorrow()
      {
         const string ical =
     @"BEGIN:VCALENDAR
PRODID:-//AxoNet Software GmbH//NONSGML Lights-Out//EN
VERSION:2.0
CALSCALE:GREGORIAN
METHOD:PUBLISH
BEGIN:VTIMEZONE
TZID:New Zealand Standard Time
BEGIN:STANDARD
TZNAME:New Zealand Standard Time
TZOFFSETTO:+1200
TZOFFSETFROM:+1300
RRULE:FREQ=YEARLY;BYDAY=SU;BYMONTH=4;BYSETPOS=1;WKST=MO
DTSTART:16010101T030000
END:STANDARD
BEGIN:DAYLIGHT
TZOFFSETFROM:+1200
TZOFFSETTO:+1300
DTSTART:20080101T020000
RRULE:FREQ=YEARLY;BYDAY=SU;BYMONTH=9;BYSETPOS=-1;WKST=MO
TZNAME:New Zealand Daylight Time
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
DTSTART;TZID=New Zealand Standard Time:20161212T164500
DTEND;TZID=New Zealand Standard Time:20161212T210002
SUMMARY:Server
DESCRIPTION:
UID:36a33ca8-92ba-49bc-a26a-712873030788
X-MICROSOFT-CDO-BUSYSTATUS:BUSY
BACKGROUND:BUSY
RRULE:FREQ=DAILY
END:VEVENT
END:VCALENDAR
";

         var collection = Calendar.LoadFromStream(new StringReader(ical));
         var checkAgainst = new DateTime(2016, 12, 13);
         var occurrences = collection.GetOccurrences<Event>(checkAgainst);

         Assert.IsTrue(occurrences.Count.Equals(1));
      }

```
Using ical.net 2.2.18.

This is a simple daily recurrence starting today (12/12). Therefore I expect to get one occurrence for each day (today,  tomorrow, the day after tomorrow, etc).

It works for other time zones but fails for New Zealand TZ (returns no occurrence after today).

### LINKED ISSUE #630: Unrepresentale `DateTime` arguments when deserializing should throw `ArgumentOutOfRangeException`

### LINKED ISSUE #633: Initializing a `CalDateTime` instance with midnight time is treated as `date-only`

### LINKED ISSUE #635: Period.StartTime returns copy of `IDateTime` with wrong `HasTime` setting
https://github.com/ical-org/ical.net/blob/95bbc53991009e64c208b827f3c89bd49680c6bc/Ical.Net/DataTypes/Period.cs#L101-L106

The new `CalDateTime` must have `HasTime` set to `false` explicitly:
```C#
new CalDateTime(new DateTime(_startTime.Value.Year, _startTime.Value.Month, _startTime.Value.Day, 
     0, 0, 0), _startTime.TzId) { HasTime = false };
```

### LINKED ISSUE #636: `CalDateTime` equality operator has missing comparison for `HasTime`
While `HasDate` is present, `HasTime` is missing

https://github.com/ical-org/ical.net/blob/95bbc53991009e64c208b827f3c89bd49680c6bc/Ical.Net/DataTypes/CalDateTime.cs#L184-L193

### LINKED ISSUE #637: Methods adding intervals to `CalDateTime` should have identical logic for setting the `IDateTime` result
In the `CalDateTime` class, methods that return an `IDateTime` by adding ticks, minutes, seconds, and other time intervals will set the `IDateTime.HasTime` property to `true` if the resulting date-time value is not a multiple of full days. This ensures that the `IDateTime` instance correctly reflects whether it includes a time component.

The methods are:
* AddTicks(long ticks)
* AddMilliseconds(int milliseconds)
* AddSeconds(int seconds)
* AddMinutes(int minutes)
* AddHours(int hours)
* AddDays(int days)
* AddMonths(int months)
* AddYears(int years)
* Add(TimeSpan ts)
* Subtract(TimeSpan ts)

These methods should set the `IDateTime.HasTime` property to `true` if the resulting date-time value includes a time component. Specifically, if the added interval results in a date-time value that is not a multiple of full days (i.e., it includes hours, minutes, seconds, or smaller units), the `HasTime` property should be set to true. This ensures that the `IDateTime` instance accurately represents whether it includes a time component.

```C#
CalDateTime dateTime = new CalDateTime(new DateTime(2023, 10, 1));
dateTime.HasTime = false;

// Adding 1 hour to the date-time value
IDateTime newDateTime = dateTime.AddHours(1);

// The resulting date-time value includes a time component, so HasTime is set to true
Assert.That(newDateTime.HasTime, Is.True);
```

axunonb:
Breaking Changes
* Removef the redundant `Calendar` argument to constructor overload of `CalDateTime`
* `CalDateTilme` only uses `DateTilmeKind.Utc` and `DateTimeKind.Unspecified`. `DateTimeKind.Local` has been dropped.
* Setting `null` for the timezone will apply the system`s local time with`AsDateTimeOffset`.
* The given timezone always takes precedence over any `DateTimeKind` of a `DateTime` value. This is also the case, if a UTC date and a timezone (not null) is used.
* `CalDateTime.TzId`: Will not read or write `CalDateTime.Parameters`. The functionality has been moved to `DateTimeSerializer.Deserialize(TextReader)
* `DateTimeSerializer.Deserialize(TextReader)` will now throw for unrepresentable date/time values.
* `CalDateTime.TzId` can be `null` or a valid timezone ID. Whitespace is not allowed.

CalDateTime
* Implement`DateOnly` and `TimeOnly` in order to return correct values for `CalDateTime.HasTime` and `CalDateTime.HasDate`, even if hour, minutes and seconds are zero.
* Updating `Value` or `TzId` will re-initialize the instance in the same way as with CTOR overloads.
* Refactored `ToString` overloads to convert the date and time to a string representation. Thanks to @nzbart for his initial PR.
* marked some properties as obsolete. 
* Updated related tests and methods to accommodate these changes.
* Enable NRT
* Complete xmldoc
* Increase code coverage to 94%

DateTimeSerializer
* Added validation in the `SerializeToString` method to ensure the TZID parameter is correctly handled based on whether the date-time is in UTC.
* Improved the Deserialize method to handle both full date-time and date-only strings, initializing the `DateOnly` and `TimeOnly` parts accordingly.
* Enable NRT
* Complete xmldoc
* Increase code coverage to 94%

Other
* Enable NRT for `IDateTime`
* Added new package reference `Portable.System.DateTimeOnly` for netstandard2.0 and netstandard2.1 for compatibility with net6.0 and later.

Resolves #630
Resolves #633
Resolves #635
Resolves #636
Resolves #637
Resolves #197


Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -7,45 +7,6 @@
...
-    /// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
-    /// A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
-    /// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
-    /// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
-    /// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
-    /// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
-    /// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
-    /// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
-    /// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-    /// </summary>
```
axunonb: Comparing our code base with the current iCal4j code its hard to say wether there are similarities left. The mention goes back to 2009. Even more important is, that their BSD 3-Clause License is not compatible to our MIT License.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -303,52 +303,12 @@ public override int GetHashCode()
...
-                        ? Value
-                        : Value.Date;
-                }
-
-                return HasTime
-                    ? Value.ToLocalTime()
-                    : Value.ToLocalTime().Date;
-            }
-        }
+        public DateTime AsSystemLocal => AsDateTimeOffset.LocalDateTime;
```
axunonb: Use existing code from AsSystemLocal

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -303,52 +303,12 @@ public override int GetHashCode()
...
-                {
-                    asUtc = DateTime.SpecifyKind(Value, DateTimeKind.Utc);
-                    return asUtc;
-                }
-
-                asUtc = DateTime.SpecifyKind(Value, DateTimeKind.Local).ToUniversalTime();
-                return asUtc;
-            }
-        }
+        public DateTime AsUtc => AsDateTimeOffset.UtcDateTime;
```
axunonb: Use existing code from AsSystemLocal

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -533,7 +493,7 @@ public IDateTime ToTimeZone(string? tzId)
         /// </summary>
         public DateTimeOffset AsDateTimeOffset =>
             string.IsNullOrWhiteSpace(TzId)
-                ? new DateTimeOffset(AsSystemLocal)
+                ? new DateTimeOffset(Value)
```
axunonb: Value contains the right `DateTimeKind` to create the `DatTimeOffset`

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -169,146 +230,174 @@ public override int GetHashCode()
...
+        /// <summary>
+        /// Toggles the <see cref="DateTime.TimeOfDay"/> part of the underlying <see cref="Value"/>.
+        /// <see langword="true"/> if the underlying <see cref="DateTime"/> <see cref="Value"/> has a 'time' part (hour, minute, second).
+        /// </summary>
         public bool HasTime
         {
-            get => _hasTime;
-            set => _hasTime = value;
+            get => _timeOnly.HasValue;
+            set => _timeOnly = value ? TimeOnly.FromDateTime(_value) : null;
```
axunonb: We can toggle on/off this way

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -169,146 +230,174 @@ public override int GetHashCode()
...
+        /// <summary>
+        /// Toggles the <see cref="DateTime.TimeOfDay"/> part of the underlying <see cref="Value"/>.
+        /// <see langword="true"/> if the underlying <see cref="DateTime"/> <see cref="Value"/> has a 'date' part (year, month, day).
+        /// </summary>
         public bool HasDate
         {
-            get => _hasDate;
-            set => _hasDate = value;
+            get => _dateOnly.HasValue;
+            set => _dateOnly = value ? DateOnly.FromDateTime(_value) : null;
```
axunonb: We toggle this, too, just to make IDateTime happy

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -325,63 +414,65 @@ public string TzId
                     return;
                 }
 
-                _asUtc = DateTime.MinValue;
-
-                var isEmpty = string.IsNullOrWhiteSpace(value);
-                if (isEmpty)
+                if (string.IsNullOrWhiteSpace(value))
                 {
+                    Initialize(_value, _timeOnly.HasValue, value, Calendar);
```
axunonb: Changing the TzId now initializes with immediate effect

Ical.Net/DataTypes/Period.cs
```diff
@@ -101,9 +104,7 @@ private void ExtrapolateTimes()
         private IDateTime _startTime;
         public virtual IDateTime StartTime
         {
-            get => _startTime.HasTime
-                ? _startTime
-                : new CalDateTime(new DateTime(_startTime.Value.Year, _startTime.Value.Month, _startTime.Value.Day, 0, 0, 0), _startTime.TzId);
+            get => _startTime;
```
axunonb: _startTime is ready to use by default

Ical.Net/DataTypes/Period.cs
```diff
@@ -137,7 +138,7 @@ public virtual TimeSpan Duration
             {
                 if (StartTime != null
                     && EndTime == null
-                    && StartTime.Value.TimeOfDay == TimeSpan.Zero)
```
axunonb: must use HasTime

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -169,146 +230,174 @@ public override int GetHashCode()
...
+                // as the value can be the same but the Kind different.
                 if (_value == value && _value.Kind == value.Kind)
                 {
                     return;
                 }
 
-                _asUtc = DateTime.MinValue;
-                _value = value;
+                // Initialize with the new value, keeping current 'HasTime' setting
+                Initialize(value, _timeOnly.HasValue, TzId, Calendar);
```
axunonb: Setting the value initializes in the same as with using the CTOR overloads

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -1,130 +1,185 @@
...
 /// In addition to the features of the <see cref="DateTime"/> class, the <see cref="CalDateTime"/>
-/// class handles time zone differences, and integrates seamlessly into the iCalendar framework.
+/// class handles time zones, and integrates seamlessly into the iCalendar framework.
 /// </remarks>
 /// </summary>
 public sealed class CalDateTime : EncodableDataType, IDateTime
 {
+    // The date and time parts that were used to initialize the instance
+    // or by the Value setter.
+    private DateTime _value;
```
minichma: I think, we shouldn't maintain both, the DateTime value *and* Date-/TimeOnly. I suggest to remove `_value` altogether.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -1,130 +1,185 @@
...
+    private void Initialize(DateTime dateTime, bool hasTime, string? tzId, Calendar? cal)
     {
-        Initialize(CoerceDateTime(year, month, day, hour, minute, second, DateTimeKind.Local), tzId, cal);
-    }
+        DateTime initialValue;
 
-    private void Initialize(DateTime value, string tzId, Calendar cal)
-    {
-        if (!string.IsNullOrWhiteSpace(tzId) && !tzId.Equals("UTC", StringComparison.OrdinalIgnoreCase))
+        if ((tzId != null && !string.IsNullOrWhiteSpace(tzId) && !tzId.Equals("UTC", StringComparison.OrdinalIgnoreCase))
```
minichma: I know, it's legacy (so nothing to necessarily deal with in this PR), but I think that we shouldn't have special handling for white space, or empty strings. Either the value is set (!= null) or not (== null). Input sanitization is not something this class should be responsible for.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -1,130 +1,185 @@
...
+            || (string.IsNullOrWhiteSpace(tzId) && dateTime.Kind == DateTimeKind.Local))
         {
             // Definitely local
-            value = DateTime.SpecifyKind(value, DateTimeKind.Local);
-            TzId = tzId;
+            _tzId = tzId;
+
+            initialValue = DateTime.SpecifyKind(dateTime, DateTimeKind.Local);
```
minichma: Also legacy, but IMHO important: I think we should avoid `DateTimeKind.Local` wherever we can, because dealing with the system's local time really is a special case. If no tzid is specified, this would usually mean that the instance represents floating time (which btw cannot be converted to a different time zone). Floating time should rather be represented by `DateTimeKind.Unspecified` than `.Local`.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Without the backing `value` we lose the `HasDate` and `HasTime` "toggle". The Date-/TimOnly fields are just the replacement for the backing `_hasDate` and `_hasTime` booleans. This implementation is mainly because `HasDate` and `HasTime` require a setter from the `IDateTime` interface.

minichma:
@axunonb I only had a very brief look at the new CalDateTime class. While I think that this PR brings some improvements, I feel that the whole design of the CalDateTime class deserves some more attention and improvements. Just a few thoughts:
* I feel that NodaTime's TimeOnly/DateOnly or maybe LocalDateTime would be better suitable for representing the inner value than System.DateTime, because its more clearly structured and doesn't have the ambiguity that comes from the `.Kind` field. Anyhow I recommend to have just either of them, not both.
* I think that CalDateTime in general should be immutable (maybe implemented as value type, though that's probably not that important).
* I'm not sure, whether we need the `IDateTime` interface, do we? `CalDateTime` seems to be the only implementation and the additional interface seems to rather introduce additional complexity than reducing it.
* Not sure, what the `AssociatedObj` property is used for but I feel that it shouldn't be required for CalDateTime. I understand that it's part of the overall Calendar model infrastructure, but maybe that's a core problem of the whole CalDateTime class, that it mixes up the representation of a value in the calendar model with the operations to be executed on top of it. In case of recurrence evaluation there's a rather clear separation between the model (e.g. RecurrencePattern) and the evaluation (RecurrencePatternEvaluator), but in the case of CalDateTime this doesn't exist, which can easily cause confusion. Maybe we should put some thoughts into how this separation could be improved. But I don't have a clear picture of how this would ideally look like either. Anyhow, as long as its used for representing the calendar model, the type should do that with as little ambiguity as possible. Any calculations on top should be performed just in time while the data of the original model should stay unchanged and represent the original value.

jm2c

Ical.Net/DataTypes/CalDateTime.cs
minichma: I thought it would be an option to calculate the DateTime value from the other two just in time, when needed.

axunonb: I think I get your point: `HasDate` and `HasTime` wouldn't have backing fields, they would just become pure boolean getters/setters. Then `_value` can disappear. Yep, that's better.

axunonb:
Todo in this PR:

### Use of `NodaTime` for Date and Time Representation

**Suggestion**: Use ~~NodaTime's `LocalDateTime`,~~ `DateOnly`, or `TimeOnly` instead of `System.DateTime`.

**Reasoning**: `NodaTime` provides a more precise and unambiguous representation of date and time values. `System.DateTime` can be ambiguous due to the `Kind` property, which can lead to confusion about whether the time is local, UTC, or unspecified.

**Action**: Consider refactoring `CalDateTime` to use `DateOnly` and`TimeOnly` types. This will make the class more robust and less prone to errors related to time zone handling.

**Timing**: In #638

### Necessity of `IDateTime` Interface

**Suggestion**: Evaluate the necessity of the `IDateTime` interface.

**Reasoning**: If `CalDateTime` is the only implementation of `IDateTime` in _ical.net_, the interface might be unnecessary and could introduce additional complexity without providing significant benefits.

**Action**: Review the usage of `IDateTime`. If it is not providing clear benefits, consider removing it and using `CalDateTime` directly.

**Timing**: Together with another breaking change. ~~In #638 mark `HasTime` and `HasDate` setters as obsolete.~~

### AssociatedObject Property

**Suggestion**: Re-evaluate the need for the `AssociatedObject` property in `CalDateTime`.

**Reasoning**: The `AssociatedObject` property might be mixing concerns by combining the representation of a date-time value with operations related to the calendar model. This can lead to confusion and make the class harder to maintain.

**Evaluation**: `AssociatedObject` is necessary during deserialization with `DateTimeSerializer` in order get the timezone for the `CalDateTime`. However, the `Calendar` argument in `CalDateTime` CTORs is never used.

**Action**: ~~Consider separating the concerns of date-time representation and calendar operations. This might involve creating separate classes or methods for operations that act on `CalDateTime` instances.~~ Remove the  the `Calendar` argument in `CalDateTime` CTORs.

**Timing**:  #638

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Will be updated as proposed.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Will be updated as proposed.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Will be updated as proposed.

axunonb: Done

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Done

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Done

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -174,271 +237,301 @@ public override int GetHashCode()
...
+    /// </summary>
+    /// <remarks>
+    /// This will also set <seealso cref="HasTime"/> to <see langword="true"/>,
+    /// if the <see cref="TimeSpan"/> is not a multiple of 24 hours.
+    /// </remarks>
     public static IDateTime operator -(CalDateTime left, TimeSpan right)
     {
-        var copy = left.Copy<IDateTime>();
+        var copy = left.Copy<CalDateTime>();
+        if (Math.Abs(right.TotalDays % 1) > AlmostZeroEpsilon)
```
minichma: Maybe `(right.Ticks % TimeSpan.TicksPerDay) != 0`? This way we could avoid dealing with FP precision issues.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -174,271 +237,301 @@ public override int GetHashCode()
...
-    {
-        get
-        {
-            if (Value.Kind == DateTimeKind.Unspecified)
-            {
-                return HasTime
-                    ? Value
-                    : Value.Date;
-            }
+    public DateTime AsSystemLocal => AsDateTimeOffset.LocalDateTime;
```
minichma, line 311: I think we should remove `AsSystemLocal` altogether. `AsDateTimeOffset.LocalDateTime` can also be executed by the caller if needed.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -1,130 +1,188 @@
...
 /// In addition to the features of the <see cref="DateTime"/> class, the <see cref="CalDateTime"/>
-/// class handles time zone differences, and integrates seamlessly into the iCalendar framework.
+/// class handles timezones, and integrates seamlessly into the iCalendar framework.
 /// </remarks>
 /// </summary>
 public sealed class CalDateTime : EncodableDataType, IDateTime
 {
-    public static CalDateTime Now => new CalDateTime(DateTime.Now);
+    // The date part that is used to return the Value property.
+    private DateOnly _dateOnly;
```
minichma, line 26: Shouldn't we define `_dateOnly` and `_timeOnly` as nullable? Then we wouldn't have to explicitly store `HasTime` and it would get become natural as

```cs
    public bool HasTime {
        get => this._timeOnly != null;
        set => this._timeOnly = (value ? (this._timeOnly ?? new TimeOnly()) : null);
    }
```

We would remove additional redundancy.

Ical.Net/Utility/DateUtil.cs
```diff
@@ -22,7 +22,7 @@ public static IDateTime EndOfDay(IDateTime dt)
         => StartOfDay(dt).AddDays(1).AddTicks(-1);
 
     public static DateTime GetSimpleDateTimeData(IDateTime dt)
-        => DateTime.SpecifyKind(dt.Value, dt.IsUtc ? DateTimeKind.Utc : DateTimeKind.Local);
+        => DateTime.SpecifyKind(dt.Value, dt.IsUtc ? DateTimeKind.Utc : DateTimeKind.Unspecified);
```
minichma: According to the improved CalDateTime, this would be as simple as `=> dt.Value`.

Ical.Net/Utility/DateUtil.cs
minichma, line 1: I think, there are a few more places to deal with local date in this class. 

* The behavior of `SimpleDateTimeToMatch` is not easy to understand. Within the sln its only used for testing, so I'd suggest to move it to a test project, where the specific behavior might be helpful.
* `MatchTimeZone` is only used in `RecurrencePatternEvaluator`. I'm pretty sure, that the way it deals with floating time is not the most natural one either (see #197). Probably we want to deal with that in a separate PR. Maybe move this method into `RecurrencePatternEvaluator` for now and/or add a comment that we should reconsider/fix how it deals with local time.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -1,130 +1,188 @@
...
+    private TimeOnly _timeOnly;
 
-    public static CalDateTime Today => new CalDateTime(DateTime.Today);
+    private const double AlmostZeroEpsilon = 1e-10;
+    private const string UtcTzId = "UTC";
+
+    /// <summary>
+    /// Gets the current date/time in the local timezone.
+    /// </summary>
+    public static CalDateTime Now => new CalDateTime(DateTime.Now, null, true);
```
minichma, line 35: If we want to offer a `Now`, then this is one of the very few places where referring to the system's local time would make sense.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -1,130 +1,188 @@
...
 public sealed class CalDateTime : EncodableDataType, IDateTime
 {
-    public static CalDateTime Now => new CalDateTime(DateTime.Now);
+    // The date part that is used to return the Value property.
+    private DateOnly _dateOnly;
+    // The time part that is used to return the Value property.
+    private TimeOnly _timeOnly;
 
-    public static CalDateTime Today => new CalDateTime(DateTime.Today);
+    private const double AlmostZeroEpsilon = 1e-10;
```
minichma: See my comment on `operator -`. I think, we can having this kind of epsilon.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -1,130 +1,188 @@
...
-    public CalDateTime(DateTime value, string tzId)
+    /// <param name="value"></param>
+    /// <param name="tzId">The specified value will override value's <see cref="DateTime.Kind"/> property.
+    /// If the timezone specified is UTC, the underlying <see cref="DateTime.Kind"/> will be
+    /// <see cref="DateTimeKind.Utc"/>. If a non-UTC timezone or no timezone is specified, the
+    /// <see cref="DateTimeKind.Unspecified"/> will be used. A timezone of <see langword="null"/> represents
+    /// the system's local timezone.
+    /// </param>
+    /// <param name="hasTime">Set to <see langword="true"/> (default), if the <see cref="DateTime.TimeOfDay"/> must be included.</param>
+    public CalDateTime(DateTime value, string? tzId, bool hasTime = true)
```
minichma, line 88: Having both, the `Kind` in the `DateTime` value, as well as the `tzId`, causes redundancy, that's not easy to understand. What, if we remove this constructor altogether? The user could easily call `new CalDateTime(value).ToTimeZone(tzId)`, with way less ambiguity.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -174,271 +237,301 @@ public override int GetHashCode()
...
     public DayOfWeek DayOfWeek => Value.DayOfWeek;
 
+    /// <inheritdoc cref="DateTime.DayOfYear"/>
     public int DayOfYear => Value.DayOfYear;
 
+    /// <inheritdoc cref="DateTime.Date"/>
     public DateTime Date => Value.Date;
 
+    /// <inheritdoc cref="DateTime.TimeOfDay"/>
     public TimeSpan TimeOfDay => Value.TimeOfDay;
```
minichma, line 447: Isn't this obsolete, now that we have the `TimeOnlyValue`?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -174,271 +237,301 @@ public override int GetHashCode()
...
     public long Ticks => Value.Ticks;
 
+    /// <inheritdoc cref="DateTime.DayOfWeek"/>
     public DayOfWeek DayOfWeek => Value.DayOfWeek;
 
+    /// <inheritdoc cref="DateTime.DayOfYear"/>
     public int DayOfYear => Value.DayOfYear;
 
+    /// <inheritdoc cref="DateTime.Date"/>
     public DateTime Date => Value.Date;
```
minichma, line 444: Obsolete with `DateOnlyValue`?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -174,271 +237,301 @@ public override int GetHashCode()
...
-    }
+    /// <summary>
+    /// Returns a representation of the <see cref="DateTime"/> in UTC.
+    /// </summary>
+    public DateTime AsUtc => AsDateTimeOffset.UtcDateTime;
+
+    /// <summary>
+    /// Gets the underlying <see cref="DateOnlyValue"/> of <see cref="Value"/>.
+    /// </summary>
+    public DateOnly? DateOnlyValue => HasDate ? _dateOnly : null;
```
minichma, line 321: Maybe just name it `Date`, in replacement of the existing, now obsolete `Date`?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -174,271 +237,301 @@ public override int GetHashCode()
...
+
+    /// <summary>
+    /// Gets the underlying <see cref="TimeOnlyValue"/> of <see cref="Value"/>.
+    /// </summary>
+    public TimeOnly? TimeOnlyValue => HasTime ? _timeOnly : null;
```
minichma, line 326: Just `Time` or `TimeOfDay`?

minichma (commented):
I love the simplification that comes with removing the ambiguous `DateTime` member field and with removing `DateTimeKind.Local`! Didn't do a full review, only `CalDateTime` for now, but I think that's the core part anyways. Left a few comments with my thoughts, maybe not all relevant for this PR but for following ones.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: So let's make a date/time without a timezone a "floating time". Then:
* `IsUtc` should return `false` - sounds reasonable
* `AsUtc` should throw, because a floating time (`TzId == null`) cannot be represented as UTC?
* `AsDateTimeOffset` should throw for the same reason
* and sure, `AsSystemLocal` must be removed, as it uses `TzId == null`
* add `IsFloating`?

Ical.Net/DataTypes/CalDateTime.cs
axunonb: The setter would then overwrite the original value that is needed in the current v4 implementation.
I did not yet investigate implications, and thus made it compatible.

In the end, `HasTime` and `HasDate`, `TzId` and `Value` should just have getters. This requires to change (or even remove entirely) `IDateTime`. That gets to huge for this PR.

minichma: The setter would only overwrite, if the value is changed true->false->true, which I would consider legit in that case. But I fully agree that HasTime/HasDate setters should be removed altogether and that this should be covered in different PR. Also the additional considerations regarding local time should be handled in a different PR, so this one doesn't get bloated too much.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: excellent, done

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Created issue  #646 for this task - will not include it into this PR

Ical.Net/Utility/DateUtil.cs
axunonb: right - done

Ical.Net/Utility/DateUtil.cs
axunonb: Done:
* Move `DateUtil.SimpleDateTimeToMatch` as private to `RecurrenceTests`
* Move `DateUtil.MatchTimeZone` as private  to `RecurrencePatternEvaluator`

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Must be done separately, because `AsSystemLocal` is a member of `IDateTime` - see new task in #646

Ical.Net/DataTypes/CalDateTime.cs
axunonb: This will be back on the agenda when we implement floating date/time. See #647

Ical.Net/Utility/DateUtil.cs
minichma: `DateUtil` is internal, so this shouldn't be breaking.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: see task in #646

Ical.Net/DataTypes/CalDateTime.cs
axunonb: see task in https://github.com/ical-org/ical.net/issues/646

Ical.Net/DataTypes/CalDateTime.cs
axunonb: see task in https://github.com/ical-org/ical.net/issues/646

Ical.Net/DataTypes/CalDateTime.cs
axunonb: see task in https://github.com/ical-org/ical.net/issues/646

axunonb:
@minichma Merged as you already reviewed twice :)

minichma:
Sorry for not being able to review this any earlier. Anyhow, this is a significant improvement, congrats!

PR #645, opened 2024-11-20
Fix some recurrence evaluation issues

minichma:
This PR fixes the following issues listed in #618:

> DTSTART:20141006T090000, RRULE:FREQ=WEEKLY;INTERVAL=1;COUNT=2;BYDAY=MO;

The trailing semicolon caused an exception and is now ignored. Additionally handling of illegal rule parts having a number of `=` chars that is different from 1. In such a case an `ArgumentException` is raised, which seems to be in line with other exceptions raised in that class but could be discussed.

> RRULE:FREQ=WEEKLY;UNTIL=20170127T000000Z;WKST=MO;BYDAY=SU,TU,TH;INTERVAL=2

The start of the week was not considered correctly.


Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -244,7 +244,20 @@ public override object Deserialize(TextReader tr)
...
+                        // This is illegal but ignore for now.
+                        continue;
+                    }
+
                     var keyValues = keywordPair.Split('=');
+                    if (keyValues.Length != 2)
+                    {
+                        // ArgumentExceptions seem to be the Exception of choise for this class. Should
+                        // probably be changed to a more specific exception type.
+                        throw new ArgumentException($"The recurrence rule contains an invalid part.");
```
axunonb: As we know it here, show the input value causing the exception (and maybe what's wrong with it)

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -244,7 +244,20 @@ public override object Deserialize(TextReader tr)
...
+                    if (keywordPair.Length == 0)
+                    {
+                        // This is illegal but ignore for now.
+                        continue;
+                    }
+
                     var keyValues = keywordPair.Split('=');
+                    if (keyValues.Length != 2)
+                    {
+                        // ArgumentExceptions seem to be the Exception of choise for this class. Should
```
axunonb, line 256: ArgumentException is fine for this PR. As we're aware now we could create an issue labelled enhancement to track implementing an Exception type for parsing and/or recurrence failures

axunonb (commented):
Great! Just a minor remark re Exception details

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
minichma: I was thinking that too but then intentionally left out the actual input data, since the data could contain sensitive information (in the end a calendar can be something very personal). I know that in other places we include the data, so if you think we should, I'll be happy to add it.

axunonb: For developers the generic message would be less helpful than the detailed one. So my vote is for the more helpful way :)

minichma: The old conflict ;-)

minichma: added some details

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
minichma: #648

axunonb (approved):

minichma:
@axunonb How should we deal with merging into v5? 3-way merge? Cherry-pick? Separate PR? I think I'd vote for the 3-way merge.

axunonb:
Yes, the 3-way merge would be beneficial. But would updating `main` besides `version/v5` aim for further v4 releases? Otherwise we could freeze v4.

minichma:
Oh, I was unsure, about your plans regarding v4. Given the creation of a v5 branch I thought `main` would continue to exist for v4 for now, but I agree that concentrating on v5 would be beneficial. But, if we don't plan for further v4 releases anyways, why not continue developing v5 on `main`? And in case we think another v4 is needed, we would create a v4 branch (or do that right now). That's how I know it from other project like libical too.

axunonb:
Really appreciate your input! Of course, that makes perfect sense.

axunonb:
Created `version/v4` and then merged `version/v5` into `main`

PR #652, opened 2024-11-25
Recurrence evaluation: Fix calculation of week numbering.

minichma:
Restore and fix one of the failing tests from #618. I.e. week numbers were not calculated correctly if the start of the week (`WKST`) was different from Monday.


Ical.Net/CalendarExtensions.cs
```diff
@@ -11,20 +11,32 @@ namespace Ical.Net;
...
+
+    /// <summary>
+    /// Calculate and return the date that represents the first day of the week the given date is
+    /// in, according to the week numbering required by RFC 5545.
+    /// </summary>
+    private static DateTime GetStartOfWeek(this DateTime t, DayOfWeek firstDayOfWeek)
+    {
+        var t0 = ((int) firstDayOfWeek) % 7;
+        var tn = ((int) t.DayOfWeek) % 7;
+        return t.AddDays(-((tn + 7 - t0) % 7));
```
axunonb, line 40: Very well explained

minichma: Thank you. Actually its a citation from RFC 5545 ;-)

axunonb (approved):
Very good, another step ahead, thanks!

PR #655, opened 2024-11-27
Remove mutable state from Evaluators and some related, poorly tested functions

minichma:
The complexity of occurrence evaluation is impacted by the internal state maintained by classes implementing `IEvaluator` and related. This PR aims to reduce that complexity by eliminating much of this state. Only a few functions rely on this state, and these are either poorly documented, untested, or better suited for user implementation. The benefits of removing this state appear to significantly outweigh the drawbacks.

Most notable the following public methods are removed:
* `ToDo.IsCompleted()`: poorly tested, i.e. only basic execution paths are tested, but not the more complex ones.
* `ToDo.IsActive()`: see `IsCompleted()`
* `CalendarEvent.Occurs*()`: untested and they seem to be better suited for user implementation.

As a benefit the following internal state can be eliminated:
* `IEvaluator.Periods`
* much of the other mutable state of `IEvaluator`, i.e. `.Evaluation*Bounds`
* `Evaluator._mEvaluation*Bounds`


axunonb:
We could set `nullable enable` for changed files. Even one file per PR would be fine.

minichma:
It would certainly be great to make some progress on the nullable front, but I'd also appreciate not to grow the PR mor than necessary and keep concerns separated.

minichma:
@axunonb What do you think of removing those methods I mentioned. It would certainly be possible to implement them also without the state that was removed, but I struggled with understanding the intended purpose, because partly untested and no sufficient docs. Do you agree that it should be ok to remove them in the first v5 version? Should people complain, we could re-evaluate and re-introduce.

axunonb:
> struggled with understanding the intended purpose, because partly untested and no sufficient docs

There is always a certain doubt when changing code without docs and missing unit tests. Sometime there is code that might aim to serve achieving a certain target, but with the full implementation left.
If cutting out doesn't break existing (other) unit tests let's be brave.

minichma:
Alright, its just the two tests in TodoTest that I removed, but they also only covered certain execution paths.

Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -13,37 +13,16 @@ namespace Ical.Net.Evaluation;
 
 public abstract class Evaluator : IEvaluator
 {
-    private DateTime _mEvaluationStartBounds = DateTime.MaxValue;
-    private DateTime _mEvaluationEndBounds = DateTime.MinValue;
```
axunonb, line 17: Just a sidenote: Initializing with `DateTime.MinValue` in ical.net is a potential risk because we can't create a `DateTimeOffset` with it.

axunonb (commented):
All fine, if you think *CodeCov* coverage warnings can be neglected.

axunonb:
> It would certainly be great to make some progress on the nullable front, but I'd also appreciate not to grow the PR mor than necessary and keep concerns separated.

Yep, let's try to pick one file after the other, instead of a big bang. If helpful, include in a PR, otherwise chose a separate PR.

Ical.Net/Evaluation/Evaluator.cs
minichma: Absolutely agree. So another benefit of them being gone. There are quite many of those magic values throughout the lib. Should implement `Nullable` in most of those places.

minichma:
> All fine, if you think CodeCov coverage warnings can be neglected.

Yes, I think so. The number of missed lines is reduced. The reduced number of hits originates from code that has been deleted, which basically is good for the code quality. We don't have new issues and I think that most new code is covered. The modified code that isn't covered hasn't been covered before as well and I don't feel in the position to write all those tests in a meaningful way at this time.

Most notably the `TimeZoneEvaluator` isn't covered, because it isn't used anywhere and not tested either. The reason seems to be that timezone evaluation is done via NodaTime, so maybe it would be best to remove it altogether, as nobody seems to be able to maintain it. And restore it, should demand arise? What do you think?

axunonb:
> timezone evaluation is done via NodaTime, so maybe it would be best to remove it altogether,

Good finding, and I'm sure nobody will even notice that it's gone. Yes, let's remove the class.

minichma:
Removed the `TimezoneEvaluator`. Was a little unfair regarding the tests of `Todo` though. There was quite some code related to testing `Todo.IsComplete` and `.IsActive`, although they don't cover all the code paths. Anyhow, I restored them and simplified the code so it doesn't rely on `IEvaluator.Period` state, which is to be removed, so everything fine with that now.

axunonb (approved):
Great, thanks

PR #657, opened 2024-11-30
Recurrence: Reproduce and fix issues if `BYWEEKNO` and `INTERVAL` are used together

minichma:
This PR reproduces and fixes more issues around `BYWEEKNO`, in particular if used together with `INTERVAL` and if `DTSTART` lies in a different year then the week it lies in. E.g. `20100102T000000` is in 2010, but the week number is 53 of 2009. In such cases the lib increments year invervals staring from 2010 rather than from 2009, which is wrong.


Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -55,11 +55,8 @@ protected IDateTime ConvertToIDateTime(DateTime dt, IDateTime referenceDate)
 
     protected void IncrementDate(ref DateTime dt, RecurrencePattern pattern, int interval)
     {
-        // FIXME: use a more specific exception.
```
axunonb, line 58: ... or even no exception...

axunonb (approved):
Really good, recurrence quality is increasing! Tnx

Ical.Net/Evaluation/Evaluator.cs
minichma: 😄 Yeah, why shouldn't we increment by 0? The method shouldn't make too many assumptions about the caller's intent.

PR #658, opened 2024-11-30
Refactor to use DateOnly and TimeOnly for date/time handling

### LINKED ISSUE #656: DTEND DATE value is not reckoned exclusively
A loaded event containing the property

> DTEND;VALUE=DATE:20241128

is returned by `GetOccurrences(new DateTime(2024, 11, 28))`. This is contrary to [the spec](https://icalendar.org/iCalendar-RFC-5545/3-6-1-event-component.html), which says

> The ["DTEND"](https://icalendar.org/iCalendar-RFC-5545/3-8-2-2-date-time-end.html) property for a ["VEVENT"](https://icalendar.org/iCalendar-RFC-5545/3-6-1-event-component.html) calendar component specifies the non-inclusive end of the event.

I understand this to mean that an event which ends _by_ 0000 hours on a date is not an occurrence _on_ that date.

### LINKED ISSUE #662: Clean-up determining the end time of a period
Currently there are several places in the lib there the end date is determined by `AddTick(-1`, `AddSeconds(-1` or alike:
https://github.com/ical-org/ical.net/blob/add58a9bc3dffb8014a59d0bc6a831dbfa5635f7/Ical.Net/Evaluation/RecurrenceUtil.cs#L16-L17

Reasoning:
1. Subtracting an arbitrary epsilon changes the user-supplied value
2. When it comes to a time of `0:00:00` the result will lead to wrong calculations

axunonb:
Increased `CalDateTime` immutability

CalDateTime:
* `Date` is of type `DateOnly` (breaking vs. v4)
* Make `UtzTzId` a public string
* Remove `DateOnlyValue`
* Remove `HasDate` (always true) (breaking vs. v4)
* Remove `Millisecond` (breaking vs. v4)
* Remove `Ticks` (breaking vs. v4)
* Remove overload CalDateTime(int year, int month, int day, string? tzId) (breaking vs. v4)
* `HasTime` only has a getter (breaking vs. v4)
* `Time` is a getter of type `TimeOnly`. Values are always rounded to the nearest second (breaking vs. v4)
* `AddMilliseconds` takes `double` as argument, like `DateTime.AddMilliseconds` (breaking vs. v4)
* Remove `TimeOnlyValue`
* `Value` only has a getter (breaking vs. v4)
* `TzId` only has a getter (breaking vs. v4)
* Remove `AsSystemLocal` (breaking vs. v4)
* Remove `AsDateTimeOffset` (breaking vs. v4)
* Remove `AddMilliseconds` (breaking vs. v4)
* Remove `AddTicks` (breaking vs. v4)
* Remove unused `TimeSpan? operator -(CalDateTime? left, IDateTime? right)` (breaking vs. v4)
* Update Equality operators for comparing to floating date/time (breaking vs. v4)

IDateTime (breaking vs. v4):
* Remove `AsSystemLocal` (breaking vs. v4)`
* Remove `AsDateTimeOffset` (breaking vs. v4)`
* Remove `AddMilliseconds` (breaking vs. v4)
* Remove `AddTicks` (breaking vs. v4)
* Remove `Millisecond` (breaking vs. v4)
* Remove `Ticks` (breaking vs. v4)
* Add explicit bool IsFloating (implicit TzId == null)
* `Date` is a getter of type `DateOnly`
* `Time` is a getter of type `TimeOnly`
* `HasTime` only has a getter
* Remove `HasDate` (always true) (breaking vs. v4)
* `Value` only has a getter
* `TzId` only has a getter
* `ToTimeZone(string?)` argument is a NRT

Miscellaneous:
* CTORs with `DateTime` arguments persist. Actually they are also comfortable, and now consistently processed
* Keep `IDateTime` - maybe in another PR
* Transitioned from DateTime to DateOnly and TimeOnly across the codebase for improved date and time handling.
* Updated methods, properties, and tests to ensure compatibility with the new approach.
* Updated xmldoc

Resolves #656
Resolves #662

### DIFF
--- a/Ical.Net.Tests/CalDateTimeTests.cs
+++ b/Ical.Net.Tests/CalDateTimeTests.cs
@@ -305,12 +305,9 @@ public class CalDateTimeTests
         Assert.Multiple(() =>
         {
             Assert.That(c2.Value, Is.EqualTo(c3.Value));
-            Assert.That(c2.Ticks, Is.EqualTo(c3.Ticks));
             Assert.That(c2.TzId, Is.EqualTo(c3.TzId));
             Assert.That(CalDateTime.UtcNow.Value.Kind, Is.EqualTo(DateTimeKind.Unspecified));
             Assert.That(CalDateTime.Today.Value.Kind, Is.EqualTo(DateTimeKind.Unspecified));
-            Assert.That(c.Millisecond, Is.EqualTo(0));
-            Assert.That(c.Ticks, Is.EqualTo(dt.Ticks));
             Assert.That(c.DayOfYear, Is.EqualTo(dt.DayOfYear));
             Assert.That(c.Time?.ToTimeSpan(), Is.EqualTo(dt.TimeOfDay));
             Assert.That(c.Subtract(TimeSpan.FromSeconds(dt.Second)).Value.Second, Is.EqualTo(0));
--- a/Ical.Net.Tests/CalDateTimeTests.cs
+++ b/Ical.Net.Tests/CalDateTimeTests.cs
@@ -73,45 +73,9 @@ public class CalDateTimeTests
             .SetName($"IANA to BCL: {ianaNy} to {bclCst}");
     }
 
-    [TestCaseSource(nameof(AsDateTimeOffsetTestCases))]
-    public DateTimeOffset AsDateTimeOffsetTests(CalDateTime incoming)
-        => incoming.AsDateTimeOffset;
-
-    public static IEnumerable AsDateTimeOffsetTestCases()
-    {
-        const string nyTzId = "America/New_York";
-        var summerDate = DateTime.Parse("2018-05-15T11:00");
-
-        var nySummerOffset = TimeSpan.FromHours(-4);
-        var nySummer = new CalDateTime(summerDate, nyTzId);
-        yield return new TestCaseData(nySummer)
-            .SetName("NY Summer DateTime returns DateTimeOffset with UTC-4")
-            .Returns(new DateTimeOffset(summerDate, nySummerOffset));
-
-        var utc = new CalDateTime(summerDate, "UTC");
-        yield return new TestCaseData(utc)
-            .SetName("UTC summer DateTime returns a DateTimeOffset with UTC-0")
-            .Returns(new DateTimeOffset(summerDate, TimeSpan.Zero));
-
-        var convertedToNySummer = new CalDateTime(summerDate, nyTzId);
-        yield return new TestCaseData(convertedToNySummer)
-            .SetName(
-                "Summer UTC DateTime converted to NY time zone by setting TzId returns a DateTimeOffset with UTC-4")
-            .Returns(new DateTimeOffset(summerDate, nySummerOffset));
-
-        var noTz = new CalDateTime(summerDate);
-        var currentSystemOffset = TimeZoneInfo.Local.GetUtcOffset(summerDate);
-        yield return new TestCaseData(noTz)
-            .SetName(
-                $"Summer DateTime with no time zone information returns the system-local's UTC offset ({currentSystemOffset})")
-            .Returns(new DateTimeOffset(summerDate, currentSystemOffset));
-    }
-
     [Test(Description = "Calling AsUtc should always return the proper UTC time, even if the TzId has changed")]
     public void TestTzidChanges()
     {
-        var x = new CalDateTime(2018, 5, 21).AsDateTimeOffset;
-
         var someTime = DateTimeOffset.Parse("2018-05-21T11:35:00-04:00");
 
         var someDt = new CalDateTime(someTime.DateTime, "America/New_York");
@@ -356,32 +320,6 @@ public class CalDateTimeTests
             Assert.That(c.ToString("dd.MM.yyyy"), Is.EqualTo("02.01.2025 Europe/Berlin"));
             // Create a date-only CalDateTime from a CalDateTime
             Assert.That(new CalDateTime(new CalDateTime(2025, 1, 1)), Is.EqualTo(new CalDateTime(2025, 1, 1)));
-        });
-    }
-
-    [Test]
-    public void Simple_PropertyAndMethod_NotHasTime_Tests()
-    {
-        var dt = new DateTime(2025, 1, 2, 0, 0, 0, DateTimeKind.Utc);
-        var c = new CalDateTime(dt, tzId: "Europe/Berlin", hasTime: false);
-
-        // Adding time to a date-only value should not change the HasTime property
-        Assert.Multiple(() =>
-        {
-            var result = c.AddHours(1);
-            Assert.That(result.HasTime, Is.True);
-
-            result = c.AddMinutes(1);
-            Assert.That(result.HasTime, Is.True);
-
-            result = c.AddSeconds(1);
-            Assert.That(result.HasTime, Is.True);
-
-            result = c.AddMilliseconds(1000);
-            Assert.That(result.HasTime, Is.True);
-
-            result = c.AddTicks(TimeSpan.FromMinutes(1).Ticks);
-            Assert.That(result.HasTime, Is.True);
         });
     }
 }
--- a/Ical.Net.Tests/CalDateTimeTests.cs
+++ b/Ical.Net.Tests/CalDateTimeTests.cs
@@ -93,8 +93,7 @@ public class CalDateTimeTests
             .SetName("UTC summer DateTime returns a DateTimeOffset with UTC-0")
             .Returns(new DateTimeOffset(summerDate, TimeSpan.Zero));
 
-        var convertedToNySummer = new CalDateTime(summerDate, "UTC");
-        convertedToNySummer.TzId = nyTzId;
+        var convertedToNySummer = new CalDateTime(summerDate, nyTzId);
         yield return new TestCaseData(convertedToNySummer)
             .SetName(
                 "Summer UTC DateTime converted to NY time zone by setting TzId returns a DateTimeOffset with UTC-4")
@@ -111,13 +110,15 @@ public class CalDateTimeTests
     [Test(Description = "Calling AsUtc should always return the proper UTC time, even if the TzId has changed")]
     public void TestTzidChanges()
     {
+        var x = new CalDateTime(2018, 5, 21).AsDateTimeOffset;
+
         var someTime = DateTimeOffset.Parse("2018-05-21T11:35:00-04:00");
 
-        var someDt = new CalDateTime(someTime.DateTime) { TzId = "America/New_York" };
+        var someDt = new CalDateTime(someTime.DateTime, "America/New_York");
         var firstUtc = someDt.AsUtc;
         Assert.That(firstUtc, Is.EqualTo(someTime.UtcDateTime));
 
-        someDt.TzId = "Europe/Berlin";
+        someDt = new CalDateTime(someTime.DateTime, "Europe/Berlin");
         var berlinUtc = someDt.AsUtc;
         Assert.That(berlinUtc, Is.Not.EqualTo(firstUtc));
     }
@@ -132,12 +133,12 @@ public class CalDateTimeTests
         var localDt = DateTime.SpecifyKind(DateTime.Parse("2018-05-21T11:35:33"), DateTimeKind.Unspecified);
 
         yield return new TestCaseData(localDt, "UTC")
-            .Returns(DateTimeKind.Utc)
-            .SetName("Explicit tzid = UTC time zone returns DateTimeKind.Utc");
+            .Returns(DateTimeKind.Unspecified)
+            .SetName("Explicit tzid = UTC time zone returns DateTimeKind.Unspecified");
 
         yield return new TestCaseData(DateTime.SpecifyKind(localDt, DateTimeKind.Utc), null)
-            .Returns(DateTimeKind.Utc)
-            .SetName("DateTime with Kind = Utc and no tzid returns DateTimeKind.Utc");
+            .Returns(DateTimeKind.Unspecified)
+            .SetName("DateTime with Kind = Utc and no tzid returns DateTimeKind.Unspecified");
 
         yield return new TestCaseData(localDt, localTz)
             .Returns(DateTimeKind.Unspecified)
@@ -170,8 +171,8 @@ public class CalDateTimeTests
             .Returns("2024-08-30T10:30:00.0000000+12:00 Pacific/Auckland")
             .SetName("Date and time with 'O' format arg, default culture");
 
-        yield return new TestCaseData(new CalDateTime(2024, 8, 30, tzId: "Pacific/Auckland"), "O", null)
-            .Returns("08/30/2024 Pacific/Auckland")
+        yield return new TestCaseData(new CalDateTime(2024, 8, 30), "O", null)
+            .Returns("2024-08-30") // Date only cannot have timezone
             .SetName("Date only with 'O' format arg, default culture");
 
         yield return new TestCaseData(new CalDateTime(2024, 8, 30, 10, 30, 0, tzId: "Pacific/Auckland"), "O",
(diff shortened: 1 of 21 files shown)


Ical.Net/DataTypes/Trigger.cs
```diff
@@ -37,7 +37,11 @@ public virtual IDateTime DateTime
             Duration = null;
 
             // Ensure date/time has a time part
-            _mDateTime.HasTime = true;
+            if (!_mDateTime.HasTime)
+            {
+                _mDateTime = new CalDateTime(_mDateTime.Date, new TimeOnly(), _mDateTime.TzId)
```
minichma, line 42: Maybe some extension method `CalDateTime.WithTime(TimeOnly t = default)` would be helpful, then wr could write something like `_mDateTime = _mDateTime.WithTime()`.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -154,29 +174,36 @@ public CalDateTime(DateOnly date, TimeOnly? time, string? tzId = null)
...
+        _tzId = HasTime ? tzId : null;
     }
 
     private void Initialize(DateOnly dateOnly, TimeOnly? timeOnly, string? tzId)
     {
-        HasTime = timeOnly.HasValue;
-        HasDate = true;
         _dateOnly = dateOnly;
-        _timeOnly = timeOnly ?? new TimeOnly();
+        _timeOnly = RoundTimeToNearestSecond(timeOnly);
```
minichma: Hm, this could easily have side effects, e.g. when used as the recurrence start date. E.g. if periodStart is 400 ms after DTSTART, then according to today's logic DTSTART would not be returned, according to the new logic it would, but not if it is 500 ms after it. I think I would rather keep the original value (you know, I'm always a fan of keeping the user's input as far as possible). *If* we want to restrict it to seconds, then I'd rather suggest to truncate the sub-second parts, not to round, because it is easier to understand and deal with.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -228,35 +253,54 @@ public override int GetHashCode()
...
+        {
+            return true;
+        }
+
+        if (left is null || right is null)
+        {
+            return false;
+        }
+
+        if (left.IsFloating || right.IsFloating)
```
minichma: Suggest to add a check if `IsFloating` equals. I.e. 

```
if (left.IsFloating != right.IsFloating)
    return false; 
```

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -266,35 +310,45 @@ public override int GetHashCode()
...
     /// <remarks>
-    /// This will also set <seealso cref="HasTime"/> to <see langword="true"/>,
+    /// This will also add a <see cref="IDateTime.Time"/> part that did not exist before the operation,
     /// if the <see cref="TimeSpan"/> is not a multiple of 24 hours.
     /// </remarks>
     public static IDateTime operator -(CalDateTime left, TimeSpan right)
     {
         var copy = left.Copy<CalDateTime>();
+        var newValue = copy.Value - right;
         if ((right.Ticks % TimeSpan.TicksPerDay) != 0)
```
minichma: Caution: If left and right both have the same time value set, then you'd keep the original time even if <> 0:00.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -266,35 +310,45 @@ public override int GetHashCode()
...
         {
-            copy.HasTime = true;
+            copy._dateOnly = DateOnly.FromDateTime(newValue);
+            copy._timeOnly = RoundTimeToNearestSecond(newValue);
```
minichma: Maybe the case should be disallowed altogether that time is added/subtracted to/from a date.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -486,126 +509,139 @@ TzId is null
...
+        return copy;
     }
 
     /// <inheritdoc cref="DateTime.AddSeconds"/>
     /// <remarks>
-    /// This will also set <seealso cref="HasTime"/> to <see langword="true"/>,
+    /// This will also add a <see cref="IDateTime.Time"/> part that did not exist before the operation,
     /// if the seconds are not a multiple of 86400.
     /// </remarks>
     public IDateTime AddSeconds(int seconds)
```
minichma, line 559: I think we should raise exceptions if `Add[sec|min|hour]` is called. If people want to convert a DATE to a DATE-TIME, then they should do this explicitly and also specify a tzid (or intentionally none).

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -486,126 +509,139 @@ TzId is null
...
     /// <inheritdoc cref="DateTime.AddMilliseconds"/>
     /// <remarks>
-    /// This will also set <seealso cref="HasTime"/> to <see langword="true"/>
+    /// This will also add a <see cref="IDateTime.Time"/> part that did not exist before the operation,
     /// if the milliseconds are not a multiple of 86400000.
     /// <para/>
     /// Milliseconds less than full seconds get truncated.
     /// </remarks>
-    public IDateTime AddMilliseconds(int milliseconds)
+    public IDateTime AddMilliseconds(double milliseconds)
```
minichma: If we generally round to seconds, then we shouldn't allow adding milliseconds. But as mentioned above, I think we should consider allowing them. But in any case we should be consistent.

Ical.Net/Evaluation/RecurrenceUtil.cs
```diff
@@ -13,8 +14,12 @@ namespace Ical.Net.Evaluation;
 
 internal class RecurrenceUtil
 {
-    public static HashSet<Occurrence> GetOccurrences(IRecurrable recurrable, IDateTime dt, bool includeReferenceDateInResults) => GetOccurrences(recurrable,
-        new CalDateTime(dt.Date), new CalDateTime(dt.Date.AddDays(1).AddSeconds(-1)), includeReferenceDateInResults);
+    public static HashSet<Occurrence> GetOccurrences(IRecurrable recurrable, IDateTime dt, bool includeReferenceDateInResults)
+    {
+        var endTimeOnly = dt.Time?.Add(TimeSpan.FromSeconds(-1));
```
minichma: Be careful, Time could be 0:00 in which case we'd end up with a negative time.

Ical.Net.Tests/CalDateTimeTests.cs
```diff
@@ -73,51 +74,16 @@ public static IEnumerable ToTimeZoneTestCases()
             .SetName($"IANA to BCL: {ianaNy} to {bclCst}");
     }
 
-    [TestCaseSource(nameof(AsDateTimeOffsetTestCases))]
-    public DateTimeOffset AsDateTimeOffsetTests(CalDateTime incoming)
```
axunonb, line 77: `DateTimeOffset`has been removed

Ical.Net.Tests/CalDateTimeTests.cs
```diff
@@ -189,58 +154,150 @@ public static IEnumerable ToStringTestCases()
...
-            .Returns("30/08/2024 Pacific/Auckland")
+        yield return new TestCaseData(new CalDateTime(2024, 8, 30), null,
+                CultureInfo.GetCultureInfo("IT")) // Date only cannot have timezone
+            .Returns("30/08/2024") 
             .SetName("Date only with 'IT' CultureInfo and default format arg");
     }
 
-    [Test]
-    public void SetValue_AppliesSameRulesAsWith_CTOR()
+    [Test, TestCaseSource(nameof(DateTimeArithmeticTestCases))]
```
axunonb, line 164: Full coverage for all arithmetic operations

Ical.Net.Tests/CalDateTimeTests.cs
```diff
@@ -189,58 +154,150 @@ public static IEnumerable ToStringTestCases()
...
+
+        yield return new TestCaseData(new Func<IDateTime, IDateTime>(dt => dt.AddTicks(TimeSpan.FromHours(1).Ticks)))
+            .Returns(dateTime.AddTicks(TimeSpan.FromHours(1).Ticks))
+            .SetName($"{nameof(IDateTime.AddTicks)} ticks of 1 hour");
     }
 
-    [Test]
-    public void SetValue_LeavesExistingPropertiesUnchanged()
+    [Test, TestCaseSource(nameof(EqualityTestCases))]
+    public bool EqualityTests(Func<IDateTime, bool> operation)
```
axunonb, line 193: Full coverage for all equality comparisons

Ical.Net.Tests/ComponentTest.cs
```diff
@@ -22,29 +22,4 @@ public void UniqueComponent1()
         Assert.That(evt.Created, Is.Null); // We don't want this to be set automatically
         Assert.That(evt.DtStamp, Is.Not.Null);
     }
-
-    [Test, Category("Components")]
```
axunonb, line 26: Value has no setter now

Ical.Net.Tests/ProgramTest.cs
```diff
@@ -52,8 +52,8 @@ public void Merge1()
 
         // Get occurrences for the first event
         var occurrences = evt1.GetOccurrences(
-            new CalDateTime(1996, 1, 1, _tzid),
```
axunonb, line 55: Resharper auto-refactoring for changed signature

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -3786,5 +3786,28 @@ private static DateTime SimpleDateTimeToMatch(IDateTime dt, IDateTime toMatch)
         }
         return dt.Value;
     }
-}
 
+    [Test]
+    public void GetOccurrenceShouldExcludeDtEnd()
```
axunonb: Added to ensure that #656 is resolved

Ical.Net/DataTypes/IDateTime.cs
```diff
@@ -10,129 +10,135 @@ namespace Ical.Net.DataTypes;
...
     TimeSpan Subtract(IDateTime dt);
-
     IDateTime AddYears(int years);
     IDateTime AddMonths(int months);
     IDateTime AddDays(int days);
     IDateTime AddHours(int hours);
     IDateTime AddMinutes(int minutes);
     IDateTime AddSeconds(int seconds);
-    IDateTime AddMilliseconds(int milliseconds);
+    IDateTime AddMilliseconds(double milliseconds);
```
axunonb: same as `DateTilme.AddMilliseconds` - `int` is not useful

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -988,6 +990,6 @@ private static IDateTime MatchTimeZone(IDateTime dt1, IDateTime dt2)
 
         return dt1.IsUtc
             ? new CalDateTime(copy.AsUtc)
-            : new CalDateTime(copy.AsSystemLocal);
```
axunonb, line 991: Isn't the code in `MatchTimeZone(IDateTime dt1, IDateTime dt2)` completely useless?

Ical.Net/Calendar.cs
```diff
@@ -238,11 +244,17 @@ public virtual HashSet<Occurrence> GetOccurrences(DateTime startTime, DateTime e
     /// <param name="dt">The date for which to return occurrences. Time is ignored on this parameter.</param>
     /// <returns>A list of Periods representing the occurrences of this object.</returns>
     public virtual HashSet<Occurrence> GetOccurrences<T>(IDateTime dt) where T : IRecurringComponent
-        => GetOccurrences<T>(new CalDateTime(dt.Date), new CalDateTime(dt.Date.AddDays(1).AddTicks(-1)));
```
axunonb, line 241: This is probably meant to be `TimeSpan.FromSeconds(-1)` because 1 tick just gets truncated?

Ical.Net/Calendar.cs
```diff
@@ -238,11 +244,17 @@ public virtual HashSet<Occurrence> GetOccurrences(DateTime startTime, DateTime e
...
+    {
+        var endTimeOnly = dt.Time?.Add(TimeSpan.FromTicks(-1));
+        return GetOccurrences<T>(new CalDateTime(dt.Date, dt.Time), new CalDateTime(dt.Date.AddDays(1), endTimeOnly));
+    }
 
     /// <inheritdoc cref="GetOccurrences(IDateTime)"/>
     public virtual HashSet<Occurrence> GetOccurrences<T>(DateTime dt) where T : IRecurringComponent
-        => GetOccurrences<T>(new CalDateTime(dt.Date), new CalDateTime(dt.Date.AddDays(1).AddTicks(-1)));
```
axunonb, line 245: This is probably meant to be `TimeSpan.FromSeconds(-1)` because 1 tick just gets truncated?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
minichma: Its not the most elegant code, but it aims to adjust UNTIL, which usually is specified in UTC, to DTSTART's time zone, which usually isn't in UTC. This allows checking comparing the adjusted UNTIL with the generated recurrences without considering the time zone again. So the functionality is needed one way or the other.

Ical.Net/Calendar.cs
minichma: Now that CalDateTime is rounded to seconds, ticks aren't helpful anymore. Question would be generally, whether subtracting an 'Epsilon' is the best idea. It would certainly be cleaner to get the comparisons later on right, i.e. to compare for endTime to be exclusive, rather than inclusive.

minichma:
Just noted, that I didn't submit my review comments from 2d ago, so please read my comments in the context of last Sat.

minichma:
This is a great improvement regarding readability and maintainability!

Ical.Net/DataTypes/CalDateTime.cs
axunonb: On the first glance I thought the proposal worked, but it doesn't.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: That's what I thought, but the way it is implemented looks like a noop. Then I removed the method invocation temporarily with no effect on unit tests. So this is just a hint re recurrences.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: > keep the original value

Before this centralized rounding there was a mixture, see e.g. `DateUtil.Truncate` and other adjustments that did not use this method. I think the general limitation to seconds brings advantages in maintenance and reproducibility. I agree truncating is more consequent. Will modify.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
minichma: I could easily be wrong, or relevant tests are missing. We'd need a test with UNTIL exactly at the last recurrence and then run the test with a TZ with positive offset and with one with a negative offset. In any case the last recurrence should be returned.

axunonb: The 2 arguments to `MatchTimeZone(IDateTime dt1, IDateTime dt2)` always have the same timezone, so there is nothing to adjust? [Edit] The method in effect is `r.Until = r.Until`.
```C#
if (r.Until != DateTime.MinValue)
{
    r.Until = MatchTimeZone(referenceDate, new CalDateTime(r.Until, referenceDate.TzId)).Value;
}
```

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Happy to hear your suggestion, as this was my first conclusion, too. It's one of "guess the users intention".
~~Let's throw an `InvalidOperationException`. Will modify._~~
[Edit] Brings  > 100 failing tests, not for this PR

Ical.Net/DataTypes/CalDateTime.cs
axunonb: InvalidOperationException, see below.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: InvalidOperationException - see below

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Yep, will remove milliseconds and ticks

Ical.Net/Evaluation/RecurrenceUtil.cs
axunonb: InvalidOperationException: see below

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Will correct

Ical.Net/Calendar.cs
```diff
@@ -205,11 +205,17 @@ public VTimeZone AddTimeZone(VTimeZone tz)
     /// <param name="dt">The date for which to return occurrences. Time is ignored on this parameter.</param>
     /// <returns>A list of occurrences that occur on the given date (<paramref name="dt"/>).</returns>
     public virtual HashSet<Occurrence> GetOccurrences(IDateTime dt)
-        => GetOccurrences<IRecurringComponent>(new CalDateTime(dt.Date), new CalDateTime(dt.Date.AddDays(1).AddSeconds(-1)));
+    {
+        var endTimeOnly = dt.Time?.Add(TimeSpan.FromSeconds(-1));
```
minichma: If Time is 0:00, you'll end up with a negative time.

Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -152,15 +152,19 @@ public virtual bool IsAllDay
         get => !Start.HasTime;
         set
         {
-            // Set whether or not the start date/time
+            // Set whether the start date/time
```
minichma: I don't think this property should be editable. Whether the event is all-day should be specified via DTSTART.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -154,29 +174,37 @@ public CalDateTime(DateOnly date, TimeOnly? time, string? tzId = null)
...
-        HasTime = timeOnly.HasValue;
-        HasDate = true;
         _dateOnly = dateOnly;
-        _timeOnly = timeOnly ?? new TimeOnly();
+        _timeOnly = TruncateTimeToSeconds(timeOnly);
 
-        _tzId = string.Equals(UtcTzId, tzId, StringComparison.OrdinalIgnoreCase) ? UtcTzId : tzId;
+        _tzId = tzId switch
+        {
+            _ when !timeOnly.HasValue && string.Equals(UtcTzId, tzId, StringComparison.OrdinalIgnoreCase) => UtcTzId,
```
minichma: Why does UTC receive special handling here?

Ical.Net/DataTypes/CalDateTime.cs
minichma: If we don't add this check, we could consider instances as equal even if one is floating and the other has a TZ set, which seems wrong. What would be the reason for that behavior?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -266,35 +311,45 @@ public override int GetHashCode()
...
-        if ((right.Ticks % TimeSpan.TicksPerDay) != 0)
+        var newValue = copy.Value - right;
+        if (!copy.HasTime && (right.Ticks % TimeSpan.TicksPerDay) == 0)
         {
-            copy.HasTime = true;
+            copy._dateOnly = DateOnly.FromDateTime(newValue);
+        }
+        else
+        {
+            copy._dateOnly = DateOnly.FromDateTime(newValue);
```
minichma, line 325: I think we shouldn't silently convert a date-only value to date-time but rather throw an exception. This is tricky in multiple ways. In particular, if DST changes. Maybe related to #660. Do we need those operators at all? If yes, we might want to be more explicit, i.e. have a 'add/subtract nominal' vs 'add/subtract accurate' in correspondence with how RFC5545 distinguishes those cases.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -266,35 +311,45 @@ public override int GetHashCode()
...
-        if ((right.Ticks % TimeSpan.TicksPerDay) != 0)
+        var newValue = copy.Value + right;
+        if (!copy.HasTime && (right.Ticks % TimeSpan.TicksPerDay) == 0)
+        {
+            copy._dateOnly = DateOnly.FromDateTime(newValue);
+        }
+        else
         {
-            copy.HasTime = true;
+            copy._dateOnly = DateOnly.FromDateTime(newValue);
```
minichma, line 348: see above

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -440,41 +435,53 @@ public string? TzId
...
-        var converted = zonedOriginal.WithZone(DateUtil.GetZone(tzId));
-
-        return converted.Zone == DateTimeZone.Utc
-            ? new CalDateTime(converted.ToDateTimeUtc(), tzId)
-            : new CalDateTime(DateTime.SpecifyKind(converted.ToDateTimeUnspecified(), DateTimeKind.Unspecified), tzId);
+        if (time is null)
+        {
+            return null;
+        }
+        var roundedSecond = time.Value.Second + (time.Value.Millisecond >= 500 ? 1 : 0);
```
minichma: This code still seems to round to the next second rather than truncate?

Ical.Net/DataTypes/IDateTime.cs
```diff
@@ -10,129 +10,133 @@ namespace Ical.Net.DataTypes;
...
-    /// Gets the second for this date/time value.
+    /// Gets the second that applies to the <see cref="Value"/>.
     /// </summary>
     int Second { get; }
 
     /// <summary>
-    /// Gets the millisecond for this date/time value.
+    /// Gets the millisecond that applies to the <see cref="Value"/>.
     /// </summary>
     int Millisecond { get; }
```
minichma: Remove?

Ical.Net/DataTypes/IDateTime.cs
```diff
@@ -10,129 +10,133 @@ namespace Ical.Net.DataTypes;
...
 
     /// <summary>
-    /// Gets the ticks for this date/time value.
+    /// Gets the ticks that applies to the <see cref="Value"/>.
     /// </summary>
     long Ticks { get; }
```
minichma: remove?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
minichma: You are fully right, its a noop. But I strongly believe that something's wrong there, because the conversion needs to be done somewhere. Should be investigated separately. Maybe the translation between UTC and tzid happened in the constructor of CalDateTime and was lost in one of the previous PRs?

Ical.Net/Evaluation/RecurrenceUtil.cs
minichma: Not sure, I understand. I think the problem still exists. If someone specifies a dt e.g. 20241203T000000, which is legit, then this code would lead to 20241203 T -000001.

Ical.Net/Utility/DateUtil.cs
```diff
@@ -19,7 +19,7 @@ public static IDateTime StartOfDay(IDateTime dt)
         => dt.AddHours(-dt.Hour).AddMinutes(-dt.Minute).AddSeconds(-dt.Second);
 
     public static IDateTime EndOfDay(IDateTime dt)
```
minichma, line 21: We should probably get rid of that. Calculations should be done precisely, adding -1 whatever shouldn't be necessary.

minichma:
Sorry for the many comments. You're working on the big ones here, so many things to consider.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: thanks, that's redundant

Ical.Net/Utility/DateUtil.cs
axunonb: Created issue #662 for a separate PR

Ical.Net/CalendarComponents/CalendarEvent.cs
axunonb: I just changed the wording of the comment :)

But yes, this one was easy. I removed the setter that was used only in 1 unit test. `AllDay` can simply be set with a `CalDateTime` without time.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Created  #661 for tracking a separate PR

Ical.Net/DataTypes/IDateTime.cs
axunonb: The question is applicable to other properties, too: Day, Month, Year...
All can be retrieved from the `Value` property. We could clean-up, when this would bring any advantage. But not in this PR.

axunonb:
No need for excuses :) - comments are all very welcome. 
There is one more commit with 2 issues from your latest comments.
For other comments we have new issues to track them - so they'll not go into this PR, which is huge already.
One important topic is the [discussion](https://github.com/ical-org/ical.net/discussions/606#discussioncomment-11451260) re comparing floating to zoned date/time, Let's resolve the topic there.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: Created #663 to track

Ical.Net/Calendar.cs
axunonb: Track in #662

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Ah, this was not recovered after by Visual Studio crash - again: thanks

Ical.Net/DataTypes/IDateTime.cs
axunonb: see below

Ical.Net/DataTypes/IDateTime.cs
minichma: Sure, my thought on Milliseconds and Ticks is just, that they are truncated starting with this PR, so providing getters for them could be confusing and considered inconsistent.

axunonb:
@minichma PR comments should all all be resolved or addressed

minichma:
I'm on a business trip until Sat. Not sure, whether I'll find the time until then.

Ical.Net/Calendar.cs
minichma: This is a new issue introduced in this PR, so shouldn't we fix it right away? The behavior is also somewhat different from the previous one. In the previous version we just truncated the time, in the new one we keep the time, if set, and observe the period until the same time next day. Is this intentional?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -428,53 +421,57 @@ public string? TzId
...
-        TzId is null
-            ? new DateTimeOffset(Value)
-            : DateUtil.ToZonedDateTimeLeniently(Value, TzId).ToDateTimeOffset();
+    private static TimeOnly? TruncateTimeToSeconds(DateTime dateTime)
+    {
+        return new TimeOnly(dateTime.Hour, dateTime.Minute, dateTime.Second);
+    }
+
+    /// <inheritdoc/>
+    public IDateTime ToTimeZone(string otherTzId)
```
minichma, line 460: Should we handle `otherTzId is null`?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -486,126 +483,93 @@ TzId is null
...
+    /// This will also add a <see cref="IDateTime.Time"/> part that did not exist before the operation,
     /// if the hours are not a multiple of 24.
     /// </remarks>
     public IDateTime AddHours(int hours)
     {
-        var dt = Copy<CalDateTime>();
-        if (!dt.HasTime && hours % 24 > 0)
+        var copy = Copy<CalDateTime>();
+        var newValue = copy.Value.AddHours(hours);
+        if (!copy.HasTime && (hours % 24 == 0))
```
minichma, line 521: We should consider (in a subsequent PR) to remove the special handling of multiples of 24. This is the kind of magic we had in v4 with midnight. A user wouldn't understand, why adding multiple of 24h behaves differently than adding other values. I suggest to either disallow adding hours to date-only instances altogether (preferred) , or, to always convert to date-time.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -954,8 +954,10 @@ public override HashSet<Period> Evaluate(IDateTime referenceDate, DateTime perio
             // This case is not defined by RFC 5545. We handle it by evaluating the rule
             // as if referenceDate had a time (i.e. set to midnight).
 
-            referenceDate = referenceDate.Copy<IDateTime>();
-            referenceDate.HasTime = true;
+            if (referenceDate.HasTime)
```
minichma: This is a redundant check, because already checked in the previous line.

Ical.Net/Evaluation/RecurrenceUtil.cs
minichma: I think, we would introduce a new bug here and should consider fixing it right away.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Yeah, I considered `null`, and finally decided for `otherTzId` `string` instead of `string?`, because `null` had the meaning "to floating". Well...

minichma: Oh, didn't notice it's not nullable. I think it's good as is. Could add a method ToFloating at some point.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: correct

Ical.Net/DataTypes/CalDateTime.cs
axunonb: issues #661 and #662 already exists since last review

axunonb:
>  introduce a new bug here and should consider fixing it right away

~~Agree about this should be fixed (#662) with another PR.~~
The fix was trivial, so it's done in https://github.com/ical-org/ical.net/pull/658/commits/5956fb998d61514362f39407a3a5ac50bb13eb88

It's existing code logic though:
https://github.com/ical-org/ical.net/blob/add58a9bc3dffb8014a59d0bc6a831dbfa5635f7/Ical.Net/Evaluation/RecurrenceUtil.cs#L16-L17

[Edit]
The remaining time adjustments were only found in class `Calendar`.
The adjustments are redundant, because
`public static HashSet<Occurrence> RecurrenceUtil.GetOccurrences(IRecurrable recurrable, IDateTime periodStart, IDateTime periodEnd, bool includeReferenceDateInResults)`
uses **precise** "LessThan" for comparing end date/times
This is now also cleaned-up with https://github.com/ical-org/ical.net/pull/658/commits/5956fb998d61514362f39407a3a5ac50bb13eb88

axunonb:
Comments resolved or tracked by issues

axunonb:
The remaining time adjustments were only found in class `Calendar`.

The adjustments are redundant, because calls finally go to
`public static HashSet<Occurrence> RecurrenceUtil.GetOccurrences(IRecurrable recurrable, IDateTime periodStart, IDateTime periodEnd, bool includeReferenceDateInResults)`
which uses **precise** "LessThan" for comparing end date/times.

minichma:
> The fix was trivial, so it's done in https://github.com/ical-org/ical.net/commit/5956fb998d61514362f39407a3a5ac50bb13eb88

> It's existing code logic though:

Great, thank you! The difference to the previous code is, that the previous version calculated 23:59:59 the previous day, but the new code calculated -0:00:01 the same day. So now its certainly better.

minichma (approved):
Great improvement!

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -488,13 +488,6 @@ public IDateTime ToTimeZone(string otherTzId)
             : new CalDateTime(converted.ToDateTimeUnspecified(), otherTzId);
     }
 
-    /// <inheritdoc />
-    /// <remarks>
-    /// If the <see cref="TzId"/> is <see langword="null"/> or does not represent an
-    /// IANA or other well-known timezone ID, the system's local timezone will be used.
-    /// </remarks>
-    public DateTimeOffset AsDateTimeOffset => DateUtil.ToZonedDateTimeLeniently(Value, TzId).ToDateTimeOffset();
```
CAPCHIK: @axunonb Hi!  

Could you explain why `AsDateTimeOffset` was removed? It was useful in v4, and `DateTimeOffset` is still used in `ToString`.  

How can I get a `DateTimeOffset` from `CalDateTime` now? `DateUtil` is internal, and in v5, it seems I have to use the same logic to extract the timezone offset from NodaTime.

I posted this as a comment because I was hoping to find the reason for the change in the commit history. Maybe it would be better to raise this as a separate issue?

axunonb: @CAPCHIK The reasons for the redesign are given at the top of this PR. This was the decision at this time.
We're currently not yet happy about `CalDateTime` as it is now,. All timezone related methods should go into a separate class, maybe as extension methods.
Any changes to `CalDateTime` should happen after the final design is clear. We welcome every helping hand with ical.net.
The new ical.net v5 release reflects the current WIP for `CalDatetime`.

> I have to use the same logic to extract the timezone offset from NodaTime.

Yes, for v5.0.0

CAPCHIK: Thanks for the explanation, I will try to find a way to help in the development.

Bluejanis:
Is there a migration documentation for this?
I updated the nuget package and now AsDateTimeOffset is gone for example.

axunonb:
> Is there a migration documentation for this?

Probably not as complete as you wished, but have a look at the [Migration Guide](https://github.com/ical-org/ical.net/wiki/Migrating-Guides) and the [API Changes](https://github.com/ical-org/ical.net/wiki/API-Changes-v4-to-v5)

PR #659, opened 2024-12-01
Reproduce and fix issues related to `RECURRENCE-ID`

minichma:
This PR simplifies the handling of `RECURRENCE-ID` and by doing so fixes two bugs:
* The RECURRENCE-ID considered only the date of occurrences to be replaced, not the time, which is wrong.
* The set of dates/times that have a `RECURRENCE-ID` and therefore replace others was constructed from the result set inside `GetOccurrences`. So if they were lying outside the observed range, they were missed, which could cause incorrect result sets.

### DIFF
--- a/Ical.Net.Tests/GetOccurrenceTests.cs
+++ b/Ical.Net.Tests/GetOccurrenceTests.cs
@@ -152,65 +152,141 @@ END:VCALENDAR";
     }
 
     [Test]
-    public void GetOccurrencesShouldEnumerate()
+    public void GetOccurrencesWithRecurrenceIdShouldEnumerate()
     {
-        const string ical =
-            @"BEGIN:VCALENDAR
-PRODID:-//github.com/rianjs/ical.net//NONSGML ical.net 2.2//EN
-VERSION:2.0
-BEGIN:VTIMEZONE
-TZID:W. Europe Standard Time
-BEGIN:STANDARD
-DTSTART:16010101T030000
-RRULE:FREQ=YEARLY;BYDAY=SU;BYMONTH=10;BYSETPOS=-1
-TZNAME:Mitteleuropäische Zeit
-TZOFFSETFROM:+0200
-TZOFFSETTO:+0100
-END:STANDARD
-BEGIN:DAYLIGHT
-DTSTART:00010101T020000
-RRULE:FREQ=YEARLY;BYDAY=SU;BYMONTH=3;BYSETPOS=-1
-TZNAME:Mitteleuropäische Sommerzeit
-TZOFFSETFROM:+0100
-TZOFFSETTO:+0200
-END:DAYLIGHT
-END:VTIMEZONE
-BEGIN:VEVENT
-BACKGROUND:BUSY
-DESCRIPTION:Backup Daten
-DTEND;TZID=W. Europe Standard Time:20150305T043000
-DTSTAMP:20161122T120652Z
-DTSTART;TZID=W. Europe Standard Time:20150305T000100
-RESOURCES:server
-RRULE:FREQ=WEEKLY;BYDAY=MO
-SUMMARY:Server
-UID:a30ed847-8000-4c53-9e58-99c8f9cf7c4b
-X-LIGHTSOUT-ACTION:START=WakeUp\;END=Reboot\,Force
-X-LIGHTSOUT-MODE:TimeSpan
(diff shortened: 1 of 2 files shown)


Ical.Net.Tests/GetOccurrenceTests.cs
```diff
@@ -209,8 +209,29 @@ public void GetOccurrencesShouldEnumerate()
...
+
+        CalDateTime[] expectedStartDates = [
+            new CalDateTime("20161114T000100", "W. Europe Standard Time"),
+            new CalDateTime("20161114T120100", "W. Europe Standard Time"),
+            new CalDateTime("20161121T000100", "W. Europe Standard Time"),
+            new CalDateTime("20161121T120100", "W. Europe Standard Time"),
+            new CalDateTime("20161128T120100", "W. Europe Standard Time"),
+            new CalDateTime("20161128T150100", "W. Europe Standard Time"),
+            new CalDateTime("20161205T000100", "W. Europe Standard Time"),
+            new CalDateTime("20161205T120100", "W. Europe Standard Time"),
```
axunonb: last comma is too much

Ical.Net.Tests/GetOccurrenceTests.cs
```diff
@@ -152,7 +152,7 @@ public void EnumerationChangedException()
     }
 
     [Test]
-    public void GetOccurrencesShouldEnumerate()
+    public void GetOccurrencesWithRecurrenceIdShouldEnumerate()
```
axunonb, line 155: Also works when `RECURRENCE-ID;VALUE=DATE` (it case this is a useful test, too)
```C#
    [Test]
    public void GetOccurrencesWithRecurrenceId_DateOnly_ShouldEnumerate()
    {
        const string ical = @"BEGIN:VCALENDAR
PRODID:-//github.com/ical-org/ical.net//NONSGML ical.net 5.0//EN
VERSION:2.0
BEGIN:VEVENT
UID:789012
DTSTART;VALUE=DATE:20231001
DTEND;VALUE=DATE:20231002
RRULE:FREQ=MONTHLY;BYMONTHDAY=1
SUMMARY:Monthly Report Due
END:VEVENT
BEGIN:VEVENT
UID:789012
RECURRENCE-ID;VALUE=DATE:20231101
DTSTART;VALUE=DATE:20231115
DTEND;VALUE=DATE:20231116
SUMMARY:Monthly Report Due (Rescheduled)
END:VEVENT
END:VCALENDAR
";

        var collection = Calendar.Load(ical);
        var startCheck = new DateTime(2023, 10 ,1);
        var occurrences = collection.GetOccurrences<CalendarEvent>(startCheck, startCheck.AddMonths(1))
            .OrderBy(x => x)
            .ToList();

        CalDateTime[] expectedStartDates = [
            new CalDateTime(2023, 10, 1),
            new CalDateTime(2023, 11, 15), // the replaced occurrence
            new CalDateTime(2023, 12,1)
        ];

        // For DTEND=20231002
        Assert.That(occurrences.Select(x => x.Period.StartTime), Is.EqualTo(expectedStartDates.Take(1)));

        // For DTEND=20231231
        var occurrences2 = collection.GetOccurrences<CalendarEvent>(new CalDateTime(startCheck), new CalDateTime(2023, 12, 31))
            .OrderBy(x => x)
            .ToList();

        Assert.Multiple(() =>
        {
            Assert.That(occurrences.Select(x => x.Period.StartTime), Is.EqualTo(expectedStartDates.Take(1)));
            Assert.That(occurrences2.Select(x => x.Period.StartTime), Is.EqualTo(expectedStartDates.Take(3)));
        });
    }
```

axunonb (dismissed):
LGTM, tnx

Ical.Net.Tests/GetOccurrenceTests.cs
minichma: Great, thank you for the test!

Ical.Net.Tests/GetOccurrenceTests.cs
minichma: Removed it. Usually I prefer to have it (the comma after the last entry of a list) for a number of reasons, because when adding entries in future changes, Git doesn't detect it as changes, its easier to move lines up and down, and generally looks more consistent. But all that's certainly also a question of taste.

axunonb (approved):
Fine, tnx

minichma:
Thank you for the review!

PR #664, opened 2024-12-06
Refactor `MatchTimeZone method` in `RecurrencePatternEvaluator`

### LINKED ISSUE #663: Refactor RecurrencePattern.Until
Remove necessity for setting the initial value to DateTime.MinValue (make it nullable)
Should `RecurrencePattern.Until` become a `IDateTime` instead of `DateTilme`?
Fix the method to determine the zoned date/time value
https://github.com/ical-org/ical.net/blob/add58a9bc3dffb8014a59d0bc6a831dbfa5635f7/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L30-L34
`MatchTimeZone` contains noop code back to v4.1.9

axunonb:
* Simplified handling of UNTIL value in `ProcessRecurrencePattern`.
* Refactored `MatchTimeZone` to ensure UNTIL matches DTSTART's time zone and type following RFC 5545 definitions
* Added unit tests to verify behavior under various scenarios, improving accuracy and consistency of recurrence evaluations.

Closes #663

### DIFF
--- a/Ical.Net.Tests/MatchTimeZoneTests.cs
+++ b/Ical.Net.Tests/MatchTimeZoneTests.cs
@@ -1,0 +1,233 @@
+﻿//
+// Copyright ical.net project maintainers and contributors.
+// Licensed under the MIT license.
+//
+
+using System;
+using System.Linq;
+using Ical.Net.DataTypes;
+using NUnit.Framework;
+
+namespace Ical.Net.Tests;
+
+[TestFixture]
+public class MatchTimeZoneTests
+{
+    [Test, Category("Recurrence")]
+    public void MatchTimeZone_LocalTimeUsaWithTimeZone()
+    {
+        // DTSTART with local time and time zone reference (negative offset), UNTIL as UTC
+        const string ical =
+            """
+            BEGIN:VCALENDAR
+            VERSION:2.0
+            PRODID:-//Example Corp//NONSGML Event//EN
+            BEGIN:VEVENT
+            UID:example1
+            SUMMARY:Event with local time and time zone
+            DTSTART;TZID=America/New_York:20231101T090000
+            RRULE:FREQ=DAILY;UNTIL=20231105T130000Z
+            DTEND;TZID=America/New_York:20231101T100000
+            END:VEVENT
+            END:VCALENDAR
+            """;
+
+        var calendar = Calendar.Load(ical);
+        var evt = calendar.Events.First();
+        var until = evt.RecurrenceRules.First().Until;
+
+        var expectedUntil = new DateTime(2023, 11, 05, 13, 00, 00, DateTimeKind.Utc);
+        var occurrences = evt.GetOccurrences(new CalDateTime(2023, 11, 01), new CalDateTime(2023, 11, 06));
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(until, Is.EqualTo(expectedUntil));
+            Assert.That(occurrences.Count, Is.EqualTo(4));
+            /*
+               Should have 4 occurrences:
+               November 1, 2023: 09:00 AM - 10:00 AM (UTC-0400) (America/New_York)
+               November 2, 2023: 09:00 AM - 10:00 AM (UTC-0400) (America/New_York)
+               November 3, 2023: 09:00 AM - 10:00 AM (UTC-0400) (America/New_York)
+               November 4, 2023: 09:00 AM - 10:00 AM (UTC-0400) (America/New_York)
+
+               November 5, 2023: 09:00 AM - 10:00 AM (UTC-0500) (America/New_York)
+               must NOT be included, because 20231105T130000Z => November 5, 2023: 08:00 AM (America/New_York)
+               (Daylight Saving Time in America/New_York ended on Sunday, November 5, 2023, at 2:00 AM)
+           */
+        });
+    }
+
+    [Test, Category("Recurrence")]
+    public void MatchTimeZone_LocalTimeJapanWithTimeZone()
+    {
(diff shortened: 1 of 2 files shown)


Ical.Net.Tests/MatchTimeZoneTests.cs
```diff
@@ -0,0 +1,233 @@
...
+        // DTSTART with local time and time zone reference (positive offset), UNTIL as UTC
+        const string ical =
+            """
+            BEGIN:VCALENDAR
+            VERSION:2.0
+            PRODID:-//Example Corp//NONSGML Event//EN
+            BEGIN:VEVENT
+            UID:example1
+            SUMMARY:Event with local time and time zone
+            DTSTART;TZID=Asia/Tokyo:20231101T090000
```
minichma: In the test above with NY tz you chose dates involving a DST change (in 2023 in NY winter DST change was Nov 5), but in this test no DST change is involved (there is no DST in Tokyo). Not a problem, just noting this to make sure, its intentional.

minichma (dismissed):

Ical.Net.Tests/MatchTimeZoneTests.cs
axunonb: Good point. I updated the test to use Sydney/Australia with DST change.

Ical.Net.Tests/MatchTimeZoneTests.cs
```diff
@@ -80,21 +80,25 @@ public void MatchTimeZone_LocalTimeJapanWithTimeZone()
...
-               November 4, 2023: 09:00 AM - 10:00 AM (UTC+0900) (Asia/Tokyo)
-               November 5, 2023: 09:00 AM - 10:00 AM (UTC+0900) (Asia/Tokyo)
-             */
+               October 1, 2024: 01:00 AM - 02:00 AM (UTC+1000) (Australia/Sydney)
+               October 2, 2024: 01:00 AM - 02:00 AM (UTC+1000) (Australia/Sydney)
+               October 3, 2024: 01:00 AM - 02:00 AM (UTC+1000) (Australia/Sydney)
+               October 4, 2024: 01:00 AM - 02:00 AM (UTC+1000) (Australia/Sydney)
+               October 5, 2024: 01:00 AM - 02:00 AM (UTC+1000) (Australia/Sydney)
+
+               October 6, 2024: 01:00 AM - 02:00 AM (UTC+1100) (Australia/Sydney)
```
minichma, line 104: Oh, actually I think, this *should* be included, because `UNTIL` must be considered *inclusive*.

> If the value
      specified by UNTIL is synchronized with the specified recurrence,
      this DATE or DATE-TIME becomes the last instance of the
      recurrence.

axunonb: Yes, the conclusion came from the description I assume. It was not updated.
Now there are 2 test cases, one when DST occurs.

minichma: Yes, I was looking at the comments. But now reconsidering the test, I still think its not correct.

October 6, 2024: 01:00 AM Sydney time = 20241005T140000Z, so it should be included in the first test and the number of occurrences should therefore be 6 in both cases.

Ical.Net.Tests/MatchTimeZoneTests.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        var calendar = Calendar.Load(ical);
+        var evt = calendar.Events.First();
+        var until = evt.RecurrenceRules.First().Until;
+
+        var expectedUntil = DateTime.ParseExact(inputUntil, "yyyyMMddTHHmmssZ",
+            System.Globalization.CultureInfo.InvariantCulture,
+            System.Globalization.DateTimeStyles.AssumeUniversal |
+            System.Globalization.DateTimeStyles.AdjustToUniversal);
+        var occurrences = evt.GetOccurrences(new CalDateTime(2024, 10, 01), new CalDateTime(2024, 10, 07));
```
minichma, line 89: In this case DTSTART has a tzid but the startTime is floating. This is quite a special case and we should considere to disallow the case. Anyhow, I think it would be preferable to specify a tzid for the start/endTime too.

Ical.Net.Tests/MatchTimeZoneTests.cs
axunonb: Isn't UTC Time: 2024-10-05 14:00:00 => Sydney Time (AEST, UTC+10:00): 2024-10-06 00:00:00?

minichma: Oh yes, of course, sorry for the confusion!

minichma (approved):

Ical.Net.Tests/MatchTimeZoneTests.cs
axunonb: Hm, CalDateTime(2024, 10, 01) resolves to a DATE value, which cannot have a timezone (RFC 5545)?

axunonb: So you mean, it should always be a zoned DATE-TIME?

minichma: Just a comment, not too relevant. Generally it would be advisable for users of the app, not to mix DATE vs DATE-TIME and floating vs non-floating, because there's always some ambiguity that causes additional complexity. In this case it really doesn't matter, so no need to change it. But as its advisable for users, not to mix it, we generally might want to avoid mixing them in the tests either (except for those cases where we explicitly want to test mixing them).

Issue #641, opened 2024-11-11
Make max. number of calculated occurrences configurable


axunonb:
Replace the constant `_maxIncrementCount` with a configurable variable
Note: The default should be "Infinite", but for occurences without an end date a limitation can be useful

https://github.com/ical-org/ical.net/blob/95bbc53991009e64c208b827f3c89bd49680c6bc/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L49-L52

axunonb:
Duplicate of #747

PR #665, opened 2024-12-07
Let `GetOccurrences()` et al return an `IEnumerable` rather than `HashSet`.

minichma:
This PR modifies `Calendar.GetOccurrences()` and related methods to return an ordered `IEnumerable` that generates the actual occurrences as they are enumerated, rather than returning a fully populated `HashSet`. When invoking `GetOccurrences`, the start- and endTime can now be omitted. The change brings several benefits:

* Even a calendar with undefinite recurrences can be enumerated without specifying an endTime, which can be helpful, if the endTime isn't known.
* A consumer of the lib can now easily find the first occurrence or the next after a certain date by calling something like`GetOccurrences(startTime: t).GetFirstOrDefault()`
* The occurrences are generated in ascending order
* The number of generated occurrences can be reduced, if not all are enumerated.
* The overall performance is increased, even if all occurrences are returned, because the implementation requires less sorting operations.

Modifications implemented:
* Return `IEnumerable` from `GetOccurrences` and `IEvaluator.Evaluate`
* Allow omitting `startTime` and `endTime` when calling `GetOccurrences`
* Rename the overload of `GetOccurrences` that returns the occurrences of a single day to `GetOccurrencesOfDay`, to avoid ambiguities with the other overloads, now that start/end time can be omitted.

### DIFF
--- a/Ical.Net.Tests/CollectionHelpersTests.cs
+++ b/Ical.Net.Tests/CollectionHelpersTests.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Ical.Net.DataTypes;
+using Ical.Net.Utility;
 using NUnit.Framework;
 
 namespace Ical.Net.Tests;
@@ -14,11 +15,7 @@ namespace Ical.Net.Tests;
 internal class CollectionHelpersTests
 {
     private static readonly DateTime _now = DateTime.UtcNow;
-    private static readonly DateTime _later = _now.AddHours(1);
-    private static readonly string _uid = Guid.NewGuid().ToString();
 
-    private static List<RecurrencePattern> GetSimpleRecurrenceList()
-        => new List<RecurrencePattern> { new RecurrencePattern(FrequencyType.Daily, 1) { Count = 5 } };
     private static List<PeriodList> GetExceptionDates()
         => new List<PeriodList> { new PeriodList { new Period(new CalDateTime(_now.AddDays(1).Date)) } };
 
@@ -36,5 +33,66 @@ internal class CollectionHelpersTests
         changedPeriod.First().First().StartTime = new CalDateTime(_now.AddHours(-1));
 
         Assert.That(changedPeriod, Is.Not.EqualTo(GetExceptionDates()));
+    }
+
+    [TestCase(new[] { 1, 3, 5, 7 }, new[] { 2, 4, 6 }, new[] { 1, 2, 3, 4, 5, 6, 7 })]
+    [TestCase(new int[] { }, new int[] { }, new int[] { })]
+    [TestCase(new int[] { }, new[] { 2, 4, 6 }, new[] { 2, 4, 6 })]
+    [TestCase(new[] { 2, 4, 6 }, new int[] { }, new[] { 2, 4, 6 })]
+    [TestCase(new[] { 3, 4 }, new int[] { 1, 2 }, new[] { 1, 2, 3, 4 })]
+    [TestCase(new[] { 1, 2, 3 }, new int[] { 2, 3, 4 }, new[] { 1, 2, 2, 3, 3, 4 })]
+    public void TestMerge(IList<int> seq1, IList<int> seq2, IList<int> expected)
+    {
+        var result = seq1.OrderedMerge(seq2).ToList();
+
+        Assert.That(result, Is.EqualTo(expected));
+    }
+
+    [TestCase(new int[] { }, new int[] { }, new int[] { })]
+    [TestCase(new int[] { }, new[] { 2, 4, 6 }, new int[] { })]
+    [TestCase(new[] { 2, 4, 6 }, new int[] { }, new[] { 2, 4, 6 })]
+    [TestCase(new[] { 1, 2, 3, 5, 6, 7 }, new[] { 2, 4, 6 }, new[] { 1, 3, 5, 7 })]
+    public void TestMergeExclude(IList<int> seq, IList<int> exclude, IList<int> expected)
+    {
+        var result = seq.OrderedExclude(exclude).ToList();
+
+        Assert.That(result, Is.EqualTo(expected));
+    }
+
+    private static IEnumerable<int> GetNaturalNumbers()
+    {
+        var i = 1;
+        while (true)
+            yield return i++;
+    }
+
+    [Test]
+    public void TestMergeIndefinite()
+    {
+        var result = GetNaturalNumbers().Select(x => x * 3).OrderedMerge(GetNaturalNumbers().Select(x => x * 2))
+            .Take(7);
+        Assert.That(result, Is.EqualTo(new[] { 2, 3, 4, 6, 6, 8, 9 }));
+    }
+
+    [Test]
+    public void TestMergeExcludeIndefinite()
+    {
+        var result = GetNaturalNumbers().Select(x => x * 3).OrderedExclude(GetNaturalNumbers().Select(x => x * 2))
+            .Take(4);
+        Assert.That(result, Is.EqualTo(new[] { 3, 9, 15, 21 }));
+    }
+
+    [Test]
+    public void TestMergeMulti()
+    {
+        var result = CollectionHelpers.OrderedMergeMany([[4], [2], [3], [1]], Comparer<int>.Default);
+        Assert.That(result, Is.EqualTo(new[] { 1, 2, 3, 4 }));
+    }
+
+    [Test]
+    public void TestOrderedDistinct()
+    {
+        var result = new[] { 1, 2, 2, 3 }.OrderedDistinct();
+        Assert.That(result, Is.EqualTo(new[] { 1, 2, 3 }));
     }
 }
--- a/Ical.Net/Calendar.cs
+++ b/Ical.Net/Calendar.cs
@@ -204,13 +204,13 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
     /// </summary>
     /// <param name="dt">The date for which to return occurrences. Time is ignored on this parameter.</param>
     /// <returns>A list of occurrences that occur on the given date (<paramref name="dt"/>).</returns>
-    public virtual IEnumerable<Occurrence> GetOccurrences(IDateTime dt)
+    public virtual IEnumerable<Occurrence> GetOccurrencesOfDay(IDateTime dt)
     {
         return GetOccurrences<IRecurringComponent>(new CalDateTime(dt.Date), new CalDateTime(dt.Date.AddDays(1)));
     }
 
-    /// <inheritdoc cref="GetOccurrences(IDateTime)"/>
-    public virtual IEnumerable<Occurrence> GetOccurrences(DateTime dt)
+    /// <inheritdoc cref="GetOccurrencesOfDay(IDateTime)"/>
+    public virtual IEnumerable<Occurrence> GetOccurrencesOfDay(DateTime dt)
     {
         return GetOccurrences<IRecurringComponent>(new CalDateTime(DateOnly.FromDateTime(dt)), new CalDateTime(DateOnly.FromDateTime(dt.Date.AddDays(1))));
     }
@@ -246,7 +246,7 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
         return GetOccurrences<T>(new CalDateTime(dt.Date), new CalDateTime(dt.Date.AddDays(1)));
     }
 
-    /// <inheritdoc cref="GetOccurrences(IDateTime)"/>
+    /// <inheritdoc cref="GetOccurrencesOfDay(IDateTime)"/>
     public virtual IEnumerable<Occurrence> GetOccurrences<T>(DateTime dt) where T : IRecurringComponent
     {
         return GetOccurrences<T>(new CalDateTime(DateOnly.FromDateTime(dt)), new CalDateTime(DateOnly.FromDateTime(dt.Date.AddDays(1))));
--- a/Ical.Net/Calendar.cs
+++ b/Ical.Net/Calendar.cs
@@ -204,13 +204,13 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
     /// </summary>
     /// <param name="dt">The date for which to return occurrences. Time is ignored on this parameter.</param>
     /// <returns>A list of occurrences that occur on the given date (<paramref name="dt"/>).</returns>
-    public virtual HashSet<Occurrence> GetOccurrences(IDateTime dt)
+    public virtual IEnumerable<Occurrence> GetOccurrences(IDateTime dt)
     {
         return GetOccurrences<IRecurringComponent>(new CalDateTime(dt.Date), new CalDateTime(dt.Date.AddDays(1)));
     }
 
     /// <inheritdoc cref="GetOccurrences(IDateTime)"/>
-    public virtual HashSet<Occurrence> GetOccurrences(DateTime dt)
+    public virtual IEnumerable<Occurrence> GetOccurrences(DateTime dt)
     {
         return GetOccurrences<IRecurringComponent>(new CalDateTime(DateOnly.FromDateTime(dt)), new CalDateTime(DateOnly.FromDateTime(dt.Date.AddDays(1))));
     }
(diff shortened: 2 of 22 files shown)


Ical.Net.Tests/CollectionHelpersTests.cs
```diff
@@ -7,7 +7,9 @@
 using System.Collections.Generic;
 using System.Linq;
 using Ical.Net.DataTypes;
+using Ical.Net.Utility;
 using NUnit.Framework;
+using NUnit.Framework.Legacy;
```
axunonb: not in use

axunonb: GitHub warnings The actual and the expected argument is the same 'GetExceptionDates()' would be nice to disappear.

axunonb: not in use
```C#
    private static readonly DateTime _later = _now.AddHours(1);
    private static readonly string _uid = Guid.NewGuid().ToString();

    private static List<RecurrencePattern> GetSimpleRecurrenceList()
        => new List<RecurrencePattern> { new RecurrencePattern(FrequencyType.Daily, 1) { Count = 5 } };
```

Ical.Net/Calendar.cs
```diff
@@ -272,14 +272,27 @@ public virtual HashSet<Occurrence> GetOccurrences<T>(IDateTime startTime, IDateT
...
-        var occurrences = new HashSet<Occurrence>(RecurringItems
+        var occurrences = RecurringItems
             .OfType<T>()
-            .SelectMany(recurrable => recurrable.GetOccurrences(startTime, endTime))
+            .Select(recurrable => recurrable.GetOccurrences(startTime, endTime))
+
+            // Enumerate the list of occurrences (not the occurrences themselves) now to ensure
+            // the initialization code is run, including validation and error handling.
+            // This way we receive validation errors early, not only when enumeration starts.
+            .ToList()
```
axunonb: very good

Ical.Net/Calendar.cs
```diff
@@ -272,14 +272,27 @@ public virtual HashSet<Occurrence> GetOccurrences<T>(IDateTime startTime, IDateT
...
+
+            // Merge the individual sequences into a single one. Take advantage of them
+            // being ordered to avoid full enumeration.
+            .OrderedMergeMany()
```
axunonb, line 245: that's really cool part :)

Ical.Net/CalendarCollection.cs
```diff
@@ -39,85 +39,57 @@ public static CalendarCollection Load(TextReader tr)
         return collection;
     }
 
-    public HashSet<Occurrence> GetOccurrences(IDateTime dt)
+    public HashSet<Occurrence> GetOccurrencesOfDay(IDateTime dt)
```
axunonb: not referenced, probably left to remove

Ical.Net/CalendarCollection.cs
```diff
@@ -39,85 +39,57 @@ public static CalendarCollection Load(TextReader tr)
...
+        .ToArray()
+        .OrderedMergeMany();
+
+    public IEnumerable<Occurrence> GetOccurrences<T>(DateTime? startTime, DateTime? endTime) where T : IRecurringComponent
+        => this
+        .Select(iCal => iCal.GetOccurrences<T>(startTime, endTime))
+        .ToArray()
+        .OrderedMergeMany();
 
     private FreeBusy CombineFreeBusy(FreeBusy main, FreeBusy current)
```
axunonb, line 70: FreeBusy method are not referenced - do we still need them?

Ical.Net/DataTypes/Occurrence.cs
```diff
@@ -4,6 +4,7 @@
 //
 
 using System;
+using System.Collections.Generic;
```
axunonb: not in use

Ical.Net/Evaluation/PeriodListEvaluator.cs
```diff
@@ -18,9 +18,9 @@ public PeriodListEvaluator(PeriodList rdt)
         _mPeriodList = rdt;
     }
 
-    public override HashSet<Period> Evaluate(IDateTime referenceDate, DateTime periodStart, DateTime periodEnd, bool includeReferenceDateInResults)
+    public override IEnumerable<Period> Evaluate(IDateTime referenceDate, DateTime? periodStart, DateTime? periodEnd, bool includeReferenceDateInResults)
     {
-        var periods = new HashSet<Period>();
+        var periods = new SortedSet<Period>();
```
axunonb, line 23: much better

Ical.Net/Evaluation/RecurrenceUtil.cs
```diff
@@ -35,21 +32,23 @@ public static HashSet<Occurrence> GetOccurrences(IRecurrable recurrable, IDateTi
...
         // so they can be used during the evaluation process.
 
-        periodStart = new CalDateTime(periodStart.Date, periodStart.Time, start.TzId);
-        periodEnd = new CalDateTime(periodEnd.Date, periodEnd.Time, start.TzId);
+        if (periodStart != null)
+            periodStart = new CalDateTime(periodStart.Date, periodStart.Time, start.TzId);
+        if (periodEnd != null)
+            periodEnd = new CalDateTime(periodEnd.Date, periodEnd.Time, start.TzId);
 
-        var periods = evaluator.Evaluate(start, DateUtil.GetSimpleDateTimeData(periodStart), DateUtil.GetSimpleDateTimeData(periodEnd),
```
axunonb, line 41: very good, I forgot that remove that :)

Ical.Net/Evaluation/RecurringEvaluator.cs
```diff
@@ -40,42 +41,38 @@ public RecurringEvaluator(IRecurrable obj)
...
+        var periodsQueries = Recurrable.RecurrenceRules.Select(rule =>
         {
             var ruleEvaluator = rule.GetService(typeof(IEvaluator)) as IEvaluator;
             if (ruleEvaluator == null)
             {
                 return Enumerable.Empty<Period>();
             }
             return ruleEvaluator.Evaluate(referenceDate, periodStart, periodEnd, includeReferenceDateInResults);
-        });
+        }).ToList();
```
axunonb: no deferred execution in order to convert the result to a list to ensure that all recurrence rules are fully evaluated before any further processing?

Ical.Net/Evaluation/RecurringEvaluator.cs
```diff
@@ -85,24 +82,22 @@ protected HashSet<Period> EvaluateRDate(IDateTime referenceDate, DateTime period
...
+        var exRuleEvaluatorQueries = Recurrable.ExceptionRules.Select(exRule =>
         {
             var exRuleEvaluator = exRule.GetService(typeof(IEvaluator)) as IEvaluator;
             if (exRuleEvaluator == null)
             {
                 return Enumerable.Empty<Period>();
             }
             return exRuleEvaluator.Evaluate(referenceDate, periodStart, periodEnd, false);
-        });
+        }).ToList();
```
axunonb: same as above
(no deferred execution in order to convert the result to a list to ensure that all recurrence rules are fully evaluated before any further processing?)

Ical.Net/Evaluation/RecurringEvaluator.cs
```diff
@@ -112,49 +107,50 @@ protected HashSet<Period> EvaluateExRule(IDateTime referenceDate, DateTime perio
...
         return periods;
     }
 
-    private static HashSet<Period> FindDateOverlaps(HashSet<Period> periods, HashSet<Period> dates)
+    /// <summary>
+    /// Compares whether the given period's date overlaps with the given EXDATE. The dates are
+    /// considered to overlap if they start at the same time, or the EXDATE is an all-day date
+    /// and the period's start date is the same as the EXDATE's date.
+    /// </summary>
+    private static int CompareDateOverlap(Period period, Period exDate)
```
axunonb, line 155: exDate still a period unfortunately, but that's for another PR

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -89,4 +89,182 @@ public static void AddRange<T>(this ICollection<T> destination, IEnumerable<T> s
             destination.Add(element);
         }
     }
-}
\ No newline at end of file
+
+    /// <summary>
```
axunonb, line 93: wonderful!

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -89,4 +89,182 @@ public static void AddRange<T>(this ICollection<T> destination, IEnumerable<T> s
...
+                hasNextIt = it.MoveNext();
+            }
+            else
+            {
+                hasNextEx = itEx.MoveNext();
+            }
+        }
+    }
+
+        /// <summary>
```
axunonb: format

Ical.Net/Utility/DateUtil.cs
```diff
@@ -24,6 +24,9 @@ public static IDateTime EndOfDay(IDateTime dt)
     public static DateTime GetSimpleDateTimeData(IDateTime dt)
         => dt.Value;
 
+    public static CalDateTime AsCalDateTime(this DateTime t)
```
axunonb, line 27: I remember, that's what you mentioned lately :)

Ical.Net/VTimeZoneInfo.cs
```diff
@@ -9,6 +9,7 @@
 using Ical.Net.CalendarComponents;
 using Ical.Net.DataTypes;
 using Ical.Net.Evaluation;
+using Ical.Net.Utility;
```
axunonb, line 12: maybe resolve 'VTimeZoneInfo' overrides Object.Equals(object o) but does not override Object.GetHashCode() ?

axunonb (dismissed):
See my remarks and decide how to deal with them.
Same with flags from SonarCloud .
Overall this is an amazing job, tnx!

axunonb:
Does this PR make #641 redundant, and could we remove `RecurrencePatternEvaluator.MaxIncrementCount = 1000`?

Ical.Net/Utility/CollectionHelpers.cs
axunonb: and also nice code coverage

Ical.Net/CalendarCollection.cs
```diff
@@ -39,85 +39,57 @@ public static CalendarCollection Load(TextReader tr)
...
-    public HashSet<Occurrence> GetOccurrences<T>(DateTime startTime, DateTime endTime) where T : IRecurringComponent
-    {
-        var occurrences = new HashSet<Occurrence>();
-        foreach (var iCal in this)
-        {
-            occurrences.UnionWith(iCal.GetOccurrences<T>(startTime, endTime));
-        }
-        return occurrences;
-    }
+    public IEnumerable<Occurrence> GetOccurrencesOfDay(DateTime dt)
```
axunonb: Would you like to keep the unused (and thus uncovered overloads)?

minichma:
@axunonb Thank you for the review, glad you like the change. Appreciate your comments. My previous few PRs have been based on branches in the ical-org repo, this one is based on my minichma repo. When based on the ical-org, a code coverage report is automatically generated, but obviously not if based on a different one. Could we change that, so we get the report for every PR, also when based on an external repo?

Ical.Net/CalendarCollection.cs
minichma: I generally didn't remove any methods, because I don't want to change the API in this PR outside of what is related to the immediate topic. If we want to change it, I think we should do that in a separate PR, so we keep concerns separated. But this one indeed is a special case. Actually I don't think, we should have a separate method just for getting the occurrences for a single day, because it can easily be implemented with the other variants and its not obvious, why we wouldn't also add versions for ByWeek, ByHour, etc. So I'd certainly vote for removing them, but also the other overloads (i.e. `GetOccurrencesOfDay(IDateTime)`).

In this case we have another challenge. Due to the params of `GetOccurrences(DateTime? = null, DateTime? = null)` et al being optional now, a call like `GetOccurrences(new DateTime(...))` would now invoke the overload with startTime, endTime, rather than the overload for getting the occurrences for a single day (which has been renamed). This could break existing code without being obvious. It will be introduced in v5, so we could argue that's ok (and I actually would do so), but we should be aware of it. And add a note to a potential migration guide. What do you think?

minichma:
> Does this PR make #641 redundant, and could we remove `RecurrencePatternEvaluator.MaxIncrementCount = 1000`?

Actually it doesn't, because `MaxIncrementCount` is related to the number of increments *between* recurrences and in this respect nothing changed. Actually the title of #641 is not fully precise on that.

Ical.Net/Utility/DateUtil.cs
minichma: Exactly. In this PR its particularly helpful, because it works well with the null-propagating operator.

Ical.Net/Evaluation/RecurringEvaluator.cs
minichma: Yes. As the PR is quite big, I tried to leave things the way they are, where possible (or implemented PRs upfront like #659), but that's not always obvious from the diff.

axunonb:
> Could we change that, so we get the report for every PR, also when based on an external repo?

Unfortunately we can't, as GitHub secrets can't be applied to external PRs, Just continue with branches inside ical-org.

Ical.Net/Evaluation/RecurringEvaluator.cs
minichma: Actually only the list of sequences is enumerated here, so we get a `List<IEnumerable<Period>>`, which is intentional, so the initialization code of each RRULE is run. The inner enumerables themselves are not enumerated. Will add some comments.

Ical.Net/Evaluation/RecurringEvaluator.cs
minichma: Same as above. Will add comments.

Ical.Net/Evaluation/PeriodListEvaluator.cs
minichma: and actually a prerequisite of merging with other IEnumerables later on.

Ical.Net/CalendarCollection.cs
minichma: I have no idea, but as they are not related to this PR, I suggest to deal with that in a different PR (see above).

Ical.Net/Calendar.cs
minichma: Yeah, I also like it :-)

Ical.Net/CalendarCollection.cs
minichma: @axunonb I removed the versions of `GetOccurrences(date)` that were handling a single day altogether (see prev comment). Only two tests existed that tested any of them at all. I don't think, they will be missing. We should add some comment in a migration guide, should one be written, about the ambiguity due to the new nullable params in the other overloads.

minichma:
Comments related to the new issues identified by Sonarcloud:

> Drop this useless call to 'ToList' or replace it by 'AsEnumerable' if you are using LINQ to Entities.

They aren't useless, because they cause an intended iteration. The reason is mentioned in the code comments.

> Add the default parameter value defined in the overridden method.

TBD. I actually left the default params out intentionally. In the interface they are there for the convenience of users, but internally I think we should be more verbose and explicitly specify the params.

> Refactor this method to reduce its Cognitive Complexity from 39 to the 15 allowed.

Not a topic of this PR. Some simplification has already been implemented in the mentioned method though.

axunonb:
> Comments related to the new issues identified by Sonarcloud:

Yep, so we could mark the issues deliberately chosen as described in the comments with //NOSONAR (or a warning disable pragma, and leave the To-dos as issues to remember.
This helps to distinguish issues eventually to be fixed and others, which are accepted and won't appear on the SonarCloud list.
Cognitive Complexity level of 15 is low, I have 20 in other projects. We could adjust for ical.net, too.

minichma:
> > Comments related to the new issues identified by Sonarcloud:
> 
> Yep, so we could mark the issues deliberately chosen as described in the comments with //NOSONAR (or a warning disable pragma, and leave the To-dos as issues to remember. This helps to distinguish issues eventually to be fixed and others, which are accepted and won't appear on the SonarCloud list. Cognitive Complexity level of 15 is low, I have 20 in other projects. We could adjust for ical.net, too.

Tweaked a little. Not too convinced this is helpful for the overall code quality. I'm not the biggest fan of cluttering code with 'disable' comments, so I undid a few changes. We might want to reconsider some of the rules over time.

axunonb (approved):
Tnx, excellent PR

axunonb:
> We might want to reconsider some of the rules over time.

Absolutely, we're currently using the default "Sonar Way" rules.

PR #667, opened 2024-12-09
Make `RecurrencePattern.Count`, `.Interval` and `WeekDay.Offset` nullable to avoid using `MinValue` as magic number.

minichma:
Make the following properties nullable to avoid using `MinValue` as magic number:

* `RecurrencePattern.Count`
* `RecurrencePattern.Interval`
* `WeekDay.Offset`

Closes #471

### DIFF
--- a/Ical.Net.Tests/DataTypeTest.cs
+++ b/Ical.Net.Tests/DataTypeTest.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license.
 //
 
+using System.Collections.Generic;
 using Ical.Net.DataTypes;
 using NUnit.Framework;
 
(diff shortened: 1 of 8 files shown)


minichma:
@axunonb I had this in the pipeline and as you just did the same for `Until` in #666 I thought I'd make a PR quickly, before we do the work twice. After both have been merged, it would be the time to remove the `MinValue` stuff from `RecurrencePatternSerializer.CheckMutuallyExclusive` (#470, #471) I assume.

axunonb (approved):
LGTM, and sorry for interfering re `UNTIL`

minichma:
> LGTM, and sorry for interfering re `UNTIL`

Oh, no worries, I didn't do anything with `UNTIL`, just wanted to avoid that you would implement the things from this PR, which would have been double work. Thanks for the review!

Issue #587, opened 2024-03-12
Read  Organizer.SentBy throw exception


idiotsky:
When Parameters.Get("SENT-BY") return null. when read Organizer.SentBy will throw exception.
`
public virtual Uri SentBy
    {
      get => new Uri(this.Parameters.Get("SENT-BY"));    
    }
`
should be something like 
`
public virtual Uri SentBy
    {
      get 
     {
           if(this.Parameters.Get("SENT-BY") != null)
          {
              new Uri(this.Parameters.Get("SENT-BY")); 
          }
          return null.
      }  
    }
`

axunonb:
Appreciate a PR. Please double-check the new version merged with #595

axunonb:
Closed with #669

PR #669, opened 2024-12-11
Fix when no "SENT-BY" defined, get SentBy will throw exception

idiotsky:


axunonb:
Thanks for submitting the PR.
Actually there several things to consider while heading to a new v5.
Please understand the comments under this aspect.

The `Organizer` class might require more similar refactoring as proposed in the PR.
Examples:

- The class should up updated to use NRT.
- Should the setter for `SentBy` remove an existing `Parameters` entry?
- The getter for `DirectoryEntry` causes a similar issue. it should return `null` if no `DIR` entry exists.
- Decision about errors when creating the `Uri` - throw or use `null` instead? This may also concern the `ParameterCollectionProxy.Set` method.
- Unit tests missing

@minichma What do you think. Should we fix this issue in an isolated or a "fix it all" context?

minichma:
I agree. This PR certainly is beneficial, but the same improvement as for `SentBy` should also be applied to `DirectoryEntry` at least.

@idiotsky Would you be ready to adjust the PR accordingly?

Generally speaking, I think that representing these values as `Uri` is not the best thing to do, so at some point we might consider replacing all the `Uri` values by simple `string` ones.

idiotsky:
> I agree. This PR certainly is beneficial, but the same improvement as for `SentBy` should also be applied to `DirectoryEntry` at least.
> 
> @idiotsky Would you be ready to adjust the PR accordingly?
> 
> Generally speaking, I think that representing these values as `Uri` is not the best thing to do, so at some point we might consider replacing all the `Uri` values by simple `string` ones.

I agree with replace the Uri to string is better idea.

idiotsky:
I submit the changes, any feed back, I'll try to fix it.

minichma (approved):
LGTM

minichma:
@idiotsky Thank you!

idiotsky:
> @idiotsky Thank you!

you are welcome

Sansoeko:
Closes #587

PR #673, opened 2024-12-18
Fix: Calculation for occurrences include DST transitions correctly

### LINKED ISSUE #298: Every day for all of time, except Tuesday,Thursday,Saturday,and Sunday" Not working
var vEvent = new Event
                {
                    Summary = "BIO CLASS",//subject
                    Description = "Details at CLASS",//description of meeting
                    Location = "Building 101",//location
                    DtStart = new CalDateTime(DateTime.Parse("2017-06-01T08:00")),
                    DtEnd = new CalDateTime(DateTime.Parse("2017-06-01T09:30")),
                    RecurrenceRules = new List<IRecurrencePattern> { new RecurrencePattern(FrequencyType.Daily, 1) },
                };

                //Define the exceptions: Sunday
                var exceptionRule = new RecurrencePattern(FrequencyType.Weekly, 1)
                {
                    ByDay = new List<IWeekDay> { new WeekDay(DayOfWeek.Sunday), new WeekDay(DayOfWeek.Saturday),
                    new WeekDay(DayOfWeek.Tuesday),new WeekDay(DayOfWeek.Thursday)}
                };
                vEvent.ExceptionRules = new List<IRecurrencePattern> { exceptionRule };

                calendar.Events.Add(vEvent);

### LINKED ISSUE #342: Bug - Invalid recurring event occurrences after New Year
Creating recurring event which is repeated every two weeks for all week days and starts in the middle of the week, is supposed to return occurrences until Sunday and then skips next week.

But when week ends on 31.12.2017 then occurrences for first week of the year are created.

Steps to reproduce:
1. Add event with following RRULE: 
     FREQ=WEEKLY;INTERVAL=2;BYDAY=MO,TU,WE,TH,FR,SA,SU
2. Set event start to 30.12.2017
3. Set event end to 29.01.2018
4. Get event occurrences.

Actual result:
1. 30.12.2017, 31.12.2017, 01.01.2018, 02.01.2018, 03.01.2018, 04.01.2018, 05.01.2018, 08.01.2018, 
    09.01.2018, 10.01.2018, 11.01.2018, 12.01.2018, 13.01.2018, 14.01.2018, 22.01.2018, 23.01.2018, 
    24.01.2018, 25.01.2018, 26.01.2018, 27.01.2018, 28.01.2018
2. Extra invalid dates: 01.01.2018, 02.01.2018, 03.01.2018, 04.01.2018, 05.01.2018

Expeted result:
30.12.2017, 31.12.2017, 08.01.2018, 09.01.2018, 10.01.2018, 11.01.2018, 12.01.2018, 13.01.2018, 14.01.2018, 22.01.2018, 23.01.2018, 24.01.2018, 25.01.2018, 26.01.2018, 27.01.2018, 28.01.2018

UT:
```c#
    [TestFixture]
    public class Tests
    {
        private static void CheckDates(DateTime startDate, DateTime endDate, DateTime[] expectedDates)
        {
            var rule = new RecurrencePattern(FrequencyType.Weekly, 2);

            rule.ByDay = new List<WeekDay>()
            {
                new WeekDay(DayOfWeek.Monday),
                new WeekDay(DayOfWeek.Tuesday),
                new WeekDay(DayOfWeek.Wednesday),
                new WeekDay(DayOfWeek.Thursday),
                new WeekDay(DayOfWeek.Friday),
                new WeekDay(DayOfWeek.Saturday),
                new WeekDay(DayOfWeek.Sunday),
            };

            var calendarEvent = new CalendarEvent()
            {
                DtStart = new CalDateTime(startDate),
                DtEnd = new CalDateTime(endDate),
                RecurrenceRules = new List<RecurrencePattern>()
                {
                    rule
                },
            };

            var occurences = calendarEvent.GetOccurrences(startDate, endDate);

            var occurrencesDates = occurences.Select(o => o.Period.StartTime.Date).ToArray();

            CollectionAssert.AreEquivalent(expectedDates, occurrencesDates);
        }

        [Test]
        public void NewYear_EachOtherWeek()
        {
            var startDate = new DateTime(2017,12,30);
            var endDate = new DateTime(2018,1, 29);
            var expectedDates = new[]
            {
                new DateTime(2017, 12, 30),
                new DateTime(2017, 12, 31),

                // 1-7 no events
                
                new DateTime(2018, 1, 8),
                new DateTime(2018, 1, 9),
                new DateTime(2018, 1, 10),
                new DateTime(2018, 1, 11),
                new DateTime(2018, 1, 12),
                new DateTime(2018, 1, 13),
                new DateTime(2018, 1, 14),

                // 15 - 21 no events

                new DateTime(2018, 1, 22),
                new DateTime(2018, 1, 23),
                new DateTime(2018, 1, 24),
                new DateTime(2018, 1, 25),
                new DateTime(2018, 1, 26),
                new DateTime(2018, 1, 27),
                new DateTime(2018, 1, 28),
            };

            //FAILS
            CheckDates(startDate, endDate, expectedDates);
        }

        [Test]
        public void December_EachOtherWeek()
        {
            var startDate = new DateTime(2017, 12, 1);
            var endDate = new DateTime(2017, 12, 31);
            var expectedDates = new[]
            {
                new DateTime(2017, 12, 1),
                new DateTime(2017, 12, 2),
                new DateTime(2017, 12, 3),
                // 4-10 no events
                
                new DateTime(2017, 12, 11),
                new DateTime(2017, 12, 12),
                new DateTime(2017, 12, 13),
                new DateTime(2017, 12, 14),
                new DateTime(2017, 12, 15),
                new DateTime(2017, 12, 16),
                new DateTime(2017, 12, 17),

                // 18 - 24 no events

                new DateTime(2017, 12, 25),
                new DateTime(2017, 12, 26),
                new DateTime(2017, 12, 27),
                new DateTime(2017, 12, 28),
                new DateTime(2017, 12, 29),
                new DateTime(2017, 12, 30),
               // new DateTime(2017, 12, 31), 
            };

            //PASS
            CheckDates(startDate, endDate, expectedDates);
        }

        [Test]
        public void NovemberEnd_December_EachOtherWeek()
        {
            var startDate = new DateTime(2017, 11, 30);
            var endDate = new DateTime(2017, 12, 31);
            var expectedDates = new[]
            {
                new DateTime(2017, 11, 30),
                new DateTime(2017, 12, 1),
                new DateTime(2017, 12, 2),
                new DateTime(2017, 12, 3),
                // 4-10 no events
                
                new DateTime(2017, 12, 11),
                new DateTime(2017, 12, 12),
                new DateTime(2017, 12, 13),
                new DateTime(2017, 12, 14),
                new DateTime(2017, 12, 15),
                new DateTime(2017, 12, 16),
                new DateTime(2017, 12, 17),

                // 18 - 24 no events

                new DateTime(2017, 12, 25),
                new DateTime(2017, 12, 26),
                new DateTime(2017, 12, 27),
                new DateTime(2017, 12, 28),
                new DateTime(2017, 12, 29),
                new DateTime(2017, 12, 30),               
            };

            // PASS
            CheckDates(startDate, endDate, expectedDates);
        }
    }
```

### LINKED ISSUE #567: GetOccurrences Creates event with invalid time due to Daylight Savings changes
Sample code to duplicate:

`var calStart = new CalDateTime(DateTimeOffset.Parse("2023-01-14T19:21:03.700Z").UtcDateTime, "UTC");
			var calFinish = new CalDateTime(DateTimeOffset.Parse("2023-03-14T18:21:03.700Z").UtcDateTime, "UTC");
			var tz = "Pacific Standard Time";
			var pattern = new RecurrencePattern("FREQ=WEEKLY;BYDAY=SU,MO,TU,WE");
			//Adjust the date to today so that the times remain constant
			DateTime localTZstartAdjust = DateTools.CombineDateTimes(DateTime.Now, TimeSpan.FromHours(2).Add(TimeSpan.FromMinutes(35)));
			DateTime localTZfinishAdjust = DateTools.CombineDateTimes(DateTime.Now, TimeSpan.FromHours(17));
			var ev = new Ical.Net.CalendarComponents.CalendarEvent {
				Class = "PUBLIC",
				IsAllDay = false,
				Start = new CalDateTime(localTZstartAdjust, tz),
				End = new CalDateTime(localTZfinishAdjust, tz),
				Sequence = 0,
				RecurrenceRules = new List<RecurrencePattern> { pattern }
			};
			var col = ev.GetOccurrences(calStart, calFinish);
			`

col will include an event that starts at 2:35 am on March 12, 2023 PST, but that time doesn't exist since this is when the time changes.

### LINKED ISSUE #623: Daylight saving transition bugs

### LINKED ISSUE #660: `VEVENT` duration not handled correctly in case of a DST change between `DTSTART` and `DTEND`
The following test fails, because there is a DST change between DTSTART and DTEND, which is not handled correctly:

```cs
    [Test]
    public void TestGetOccurrence_DtEndExcluded()
    {
        var ical = """
            BEGIN:VCALENDAR
            VERSION:2.0
            PRODID:Video Wall
            BEGIN:VEVENT
            UID:VW6
            DTSTAMP:20240630T000000Z
            DTSTART;TZID=Europe/London:20241001T000000
            DTEND;TZID=Europe/London:20241128T000000
            SUMMARY:New home speech.mp4
            COMMENT:New location announcement; may need update before Thanksgiving
            END:VEVENT
            END:VCALENDAR
            """;

        var calendar = Calendar.Load(ical);
        var occurrences = calendar.GetOccurrences(new CalDateTime("20241128T000000", "Europe/London"), new CalDateTime("20250101T000000", "Europe/London")).ToList();

        Assert.That(occurrences, Is.Empty);
    }
```

Related issues may appear in cases with recurring events where some of the occurrences do have DST changes between DTSTART and DTEND and others don't.

The following specification from RFC 5545 3.8.5.3 is also related to the topic, because it makes a distinction between the duration being specified via DTEND or via DURATION. Also related to #598

> If the duration of the recurring component is specified with the
      "DTEND" or "DUE" property, then the same exact duration will apply
      to all the members of the generated recurrence set.  Else, if the
      duration of the recurring component is specified with the
      "DURATION" property, then the same nominal duration will apply to
      all the members of the generated recurrence set and the exact
      duration of each recurrence instance will depend on its specific
      start time.  For example, recurrence instances of a nominal
      duration of one day will have an exact duration of more or less
      than 24 hours on a day where a time zone shift occurs.  The
      duration of a specific recurrence may be modified in an exception
      component or simply by using an "RDATE" property of PERIOD value
      type.

This is a follow-up of #656. While #656 discusses the case that DTSTART/DTEND are specified in floating time, this issue handles the slightly different case that DTSTART/DTEND specify a time zone.

### LINKED ISSUE #671: GetOccurences does not work properly when used with TimeZones
When using `GetOccurences` on recurring elements with a time zone, the data is not returned properly.
My example here is I set a recurring event on the 2nd of December, from midnight to 8am PST, repeating on Mondays.

Without time zones it works just fine.
```
private static void CheckCalendar(ServiceProvider provider)
{
    var calendar = new Ical.Net.Calendar();
    var blockDate = new DateTime(2024, 12, 2, 8, 0, 0, DateTimeKind.Utc);
    calendar.Events.Add(new CalendarEvent
    {
        Summary = "Some work",
        DtStart = new CalDateTime(blockDate),
        DtEnd = new CalDateTime(blockDate.Add(new TimeSpan(8, 0, 0))),
        IsAllDay = false,
        RecurrenceRules = new List<RecurrencePattern>
        {
            new RecurrencePattern
            {
                Frequency = Ical.Net.FrequencyType.Weekly,
                Interval = 1,
                ByDay = new List<WeekDay>
                {
                    new WeekDay(DayOfWeek.Monday)
                }
            }
        }
    });

    var start = new DateTime(2024, 12, 9, 8, 5, 0, DateTimeKind.Utc);
    var end = new DateTime(2024, 12, 10, 7, 59, 59, DateTimeKind.Utc);
    var x = calendar.GetOccurrences(start, end);
    Console.WriteLine(x.Count); // Is 1
}
```

Notice that my `start` is 5 minutes after the "DtStart" event, but I still get 1 entry.

Now I do the same thing but I attach the time zone on the `CalDateTime`

```
private static void CheckCalendarZone(ServiceProvider provider)
{
    var zone = "America/Los_Angeles";
    var calendar = new Ical.Net.Calendar();
    var blockDate = new DateTime(2024, 12, 2, 8, 0, 0, DateTimeKind.Utc);
    calendar.Events.Add(new CalendarEvent
    {
        Summary = "Unavailable Before Work",
        DtStart = new CalDateTime(blockDate).ToTimeZone(zone),
        DtEnd = new CalDateTime(blockDate.Add(new TimeSpan(8, 0, 0))).ToTimeZone(zone),
        IsAllDay = false,
        RecurrenceRules = new List<RecurrencePattern>
        {
            new RecurrencePattern
            {
                Frequency = Ical.Net.FrequencyType.Weekly,
                Interval = 1,
                ByDay = new List<WeekDay>
                {
                    new WeekDay(DayOfWeek.Monday)
                }
            }
        }
    });

    var start = new DateTime(2024, 12, 9, 8, 5, 0, DateTimeKind.Utc);
    var end = new DateTime(2024, 12, 10, 7, 59, 59, DateTimeKind.Utc);
    var x = calendar.GetOccurrences(start, end);
    Console.WriteLine(x.Count); // Should be 1
}
```

This time, I get nothing in my occurrences.
The only way I can get something is if my `start` is BEFORE midnight, so if I replace `start` with `var start = new DateTime(2024, 12, 9, 8, 0, 0, DateTimeKind.Utc).AddMilliseconds(-1);`
it starts working.

axunonb:
Updated `CalendarEvent` and `EventEvaluator` to enhance event handling, particularly around time zones and daylight saving time transitions.

Key changes include:

- Enabled nullable reference types in `CalendarEvent` and `EventEvaluator.`
- Updated `CalendarEvent` to throw `InvalidOperationException` when trying to set both `DtEnd` and `Duration`.
- Event end time and duration calculations in `EventEvaluator` reflect any DST transitions.
- Added new test class `RecurrenceTests_From_Issues` to validate changes with various edge cases and scenarios.

Resolves #660
Resolves #623
Resolves #671
Resolves #567
Resolves #342
Resolves #298

### DIFF
--- a/Ical.Net.Tests/CalendarEventTest.cs
+++ b/Ical.Net.Tests/CalendarEventTest.cs
@@ -427,34 +427,37 @@ END:VCALENDAR";
     public void HourMinuteSecondOffsetParsingTest()
     {
         const string ical =
-            @"BEGIN:VCALENDAR
-PRODID:-//1&1 Mail & Media GmbH/GMX Kalender Server 3.10.0//NONSGML//DE
-VERSION:2.0
-CALSCALE:GREGORIAN
-METHOD:REQUEST
-BEGIN:VTIMEZONE
-TZID:Europe/Brussels
-TZURL:http://tzurl.org/zoneinfo/Europe/Brussels
-X-LIC-LOCATION:Europe/Brussels
-BEGIN:DAYLIGHT
-TZOFFSETFROM:-001730
-TZOFFSETTO:-001730
-TZNAME:CEST
-DTSTART:19810329T020000
-RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU
-END:DAYLIGHT
-BEGIN:STANDARD
-TZOFFSETFROM:+001730
-TZOFFSETTO:+001730
-TZNAME:BMT
-DTSTART:18800101T000000
-RDATE:18800101T000000
-END:STANDARD
-END:VTIMEZONE
-END:VCALENDAR";
+            """
+            BEGIN:VCALENDAR
+            PRODID:-//1&1 Mail & Media GmbH/GMX Kalender Server 3.10.0//NONSGML//DE
+            VERSION:2.0
+            CALSCALE:GREGORIAN
+            METHOD:REQUEST
+            BEGIN:VTIMEZONE
+            TZID:Europe/Brussels
+            TZURL:http://tzurl.org/zoneinfo/Europe/Brussels
+            X-LIC-LOCATION:Europe/Brussels
+            BEGIN:DAYLIGHT
+            TZOFFSETFROM:-001730
+            TZOFFSETTO:-001730
+            TZNAME:CEST
+            DTSTART:19810329T020000
+            RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU
+            END:DAYLIGHT
+            BEGIN:STANDARD
+            TZOFFSETFROM:+001730
+            TZOFFSETTO:+001730
+            TZNAME:BMT
+            DTSTART:18800101T000000
+            RDATE:18800101T000000
+            END:STANDARD
+            END:VTIMEZONE
+            END:VCALENDAR
+            """;
         var timezones = Calendar.Load(ical)
             .TimeZones.First()
-            .Children.Cast<CalendarComponent>();
+            .Children.Cast<CalendarComponent>()
+            .ToArray();
 
         var positiveOffset = timezones
             .Skip(1).Take(1).First()
@@ -472,78 +475,99 @@ END:VCALENDAR";
 
 
     [Test, Category("CalendarEvent")]
-    public void TestGetEffectiveDuration()
+    public void GetNominalDurationTests()
     {
-        var now = _now.Subtract(TimeSpan.FromTicks(_now.Ticks % TimeSpan.TicksPerSecond));
+        var dt = new DateTime(2025, 3, 1, 14, 30, 0);
+        const string tzIdStart = "America/New_York";
+        const string tzIdEnd = "Europe/London";
 
-        var evt = new CalendarEvent()
+        var evt = new CalendarEvent
         {
-            DtStart = new CalDateTime(DateOnly.FromDateTime(now), TimeOnly.FromDateTime(now)),
-            DtEnd = new CalDateTime(DateOnly.FromDateTime(now.AddHours(1)), TimeOnly.FromDateTime(now.AddHours(1)))
+            DtStart = new CalDateTime(DateOnly.FromDateTime(dt), TimeOnly.FromDateTime(dt), tzIdStart),
+            DtEnd = new CalDateTime(DateOnly.FromDateTime(dt.AddHours(1)), TimeOnly.FromDateTime(dt.AddHours(1)), tzIdEnd)
         };
 
         Assert.Multiple(() =>
         {
-            Assert.That(evt.DtStart.Value, Is.EqualTo(now));
-            Assert.That(evt.DtEnd.Value, Is.EqualTo(now.AddHours(1)));
-            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
+            Assert.That(evt.DtStart.Value, Is.EqualTo(dt));
+            Assert.That(evt.DtEnd.Value, Is.EqualTo(dt.AddHours(1)));
+            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
         });
 
-        evt = new CalendarEvent()
+        evt = new CalendarEvent
         {
-            DtStart = new CalDateTime(DateOnly.FromDateTime(now.Date), TimeOnly.FromDateTime(now.Date)),
-            DtEnd = new CalDateTime(DateOnly.FromDateTime(now.Date.AddHours(1)), TimeOnly.FromDateTime(now.Date.AddHours(1)))
+            DtStart = new CalDateTime(DateOnly.FromDateTime(dt.Date)),
+            DtEnd = new CalDateTime(DateOnly.FromDateTime(dt.Date))
         };
 
         Assert.Multiple(() =>
         {
-            Assert.That(evt.DtStart.Value, Is.EqualTo(now.Date));
-            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
+            Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
+            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.Zero));
         });
 
-        evt = new CalendarEvent()
+        evt = new CalendarEvent
         {
-            DtStart = new CalDateTime(DateOnly.FromDateTime(now)),
+            DtStart = new CalDateTime(DateOnly.FromDateTime(dt)),
         };
 
         Assert.Multiple(() =>
         {
-            Assert.That(evt.DtStart.Value, Is.EqualTo(now.Date));
-            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromDays(1)));
+            Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
+            Assert.That(evt.Duration, Is.Null);
+            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromDays(1)));
         });
 
-        evt = new CalendarEvent()
+        evt = new CalendarEvent
         {
-            DtStart = new CalDateTime(DateOnly.FromDateTime(now), TimeOnly.FromDateTime(now)),
+            DtStart = new CalDateTime(DateOnly.FromDateTime(dt), TimeOnly.FromDateTime(dt)),
             Duration = TimeSpan.FromHours(2),
         };
 
         Assert.Multiple(() => {
-            Assert.That(evt.DtStart.Value, Is.EqualTo(now));
+            Assert.That(evt.DtStart.Value, Is.EqualTo(dt));
             Assert.That(evt.DtEnd, Is.Null);
-            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromHours(2)));
+            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromHours(2)));
         });
 
         evt = new CalendarEvent()
         {
-            DtStart = new CalDateTime(DateOnly.FromDateTime(now.Date), TimeOnly.FromDateTime(now.Date)),
+            DtStart = new CalDateTime(DateOnly.FromDateTime(dt.Date), TimeOnly.FromDateTime(dt.Date)),
             Duration = TimeSpan.FromHours(2),
         };
 
         Assert.Multiple(() => {
-            Assert.That(evt.DtStart.Value, Is.EqualTo(now.Date));
-            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromHours(2)));
+            Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
+            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromHours(2)));
         });
 
         evt = new CalendarEvent()
         {
-            DtStart = new CalDateTime(DateOnly.FromDateTime(now)),
+            DtStart = new CalDateTime(DateOnly.FromDateTime(dt)),
             Duration = TimeSpan.FromDays(1),
         };
 
         Assert.Multiple(() => {
-            Assert.That(evt.DtStart.Value, Is.EqualTo(now.Date));
-            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromDays(1)));
+            Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
+            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromDays(1)));
+        });
+    }
+
+    [Test]
+    public void EitherEndTime_OrDuraction_CanBeSet()
+    {
+        var evt = new CalendarEvent()
+        {
+            DtStart = new CalDateTime(2025, 10, 11, 12, 13, 14, CalDateTime.UtcTzId)
+        };
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(() => evt.DtEnd = new CalDateTime(2025, 12, 11), Throws.Nothing);
+            Assert.That(() => evt.Duration = TimeSpan.FromDays(1), Throws.InvalidOperationException);
+            Assert.That(() => evt.DtEnd = null, Throws.Nothing);
+            Assert.That(() => evt.Duration = TimeSpan.FromDays(1), Throws.Nothing);
+            Assert.That(() => evt.DtEnd = new CalDateTime(2025, 12, 11), Throws.InvalidOperationException);
         });
     }
 }
--- a/Ical.Net.Tests/CalendarEventTest.cs
+++ b/Ical.Net.Tests/CalendarEventTest.cs
@@ -491,7 +491,7 @@ END:VCALENDAR";
         {
             Assert.That(evt.DtStart.Value, Is.EqualTo(dt));
             Assert.That(evt.DtEnd.Value, Is.EqualTo(dt.AddHours(1)));
-            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
+            Assert.That(evt.GetTimeSpanToAddToPeriodStartTime(), Is.EqualTo(TimeSpan.FromHours(1)));
         });
 
         evt = new CalendarEvent
@@ -503,7 +503,7 @@ END:VCALENDAR";
         Assert.Multiple(() =>
         {
             Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
-            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.Zero));
+            Assert.That(evt.GetTimeSpanToAddToPeriodStartTime(), Is.EqualTo(TimeSpan.Zero));
         });
 
         evt = new CalendarEvent
@@ -515,7 +515,7 @@ END:VCALENDAR";
         {
             Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
             Assert.That(evt.Duration, Is.Null);
-            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromDays(1)));
+            Assert.That(evt.GetTimeSpanToAddToPeriodStartTime(), Is.EqualTo(TimeSpan.FromDays(1)));
         });
 
         evt = new CalendarEvent
@@ -527,7 +527,7 @@ END:VCALENDAR";
         Assert.Multiple(() => {
             Assert.That(evt.DtStart.Value, Is.EqualTo(dt));
             Assert.That(evt.DtEnd, Is.Null);
-            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromHours(2)));
+            Assert.That(evt.GetTimeSpanToAddToPeriodStartTime(), Is.EqualTo(TimeSpan.FromHours(2)));
         });
 
         evt = new CalendarEvent()
@@ -538,7 +538,7 @@ END:VCALENDAR";
 
         Assert.Multiple(() => {
             Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
-            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromHours(2)));
+            Assert.That(evt.GetTimeSpanToAddToPeriodStartTime(), Is.EqualTo(TimeSpan.FromHours(2)));
         });
 
         evt = new CalendarEvent()
@@ -549,7 +549,7 @@ END:VCALENDAR";
 
         Assert.Multiple(() => {
             Assert.That(evt.DtStart.Value, Is.EqualTo(dt.Date));
-            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromDays(1)));
+            Assert.That(evt.GetTimeSpanToAddToPeriodStartTime(), Is.EqualTo(TimeSpan.FromDays(1)));
         });
     }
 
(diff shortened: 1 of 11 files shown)


Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -42,132 +54,136 @@ public class CalendarEvent : RecurringComponent, IAlarmContainer, IComparable<Ca
...
-    /// Otherwise it will return `default(TimeSpan)`.
-    /// Note that for recurring events, the duration of individual occurrences may vary
-    /// if they span a DST change.
+    /// Note: For recurring events, the <b>effective duration</b> of individual occurrences may vary due to DST transitions
+    /// of the given <see cref="RecurringComponent.DtStart"/> and <see cref="CalendarEvent.DtEnd"/> timezones.
     /// </remarks>
-    /// <returns>The effective duration of this event.</returns>
-    public virtual TimeSpan GetFirstDuration()
+    /// <returns>The <b>nominal duration</b> of this event.</returns>
+    public virtual TimeSpan CalcFirstNominalDuration()
```
minichma: Regarding the naming I'd rather go for just 'GetFirst...'. That calculations are done internally doesn't seem to be relevant from an outside perspective.

minichma: As long as we need to decide for *either* nominal *or* exact durations, I suggest going for exact, because if DTSTART and DTEND have different tzids, then there is no reasonable way to calculate a nominal duration. Nominal only makes sense in case of floating time or if both are in the same tz. If we just ignore the tz then the result might be significantly wrong, certainly more than when using exact time in cases where nominal would be correct. This is probably why the RFC specifies that the duration shall be considered exact if specivied via DTSTART/DTEND.

A better solution in a next step would probably be to return the duration in a way that allows the caller understand whether it should be applied as nominal or exact duration.

Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -42,132 +54,136 @@ public class CalendarEvent : RecurringComponent, IAlarmContainer, IComparable<Ca
...
+                3.8.5.3. Recurrence Rule:
+                If the duration of the recurring component is specified with the
+                "DTEND" or "DUE" property, then the same EXACT duration will apply
+                to all the members of the generated recurrence set.
+
+                We use the nominal duration here, because the caller will set the period end time to the
+                same timezone as the event end time, finally leading to an exact duration
+                calculation from the zoned start time to the zoned end time.
+             */
+            return DtEnd.Value - DtStart.Value;
```
minichma, line 134: As mentioned above I suggest to go for the exact duration for now (i.e. `DtEnd.AsUtc - DtStart.AsUtc`). *If* calculating the nominal duration, we should throw if `DtEnd.TzId != DtStart.TzId`.

Ical.Net/Evaluation/EventEvaluator.cs
```diff
@@ -27,37 +31,69 @@ public EventEvaluator(CalendarEvent evt) : base(evt) { }
...
+        {
+            // Calculate the end time of the event as a DateTime
+            var endDt = period.StartTime.Value.Add(nominalDuration);
+
+            // Create a CalDateTime object with the calculated end time.
+            // Ensure date-only or date-time consistency with the start time,
+            // and use the timezone from the event's definition.
+            endTime = new CalDateTime(
+                DateOnly.FromDateTime(endDt),
+                period.StartTime.HasTime ? TimeOnly.FromDateTime(endDt) : null,
```
minichma, line 89: Here we risk confusing the tzid. Simply use `var endDt = period.StartTime.Add(nominalDuration)`? We must make sure that Add() and GetFirst..() both operate on the same kind of duration, nominal vs exact.

Ical.Net/Evaluation/EventEvaluator.cs
```diff
@@ -27,37 +31,69 @@ public EventEvaluator(CalendarEvent evt) : base(evt) { }
...
+                CalendarEvent.End?.TzId ?? CalendarEvent.Start.TzId);
+        }
+
+        // Return the Period object with the calculated end time and duration.
+        period.Duration = endTime.Subtract(period.StartTime); // exact duration
```
minichma, line 94: This might be the right time to clean up `Period` too, i.e. remove the redundant data and the `Extrapolate...` as we did in `CalendarEvent`.

Ical.Net.Tests/CalendarEventTest.cs
```diff
@@ -472,78 +475,99 @@ public void HourMinuteSecondOffsetParsingTest()
...
         };
 
         Assert.Multiple(() =>
         {
-            Assert.That(evt.DtStart.Value, Is.EqualTo(now));
-            Assert.That(evt.DtEnd.Value, Is.EqualTo(now.AddHours(1)));
-            Assert.That(evt.GetFirstDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
+            Assert.That(evt.DtStart.Value, Is.EqualTo(dt));
+            Assert.That(evt.DtEnd.Value, Is.EqualTo(dt.AddHours(1)));
+            Assert.That(evt.CalcFirstNominalDuration(), Is.EqualTo(TimeSpan.FromHours(1)));
```
minichma: It should not be possible to calculate a nominal duration if the tzids don't match as the value usually doesn't make much sense.

Ical.Net/Evaluation/EventEvaluator.cs
minichma: Oh, I guess I missed your point. Of course, you apply the same 'nominal' duration (not sure whether this is the right term if the tz is different) and preserve the tz. So in the end the exact duration is close to the original one (except for the DST changes). Yeah, I guess this is a reasonable approach, but I would still discourage implementing it like that. The RFC requires applying the nominal duration *only* if specified via the `DURATION` property, in which case there is only a single tz. I guess this is due to the ambiguity that comes from comparing times in different zones. Anyhow, to my understanding there's no hint in the RFC of how to apply 'nominal' durations in case of different tz.

I therefore keep my suggestion to calculate in exact durations for now until we precisely implement keeping both apart (in which case we also don't have to deal with this ambiguity). This way we are not fully precise if a duration is specified via the `DURATION` property, but at lease if specified via `DTEND`. In the way it's implemented in the PR we could be off in both cases.

Ical.Net/CalendarComponents/CalendarEvent.cs
minichma: Following up on [my comment in EventEvaluator](https://github.com/ical-org/ical.net/pull/673/files#r1893119377) - Got your point now of how the 'nominal' duration should be used, i.e. by requiring the caller to preserve DTEND's tz. However, this is a very specific case and I think we shouldn't offer this calculation as a built in functionality of CalendarEvent. If we (or anyone) needs this kind of calculation on times with different TZ, it should be calculated in place, exactly as its needed there, but not in such a central place. I'm not sure that 'nominal' is the right terminology if the TZ is different. Even if the terminology could be considered correct, its certainly very confusing and could be a source of coding flaws very easily (always talking just about the case with different tz). I therefore suggest to offer calculating the nominal duration *only* if the TZ is the same (or if the duration is specified via the 'DURATION' property'). Actually the only correct way of calculating the effective duration would be to calculate it in a way that preserves the info of whether it is nominal or exact rather than letting the caller choose whether they want nominal or exact.

This could be achieved in several ways:
1. Return something like `(TimeSpan nominal, TimeSpan exact)` that allows keeping both apart
2. Return something like `(TimeSpan d, bool isNominal)` - less expressive and not able to express a combination of both as required for the `DURATION` property.
3. Introduce a new `Duration` type that stores Days/Weeks and time independently, i.e. closer to how durations are represented in the RFC (`P1W2DT3H`). The RFC requires the days to be considered nominal while the time shall be considered exact. As we are free to return a time part > 24h, we would be free to represent `DTSTART - DTEND` just in the time part, which would imply its exact. This would require the Duration property of `CalendarEvent` to be represented as `Duration` rather than as `TimeSpan` either, which might be a bigger change.

I feel no 3 is probably the cleanest one while 1 is the most straight forward one.

minichma: If we improve the return value as suggested below, i.e. that it can represent the duration in a way that allows keeping nominal/exact apart, then we even wouldn't be talking about the 'first' duration anymore, so then a proper name could be like `GetEffectiveDuration()`. Like

`Duration GetEffectiveDuration()`

In `CalDateTime` we could then offer something like `Add(Duration d)` taking care of the difference between both parts.

axunonb:
Thanks for the review and valuable comments.

#### Timezone-aware occurrences
I agree that the solution for determining the end time of the period is unwieldy or unintuitive.
In fact, I have considered introducing a new duration class as mentioned in a previous discussion. However, this would have required adjustments that go well beyond the purpose of a bugfix PR.
The `CalendarEvent.CalcFirstNominalDuration` method is now renamed to `CalendarEvent.GetTimeSpanToAddToPeriodStartTime`, which hopefully is less ambiguous than the term "duration".
How about merging this bugfix PR, because that way at least 6 of the open issues are closed.

#### Duration
With regard to duration, we could create a new issue #675 now, which would then be processed in a separate PR. 
The new `Duration` could be used in the `CalDateTime` arithmetic, the `Period` class and it would also replace `CalendarEvent.GetTimeSpanToAddToPeriodStartTime`.
Would that be something you could adopt?

#### Refactored `Period` class
- As suggested, the private `ExtrapolateTimes()` method is replaced.
- Shortcoming: `Period.Duration` may still implicitly return a nominal or an exact duration. See above.

Ical.Net/CalendarComponents/CalendarEvent.cs
axunonb: See the [comment](https://github.com/ical-org/ical.net/pull/673#issuecomment-2558053226)

axunonb: I'd prefer option 3, too, but without replacing one intermediate implementation with another one.
See the [comment](https://github.com/ical-org/ical.net/pull/673#issuecomment-2558053226).

Ical.Net/Evaluation/EventEvaluator.cs
axunonb: See [comment](https://github.com/ical-org/ical.net/pull/673#issuecomment-2558053226)

Ical.Net/Evaluation/EventEvaluator.cs
axunonb: Added a commit with Extrapolate... replaced

Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -42,132 +54,137 @@ public class CalendarEvent : RecurringComponent, IAlarmContainer, IComparable<Ca
...
-    /// Otherwise it will return `default(TimeSpan)`.
-    /// Note that for recurring events, the duration of individual occurrences may vary
-    /// if they span a DST change.
+    /// Note: For recurring events, the <b>exact duration</b> of individual occurrences may vary due to DST transitions
+    /// of the given <see cref="RecurringComponent.DtStart"/> and <see cref="CalendarEvent.DtEnd"/> timezones.
     /// </remarks>
-    /// <returns>The effective duration of this event.</returns>
-    public virtual TimeSpan GetFirstDuration()
+    /// <returns>The time span that gets added to the period start time to get the period end time.</returns>
+    public virtual TimeSpan GetTimeSpanToAddToPeriodStartTime()
```
minichma: Should we maybe make this internal for now?

minichma (dismissed):

Ical.Net/CalendarComponents/CalendarEvent.cs
axunonb: oh, yes, as this is only temporary

Issue #676, opened 2024-12-21
Make parameterless CTORs only needed for `ICopyable` internal


axunonb:
Reasoning: Parameterless CTORs are error prone if not explicitly needed in the public API.

PR #680, opened 2024-12-26
Introduce `Duration` type to keep apart nominal vs exact durations.

minichma:
This PR addresses shortcomings of the lib regarding the different requirements of nominal vs exact duration as specified by RFC 5545. It introduces the new `Duration` type that replaces `TimeSpan` in most places. In contrast to the latter, `Duration` stores the individual duration parts (weeks, days, hours, minutes, seconds) individually. This is required because according to the RFC weeks and days are considered nominal values while hours, minutes, seconds are considered exact (see https://www.rfc-editor.org/rfc/rfc5545#section-3.3.6, https://www.rfc-editor.org/rfc/rfc5545#section-3.8.5.3).

Most relevant changes:

* `CalendarEvent.Duration` et al and `Period.Duration` changed from `TimeSpan` to `Duration`
* `CalendarEvent.GetTimeSpanToAddToPeriodStartTime` replaced by `CalendarEvent.GetEffectiveDuration`, which takes care of keeping apart nominal vs. exact durations. I.e. if the duration is specified via DTEND, then the returned duration only uses time-related fields, so it is considered exact. If specified via `DURATION`, the duration is returned as specified by the caller.
* `CalDateTime.Add(Duration)` and `.Subtract()` handle the different parts of `Duration` conformant to the RFC.


axunonb:
I'll be waiting until this PR gets merged before submitting a PR for `EXDATE` and `RDATE` fixes.

minichma:
> I'll be waiting until this PR gets merged before submitting a PR for `EXDATE` and `RDATE` fixes.

@axunonb I think the PR should be ready for review, would be great if you'd find some time for it. It's quite some code that has changed, but all in all not too complex, once the basics regarding nominal vs. exact are understood. Probably it makes sense to review the code commit by commit. I tried to keep the commits reasonably separated, but probably not perfectly everywhere. There are certainly some smaller issues around nonexistent times remaining as outlined in #681.

Ical.Net/Serialization/DataTypes/DurationSerializer.cs
```diff
@@ -0,0 +1,108 @@
...
+
+public class DurationSerializer : SerializerBase
+{
+    public DurationSerializer() { }
+
+    public DurationSerializer(SerializationContext ctx) : base(ctx) { }
+
+    public override Type TargetType => typeof(Duration);
+
+    public override string SerializeToString(object obj)
```
axunonb, line 22: Maybe just
```cs 
public override string SerializeToString(object obj)
    => !(obj is Duration duration) ? null : SerializeToString(duration);
```

Ical.Net/Serialization/DataTypes/DurationSerializer.cs
```diff
@@ -0,0 +1,108 @@
...
+    private static string SerializeToString(Duration ts)
+    {
+        if (ts.IsEmpty)
+            return "P0D";
+
+        var sb = new StringBuilder();
+
+        var sign = ts.Sign;
+        if (sign < 0)
+            sb.Append("-");
```
axunonb: better use StringBuilder.Append(char) instead of string?

Ical.Net/DataTypes/Duration.cs
```diff
@@ -0,0 +1,222 @@
...
+        => ((Weeks ?? 0) == 0)
+            && ((Days ?? 0) == 0)
+            && ((Hours ?? 0) == 0)
+            && ((Minutes ?? 0) == 0)
+            && ((Seconds ?? 0) == 0);
+
+    /// <summary>
+    /// Gets a value indicating whether the duration is empty, that is, all fields are null.
+    /// </summary>
+    internal bool IsEmpty
```
axunonb, line 196: Would `Duration? x = null; if (x == null)` be more clear/simple than than `Duration x = new Duration(); if (x.IsEmpty)`? Then we could disallow empty.

Ical.Net/DataTypes/IDateTime.cs
```diff
@@ -115,9 +115,27 @@ public interface IDateTime : IEncodableDataType, IComparable<IDateTime>, IFormat
...
+
+    /// <summary>
+    /// Add the specified <see cref="Duration"/> to this instance/>.
+    /// </summary>
+    /// <remarks>
+    /// In correspondence to RFC5545, the weeks and day fields of a duration are considered nominal durations while the time fields are considered exact values.
+    /// </remarks>
+    IDateTime Add(Duration d);
+
+    /// <summary>Returns a new <see cref="TimeSpan" /> from subtracting the specified <see cref="IDateTime"/> from to the value of this instance.</summary>
```
axunonb: A xmldoc hint for SubtractExact vs Subtract similar like in static Duration FromTimeSpan(TimeSpan t) might be useful.

"from to the value" - "to" is too much?

Ical.Net/DataTypes/Duration.cs
```diff
@@ -0,0 +1,222 @@
...
+    /// </summary>
+    public static Duration operator-(Duration d) =>
+        new Duration(-d.Weeks, -d.Days, -d.Hours, -d.Minutes, -d.Seconds);
+
+    /// <inheritdoc/>
+    public override string ToString()
+        => new DurationSerializer().SerializeToString(this);
+
+    private static int? GetSign(int? v) =>
+        v switch
```
axunonb, line 214: or
```cs
v switch
{
    null => null,
    >= 0 => 1,
    _ => -1
};
```

axunonb:
The plan for a next PR re `Period`was:
- Change `public Perdiod()` to internal (should not be part of the public API, see #676). Change`Activator.CreateInstance(type, true)` in `CalendarDataType.Copy`.
- Timezone of StartTime and EndTime (if set) must be the same (as needed in the iCalendar context anyway)
That's necessary for EXDATE and RDATE.

Ical.Net/DataTypes/Period.cs
```diff
@@ -138,34 +121,38 @@ public virtual IDateTime? EndTime
...
-    }
+    public virtual Duration? Duration { get; set; }
 
     /// <summary>
     /// Gets the original start time, end time, and duration as they were set,
     /// while <see cref="EndTime"/> or <see cref="Duration"/> properties may be calculated.
     /// </summary>
     /// <returns></returns>
-    public (IDateTime StartTime, IDateTime? EndTime, TimeSpan? Duration) GetOriginalValues()
+    public (IDateTime StartTime, IDateTime? EndTime, Duration? Duration) GetOriginalValues()
```
axunonb: Isn't this redundant now as these properties already represent the original values?

Ical.Net/DataTypes/Period.cs
```diff
@@ -138,34 +121,38 @@ public virtual IDateTime? EndTime
...
     {
-        return (StartTime, _endTime, _duration);
+        return (StartTime, EndTime, Duration);
+    }
+
+    internal Duration GetEffectiveDuration()
```
axunonb, line 126: Like it!

Ical.Net/DataTypes/Period.cs
```diff
@@ -138,34 +121,38 @@ public virtual IDateTime? EndTime
...
+        if (EndTime is { } endTime)
+            return endTime.Subtract(StartTime);
+
+        if (!StartTime.HasTime)
+            return DataTypes.Duration.FromDays(1);
+
+        return DataTypes.Duration.Zero;
+    }
+
+    internal IDateTime GetEffectiveEndTime()
```
axunonb, line 140: Nice, too!

Ical.Net/DataTypes/Period.cs
```diff
@@ -198,17 +186,10 @@ public virtual bool Contains(IDateTime? dt)
...
-        var startContained = Contains(period.StartTime);
-        var endContained = period.EndTime != null && Contains(period.EndTime);
-
-        // Check if the current period is completely within the given period
-        var currentStartContained = period.Contains(StartTime);
-        var currentEndContained = EndTime != null && period.Contains(EndTime);
-
-        return startContained || endContained || currentStartContained || currentEndContained;
-    }
+            => Contains(period.StartTime)
```
axunonb, line 179: much better

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -414,53 +414,91 @@ public DateTime Value
...
 
     /// <inheritdoc/>
-    public IDateTime ToTimeZone(string otherTzId)
-    {
-        if (IsFloating) return new CalDateTime(_dateOnly, _timeOnly, otherTzId);
+    IDateTime IDateTime.ToTimeZone(string otherTzId) => ToTimeZone(otherTzId);
 
-        var zonedOriginal = DateUtil.ToZonedDateTimeLeniently(Value, TzId);
-        var converted = zonedOriginal.WithZone(DateUtil.GetZone(otherTzId));
+    public CalDateTime ToTimeZone(string otherTzId)
```
axunonb: You asked whether `otherTzId` should be `string?`, and the answer was no.
But, well, sometimes it would be convenient to easily change a zoned date/time to Floating this way., WDYT - keep or change?

Ical.Net/CalendarComponents/Todo.cs
```diff
@@ -190,26 +176,4 @@ protected override void OnDeserializing(StreamingContext context)
         //ToDo: a necessary evil, for now
         base.OnDeserializing(context);
     }
-
-    private void ExtrapolateTimes(int source)
```
axunonb, line 194: One more time...

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -414,53 +414,91 @@ public DateTime Value
...
+        AssociateWith(this);
+
+        return newDateTime;
     }
 
-    /// <summary>Returns a new <see cref="TimeSpan" /> from subtracting the specified <see cref="IDateTime"/> from to the value of this instance.</summary>
-    /// <param name="dt"></param>
-    public TimeSpan Subtract(IDateTime dt) => AsUtc - dt.AsUtc;
+    /// <inheritdoc/>
+    public TimeSpan SubtractExact(IDateTime dt) => AsUtc - dt.AsUtc;
```
axunonb, line 492: A xmldoc hint for SubtractExact vs Subtract similar like in static Duration FromTimeSpan(TimeSpan t) might be useful.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -414,53 +414,91 @@ public DateTime Value
...
-    /// and the time span is not a multiple of full days.
-    /// </exception>
-    public IDateTime Subtract(TimeSpan ts) => Add(-ts);
+    /// <inheritdoc/>
+    public Duration Subtract(IDateTime dt)
+    {
+        if (this.TzId is not null)
+            return SubtractExact(dt).ToDurationExact();
+
+        if (dt.HasTime != HasTime)
```
axunonb, line 500: good

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -130,19 +134,18 @@ public void DailyCount1()
...
-                new CalDateTime(2006, 07, 24, 10, 00, 00, _tzid),
-                new CalDateTime(2006, 07, 26, 10, 00, 00, _tzid),
-                new CalDateTime(2006, 07, 28, 10, 00, 00, _tzid),
-                new CalDateTime(2006, 07, 30, 10, 00, 00, _tzid),
-                new CalDateTime(2006, 08, 01, 10, 00, 00, _tzid),
-                new CalDateTime(2006, 08, 03, 10, 00, 00, _tzid),
-                new CalDateTime(2006, 08, 05, 10, 00, 00, _tzid)
-            },
+            [
+                new(new CalDateTime(2006, 07, 18, 10, 00, 00, _tzid), Duration.FromHours(1)),
```
axunonb, line 138: Shouldn't we reduce the size of the CalDateTime lists in RecurrenceTests to the relevant entries (which is much smaller)?

axunonb (commented):
Excellent.
See some minor remarks.
{Edit] Just forgot to mention the nullability warnings like Possible null reference argument for parameter 'dt' in 'Duration IDateTime.Subtract(IDateTime dt)' et al

Ical.Net/DataTypes/Duration.cs
minichma: > Would Duration? x = null; if (x == null) be more clear/simple than than Duration x = new Duration(); if (x.IsEmpty)?

Basically yes, but we don't use `new Duration()` like that anyways. The only place we use it is to initialize `public static Duration Zero { get; } = new Duration();`. Not sure whether it would be desirable to disallow an empty Duration. Would rather say we should allow it. TBD.

Ical.Net/Serialization/DataTypes/DurationSerializer.cs
minichma: Certainly, thanks. This class is based on a copy of the previous `TimeSpanSerializer`,  could easily see some leftovers from there.

Ical.Net/DataTypes/Duration.cs
axunonb: Oh yes, right, then `static Duration Zero { get; } = new Duration();` returns a duration that `IsEmpty`, not `IsZero`?

Ical.Net/DataTypes/Duration.cs
minichma: Nice, haven't been aware of that syntax.

Ical.Net/DataTypes/Period.cs
minichma: yes, absolutely

Ical.Net/DataTypes/CalDateTime.cs
minichma: As TzId `null` represents floating time everywhere throughout the class, it would be logical to allow `null` here. Otherwise we should introduce a method `ToFloating()`, but allowing null here keeps the API more straight forward. So I'd say, `null` should be allowed.

Ical.Net/CalendarComponents/Todo.cs
minichma: No more `Extrapolate` for now :-)

Ical.Net/DataTypes/CalDateTime.cs
minichma: Yeah, I also tried throwing if `IsFloating != dt.IsFloating`, but then all kinds of tests fail. Obviously there are quite some cases where `DTSTART` has a tzid set but `DTEND` hasn't, which isn't allowed by the RFC though. We should double-check whether we still want to support this case (probably yes, if its used in the wild). If so, we should probably deal with the case in a different place, probably during deserialization, and be more strict here, because the current implementation is somewhat unclear regarding that case.

Ical.Net.Tests/RecurrenceTests.cs
minichma: Don't know enough about the individual tests to comment on that, but sounds like quite some work. :-) Certainly not in scope of this PR though.

Ical.Net/DataTypes/Duration.cs
minichma: Actually it is both, empty and zero.

I.e. `new Duration()` is both, empty and zero,
`new Duration(0)` is zero but not empty.

Both are `internal`, as I'm aware that it could be confusing.

Ical.Net/DataTypes/CalDateTime.cs
minichma: Added support for null as I'm on it anyways.

minichma:
@axunonb Thanks for the review! Tried to improve.

axunonb (approved):
Great job, thanks a lot

PR #684, opened 2024-12-29
Add abstraction layer to create `EXDATE` and `RDATE` properties

### LINKED ISSUE #590: Exdate not taken correctly into account when getting the FreeBusyStatus?

### LINKED ISSUE #591: Serialized EXDATE is unpermitted PERIOD value type when originated as System.DateTime

### LINKED ISSUE #614: EXDATE should serialize TZID

### LINKED ISSUE #676: Make parameterless CTORs only needed for `ICopyable` internal

axunonb:
### Add abstraction layer to create `EXDATE` and `RDATE` properties
- Introduce `ExceptionDate` and `RecurrenceDates` classes as wrappers for `IList<PeriodList>` properties.
- `Period`s and `CalDateTime`s to them can be done without having to care for RFC 5545 rules regarding serialization:. This happens behind the scenes.

### IRecurrable
- Change the type of `ExceptionDates` from `IList<PeriodList>` to `ExceptionDates`.
- Change the type of `RecurrenceDates` from `IList<PeriodList>` to `RecurrenceDates`.

### RecurringComponent (inhereted by CalendarEvent and other classes), VTimeZoneInfo
- Rename `IList<PeriodList> ExceptionDates` to `ExceptionDatesPeriodLists`. This serves as the backing storage for the new `ExceptionDates` class/property.
- Rename `IList<PeriodList> RecurrenceDates` to `RecurrenceDatesPeriodLists`. This serves as the backing storage for the new `RecurrenceDates` class/property.
- Add the creation of instances of `ExceptionDates` and `RecurrenceDates` to the `Initialize()` method.

### RecurringEvaluator and TodoEvaluator
- Implement the new `ExceptionDates` and `RecurrenceDates` classes.

### VTimeZone
- Enhanced serialization:
  - Before: Each `RDATE` date/time was serialized as a separate `RDATE` propery
  - Now: `RDATE` date/times which can be clustered go into one single `RDATE` property

### Period
- Make parameterless CTOR `internal` to ensure proper initialization by users
- Add internal enum `PeriodKind` to identify the period kind as `Undefined`, `DateTime`, `DateOnly` or `Period` and property `PeriodKind`
- `StartTime`, `EndTime` and `Duration` are immutable
- Timezones and date-only or date-time of `StartTime`and (optional) `EndTime` must be the same
- `CompareTo` uses `AsUtc` for comparing the `StartTime`
- Remove returning a "magic" duration of 1 day, if `EndTime` is null and `StartTime` is date-only. This broke `EXDATE` and `RDATE` of an event, when only a start time exists. Full day is handled by `PeriodSerializer`.
- Added `EffectiveEndTime` and `EffectiveDuration` properties to provide calculated values based on the set values.

### PeriodList
- Change access modifier to `internal`.
- Add `internal PeriodKind PeriodKind`
- Add `internal string? TzId` , returning the TzId of the first period
- `EnsureConsistentTimezoneAndPeriodKind`: The first period determines the `TzId` and `PeriodKind` of the `PeriodList` and all other `Period`s added or inserted must have these properties as well.
- Add `SetService(new PeriodListEvaluator(this))` for `StringReader` CTOR overload
- Add `Add(IDateTime)`
- Remove `public` method `GetGroupedPeriods`
- nullable enable

### PeriodListEvaluator
- Change access modifier to `internal`

### EventEvaluator:
- Update to handle `RDATE` durations correctly

### TodoEvaluator:
- Refactor using `ExceptionDate` and `RecurrenceDates` classes
- Remove method `PeriodWithDuration(Period)` as it became redudant with the refactored `Period` class.

### RecurringEvaluator:
- Refactor using `ExceptionDate` and `RecurrenceDates` classes

### DataTypeSerializer and other serializers, CalendarObjectBase:
- `Activator.CreateInstance(TargetType, true)` allows for not `public` CTORs, so that parameterless CTORs can be excluded from the public API, if proper initialization can't be assured (like with `Period`).

### PeriodListSerializer, PeriodSerializer, DateTimeSerializer, PropertySerializer
- Update serializers to handle the modified `PeriodList` implementation.
- Never set the UTC timezone ID (use appending 'Z') (Section 3.2.19: The "TZID" property parameter MUST NOT be applied to DATE properties and DATE-TIME or TIME properties whose time values are specified in UTC.)
- `RDATE` and `EXDATE` serialize following RFC 5545 sections 3.8.5.2 and 3.8.5.1

Resolves #590
Resolves #591
Resolves #614
Resolves #676


axunonb:
@minichma The code changes became more than expected, because it was necessary to revisit `Period` and `PeriodList`.
I feel like EXDATE and RDATE should have more unit tests with different scenarios than those in `RecurrenceWithRDateTests` and `RecurrenceWithExDateTests`. If so, let me know.
Using the `List<PeriodList>` for `EXDATE` brings a disadvantage: Adding a `Period` with `Duration` will not throw. But there is a "how to" in the xmldoc, and we have specific methods to add an `IDateIme` to a `PeriodList`. This way we don't need extra classes just for `EXDATE`. Would this be acceptable?

minichma:
> Using the `List<PeriodList>` for `EXDATE` brings a disadvantage: Adding a `Period` with `Duration` will not throw. But there is a "how to" in the xmldoc, and we have specific methods to add an `IDateIme` to a `PeriodList`. This way we don't need extra classes just for `EXDATE`. Would this be acceptable?

I'm pretty sure it is. I see the challenges that come with introducing separate types for lists of Periods and IDateTimes - redundant code, challenges with type safety, inheritance. At the same time I remember the challenges I had when starting with ical.net and not knowing much about the details of RFC 5545. The ExDate property being a list of Periods was not exactly easy to understand at that time. So yes, doc comments will certainly be helpful in this respect. We can still continue thinking about alternatives in the future (maybe not v5 though once it has been released).

axunonb:
Oops, just noticed that
```RDATE:20231002/P1D,20231003T100000```
is currently possible with `PeriodList`, but not allowed. The VALUE type of containing elements must be the same - as with timezone. Will fix that.

**Comment by axunonb** [comment]
> The ExDate property being a list of Periods was not exactly easy to understand

Creating `EXDATE` and `RDATE` can indeed be a pain for users. Now there is an abstraction layer that makes this a breeze.

**Comment by minichma** [comment]
@axunonb Wish you a great 2025! Hope I will find some time for a review this afternoon when I'll be on a train or otherwise tomorrow.

**Comment by axunonb** [comment]
@minichma Happy New Year and thank you!
Just noticed that as of section 3.8.5.1 for day-only `EXDATE;VALUE=DATE:` must be serialized...
We could start the new year with a first v5 pre release, couldn't we?

File: Ical.Net/DataTypes/Period.cs [hunk 1]
```diff
@@ -98,51 +115,104 @@ public override string ToString()
...
     /// </summary>
-    public virtual Duration? Duration { get; set; }
+    public virtual Duration? EffectiveDuration => _duration ?? (_endTime != null ? GetEffectiveDuration() : null);
 
-    internal Duration GetEffectiveDuration()
+    private static void EnsureConsistentTimezones(IDateTime start, IDateTime? end)
     {
-        if (Duration is { } d)
-            return d;
+        if (end?.TzId != null && start.TzId != end.TzId) throw new ArgumentException($"Start time ({start}) and end time ({end}) must have the same timezone.");
```
minichma: Here we allow end being floating while start isn't. Should we allow that? I feel we shouldn't because this way we create additional cases that need to be tested and therefore increase overall complexity. I.e. either require the date to be floating, meaning that TzId equals that of Start or require TzId to be set. I recommend the latter (require `left.IsFloating == right.IsFloating`).

Ical.Net/DataTypes/Period.cs
```diff
@@ -98,51 +115,104 @@ public override string ToString()
...
+    public virtual IDateTime? EndTime
+    {
+        get => _endTime;
+        set
+        {
+            EnsureConsistentTimezones(_startTime, value);
+            _endTime = value;
+            if (_endTime != null)
+            {
+                _duration = null;
```
minichma: I think we shouldn't allow setting the end time if a duration has already been set. Ideally the whole class would be immutable, but for now I think we should at least not allow changing the structure/kind of an object.

Ical.Net/DataTypes/Period.cs
```diff
@@ -98,51 +115,104 @@ public override string ToString()
...
+    /// </summary>
+    public virtual Duration? Duration
+    {
+        get => _duration;
+        set
+        {
+            _duration = value;
+            if (_duration != null)
+            {
+                _endTime = null;
```
minichma: See above

> I think we shouldn't allow setting the end time if a duration has already been set. Ideally the whole class would be immutable, but for now I think we should at least not allow changing the structure/kind of an object.

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -48,16 +73,18 @@ public override void CopyFrom(ICopyable obj)
...
+    /// Gets the string representation of the list.
+    /// </summary>
+    /// <returns></returns>
+    public override string? ToString() => new PeriodListSerializer().SerializeToString(this);
 
-    public static Dictionary<string, List<Period>> GetGroupedPeriods(IList<PeriodList> periodLists)
+    /// <summary>
+    /// Used for equality comparison of two lists of periods.
+    /// </summary>
+    public static Dictionary<string, IList<Period>> GetGroupedPeriods(IList<PeriodList> periodLists)
```
minichma: If I remember correctly, it should be possible to remove this one. Could be a good time now as updating it is not too easy. Should probably be at least made internal.

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -3201,10 +3201,10 @@ public void OccurrenceMustBeCompletelyContainedWithinSearchRange()
     /// Evaluate relevancy and validity of the request.
     /// Find a solution for issue #120 or close forever
     /// </summary>
-    [Test, Ignore("No solution for issue #120 yet", Until = "2024-12-31")]
+    [Test, Ignore("No solution for issue #120 yet", Until = "2025-02-28")]
```
axunonb: As also mentioned in #685, this test is pending since 2016 and should be removed.

Ical.Net/Calendar.cs
```diff
@@ -275,20 +275,14 @@ public virtual IEnumerable<Occurrence> GetOccurrences<T>(IDateTime startTime = n
...
         {
-            this.AddChild(obj);
-            return (T) obj;
+            this.AddChild(cal);
+            return (T) cal;
         }
         return default(T);
     }
 
-    public void Dispose()
```
axunonb, line 287: Not in use and and implementing IDisposible

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -15,26 +16,50 @@
...
+
+        SetService(new PeriodListEvaluator(this));
     }
 
+    /// <summary>
+    /// Creates a new instance of the <see cref="PeriodList"/> class from the <see cref="StringReader"/> object.
+    /// </summary>
+    /// <param name="value"></param>
+    /// <exception cref="ArgumentException"></exception>
+    public static PeriodList FromStringReader(StringReader value) => new PeriodList(value);
```
minichma, line 64: Do we expect this to be commonly used? Maybe internal?

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -15,26 +16,50 @@
...
         SetService(new PeriodListEvaluator(this));
     }
 
-    public PeriodList(string value) : this()
+    /// <summary>
+    /// Creates a new instance of the <see cref="PeriodList"/> class from the <see cref="StringReader"/>.
+    /// </summary>
+    /// <param name="value"></param>
+    /// <exception cref="ArgumentException"></exception>
+    private PeriodList(StringReader value)
```
minichma, line 48: Do we expect this to be commonly used? Maybe internal?

Ical.Net/DataTypes/Period.cs
```diff
@@ -98,51 +115,104 @@ public override string ToString()
...
+    internal PeriodKind GetPeriodKind()
+    {
+        if (EffectiveDuration != null)
+        {
+            return PeriodKind.Period;
+        }
+        return StartTime.HasTime ? PeriodKind.DateTime : PeriodKind.DateOnly;
+    }
+
+    private Duration GetEffectiveDuration()
```
minichma: Shouldn't this be replaced by `EffectiveDuration`? The difference in the current implementation is that the property can return null while the method can't. However, maybe it's easier to deal with that difference at the call site.

Ical.Net/DataTypes/Period.cs
```diff
@@ -98,51 +115,104 @@ public override string ToString()
...
+            return d;
+
+        if (_endTime is { } endTime)
+            return endTime.Subtract(_startTime);
 
         return DataTypes.Duration.Zero;
     }
 
-    internal IDateTime GetEffectiveEndTime()
+    private IDateTime GetEffectiveEndTime()
```
minichma: same as above: Remove in favor of the corresponding property?

Ical.Net/Collections/PeriodCollectionBase.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Ical.Net.DataTypes;
+
+namespace Ical.Net.Collections;
+
+/// <summary>
+/// Represents a base class for collections of periods.
+/// </summary>
+public abstract class PeriodCollectionBase : ICollection<Period>
```
minichma: If we have a `PeriodCollectionBase` shouldn't it be the base class also of `PeriodList`?

Thinking from a different side - the class basically is an implementation of ICollection but doesn't do any additional validation of the entries. Would it make sense to replace the whole class, just like `ExceptionDateCollection` with generic `List<IDateTime>` or the like and implement the specific functionality in extension methods?

Ical.Net/Collections/ExceptionDateCollection.cs
```diff
@@ -0,0 +1,66 @@
...
+/// This class is used to manage dates that should be excluded from recurring events.
+/// </para>
+/// <para>
+/// The main feature of this class is the <see cref="ToExceptionDates"/> method, which aggregates
+/// and converts the exception <see cref="CalDateTime"/> objects into a list of <see cref="PeriodList"/> objects.
+/// This method ensures that the periods are grouped by their timezone IDs and period kinds, and that
+/// each <see cref="PeriodList"/> contains only distinct periods.
+/// </para>
+/// </summary>
+public class ExceptionDateCollection : PeriodCollectionBase
```
minichma: Wouldn't it be possible (and cleaner) letting the `ExceptionDateCollection` be a collection of type `IDateTime` rather than `Period`? In case of `RDATE` we need to deal with the different type of entries but in case of EXDATEs this isn't necessary, so it would probably be cleaner to have it strongly typed. Also see my comment in `PeriodCollectionBase`, maybe replace the whole class with just `List<IDateTime>` and implement the specific functionality as extension methods?

minichma: Oh, just realizing that the class is only used in test cases. Shouldn't we move this to the test projects then?

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -73,57 +100,116 @@ public static Dictionary<string, List<Period>> GetGroupedPeriods(IList<PeriodLis
...
+    /// The timezone period kind of the first value added determines the timezone for the whole list.
+    /// <para/>
+    /// Use <see cref="Collections.ExceptionDateCollection"/> and <see cref="Collections.RecurrencePeriodCollection"/>
+    /// to simplify creating 'EXDATE' and 'RDATE' properties.
+    /// </summary>
+    /// <param name="item">The <see cref="Period"/> for an 'RDATE'.</param>
+    /// <exception cref="ArgumentException"></exception>
+    public void Add(Period item)
+    {
+        EnsureConsistentTimezoneAndPeriodKind(item);
```
minichma, line 143: I feel this validation is somewhat hopeless at this time, because there are several other places to verify (Insert, Indexer, CopyTo), but even if we tackle all of them, Period and IDateTime are still mutable, so they could still be changed (another reason why it would be great to have them immuntable).

A question that comes to my mind is whether it would be a viable alternative to allow mixing different kinds altogether, having just a single list in `CalendarEvent` and split into multiple just during serialization? We wouldn't preserve the original structure after a deserialization/serialization cycle, but that should be ok as the semantics would stay unchanged. Would also relieve us from some 'GetGrouped...' kind of code.

Ical.Net/Serialization/DataTypes/PeriodSerializer.cs
```diff
@@ -34,30 +34,36 @@ public override string SerializeToString(object obj)
...
 
-            // Serialize the duration or end time
-            if (p.EndTime != null)
+            // RFC 5545 section 3.6.1:
+            // For cases where a "VEVENT" calendar component
+            // specifies a "DTSTART" property with a DATE value type but no
+            // "DTEND" nor "DURATION" property, the event’s duration is taken to
+            // be one day:
+
+            if (p.EndTime is { HasTime: true })
```
minichma: We need to serialize EndTime also if it is date-only.

minichma (commented):
This is a challenging one! Left a few comments but they might not be fully consistent. Just take what you think makes sense. Unfortunately I don't have the time right now to go into all the detail that would be required.

Ical.Net/DataTypes/Period.cs
axunonb: Absolutely, I had arithmetic in mind here - so be fixed.

Ical.Net/DataTypes/Period.cs
axunonb: Yes, that's what I had in mind as well. When deserializing non-standard ics, this might happen. Disallowing to change the kind: absolutely.

Ical.Net/DataTypes/PeriodList.cs
axunonb: Should not be too common. Will remove.

Ical.Net/DataTypes/Period.cs
axunonb: Yep

Ical.Net/Collections/PeriodCollectionBase.cs
axunonb: Good point. I'll try to make `PeriodList` internal as it's only a vehicle for de/serializing. Will have a look.

Ical.Net/Serialization/DataTypes/PeriodSerializer.cs
axunonb: Fixed already.

axunonb:
@minichma  Looks like the failing tests for .Net Framework go back to GitHub upgrading to `ubuntu 24.04` as "latest", which is missing `mono`. Downgrading to `ubuntu-22.04` might help: https://github.com/microsoft/sbom-tool/issues/855

Ical.Net/DataTypes/Period.cs
minichma: Well, if we want to support non-standard ICS, then we'd have to allow setting both, but how would we deal with that later on? Probably the deserializer should deal with that case and either raise an error or ignore one of both values.

axunonb: @minichma With the last but one commit StartTime, EndTime and Duration became immutable.
If a Period should change a new instance is the way to. In our code, there are only about 3 places (e.g. recurrence) where this happens.
There is also a new helper method
`internal static Period Create(IDateTime start, IDateTime? end = null, Duration? duration = null, ICalendarObject? associatedObject = null)`
that defines the rule we use, when Duration and EndTime exist. Anyway, there can be only one of them.
If time allows, have a look at the refactored Period class.

Ical.Net/DataTypes/PeriodList.cs
axunonb: Will go into another (smaller) PR

axunonb:
@minichma Also a long story has end...
The last step was to migrate usage of `IList<PeriodList>` to `ExceptionDates` and `RecurrenceDates`.
Finally `PeriodList` and `PeriodListEvaluator` have the `internal` access modifier
I'll do my own review for all changes tomorrow. Could you have another look at the PR (at least at most relevant parts like  `ExceptionDates`, `RecurrenceDates`, `Period` and the serializers? I tried to separate migration steps in the latest 4 commits.

minichma:
> Could you have another look at the PR (at least at most relevant parts like ExceptionDates, RecurrenceDates, Period and the serializers? I tried to separate migration steps in the latest 4 commits.

@axunonb Will be happy to have a look. Could still take a few days though because of family obligations and a busy week.

Ical.Net/DataTypes/Period.cs
```diff
@@ -11,47 +11,97 @@ namespace Ical.Net.DataTypes;
...
+        if (end != null && start.HasTime != end.HasTime)
+            throw new ArgumentException(
+                $"Start time ({start}) and end time ({end}) must both have a time or both be date-only.");
+
+        if (end != null && end.LessThan(start))
+            throw new ArgumentException($"End time ({end}) must be greater than start time ({start}).", nameof(end));
 
-        StartTime = start ?? throw new ArgumentNullException(nameof(start));
-        EndTime = end;
+        _startTime = start ?? throw new ArgumentNullException(nameof(start), "Start time cannot be null.");
```
minichma: We'd have seen a NPE if start was null, at least if end was not nulll.

Ical.Net/DataTypes/Period.cs
```diff
@@ -89,60 +139,81 @@ public override int GetHashCode()
...
-
-        if (EndTime is { } endTime)
-            return endTime.Subtract(StartTime);
+        get
+        {
+            return _endTime switch
+            {
+                null when _duration is null => null,
+                { } endTime => endTime,
+                _ => EffectiveDuration is not null
```
minichma: avoid evaluating EffectiveDuration twice?

Ical.Net/DataTypes/Period.cs
```diff
@@ -155,14 +226,14 @@ internal IDateTime GetEffectiveEndTime()
...
+        if (dt == null || !_startTime.LessThanOrEqual(dt))
         {
             return false;
         }
 
-        var endTime = GetEffectiveEndTime();
+        var endTime = EffectiveEndTime;
         // End time is exclusive
-        return endTime == null || endTime.GreaterThan(dt);
+        return endTime?.GreaterThan(dt) ?? false;
```
minichma: This is a functional change, i.e. if EffectiveEndTime is null. Before dt was considered contained and now it isn't. I assume the new version is better, but just to be sure, this was intentional. Could also throw in such a case, because 'Contains' doesn't make much sense in such a case.

Ical.Net/DataTypes/Period.cs
```diff
@@ -155,14 +226,14 @@ internal IDateTime GetEffectiveEndTime()
     public virtual bool Contains(IDateTime? dt)
```
minichma, line 227: This kind of convenience methods could be changed to extension methods (in some future PR)

Ical.Net/DataTypes/Period.cs
```diff
@@ -178,21 +249,22 @@ public virtual bool Contains(IDateTime? dt)
...
+    /// <inheritdoc/>
     public int CompareTo(Period? other)
     {
         if (other == null)
         {
             return 1;
         }
 
-        if (StartTime.Equals(other.StartTime))
+        if (StartTime.AsUtc.Equals(other.StartTime.AsUtc))
```
minichma, line 264: Why do we do a UTC comparison here? I.e. equality is quite a topic in CalDateTime, so why would we use a different comparison here? Do we need the comparison at all? It seems to implement a somewhat special case. I don't think there's a natural order if the duration isn't considered. Maybe remove the comparison altogether and let the caller choose how to compare?

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -3,38 +3,66 @@
...
 
 namespace Ical.Net.DataTypes;
 
 /// <summary>
-/// An iCalendar list of recurring dates (or date exclusions)
+/// An iCalendar list used to represent a list of <see cref="Period"/> objects
+/// for EXDATE and RDATE properties.
 /// </summary>
-public class PeriodList : EncodableDataType, IList<Period>
+internal class PeriodList : EncodableDataType, IList<Period>
```
minichma, line 21: Like that it's internal now.

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -48,82 +76,113 @@ public override void CopyFrom(ICopyable obj)
...
-
-        // String assignments create new instances
-        TzId = list.TzId;
     }
 
-    public override string ToString() => new PeriodListSerializer().SerializeToString(this);
-
-    public void Add(IDateTime dt) => Periods.Add(new Period(dt));
-
-    public static Dictionary<string, List<Period>> GetGroupedPeriods(IList<PeriodList> periodLists)
```
minichma, line 60: Great that GetGroupedPeriods is removed!

Ical.Net/DataTypes/PeriodListWrapperBase.cs
```diff
@@ -0,0 +1,109 @@
...
+        var periodList = GetPeriodList(period);
+
+        if (periodList != null) return periodList;
+
+        periodList = new PeriodList();
+        ListOfPeriodList.Add(periodList);
+        return periodList;
+    }
+
+    private protected PeriodList? GetPeriodList(IDateTime dt)
```
minichma, line 84: I don't think you can get a single list for a date. There could be multiple lists having the same TzId and PeriodKind and a single date could be included in more than one list.

Ical.Net/DataTypes/PeriodListWrapperBase.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// Clears all elements from the list.
+    /// </summary>
+    public void Clear() => ListOfPeriodList.Clear();
+
+    /// <summary>
+    /// Determines whether the list contains the <paramref name="dt"/>.
+    /// </summary>
+    public bool Contains(IDateTime dt)
+    {
+        var periodList = GetPeriodList(dt);
```
minichma, line 44: You get the first list of dt's kind and tzid but not necessarily the one that contains this dt.

Ical.Net/DataTypes/PeriodListWrapperBase.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        return periodList?
+            .FirstOrDefault(period => Equals(period.StartTime, dt)) != null;
+    }
+
+    /// <inheritdoc cref="PeriodList.Remove"/>
+    public bool Remove(IDateTime dt)
+    {
+        var periodList = GetPeriodList(dt);
```
minichma, line 53: You get the first list of dt's kind and tzid but not necessarily the one that contains this dt. And dt could be included more than once.

Ical.Net/DataTypes/PeriodListWrapperBase.cs
```diff
@@ -0,0 +1,109 @@
...
+        return ListOfPeriodList
+            .FirstOrDefault(p =>
+                p.TzId == dt.TzId
+                && p.PeriodKind == (dt.HasTime ? PeriodKind.DateTime : PeriodKind.DateOnly));
+    }
+
+    private protected PeriodList? GetPeriodList(Period period)
+    {
+        // The number of PeriodLists is expected to be small, so a linear search is acceptable.
+        return ListOfPeriodList
```
minichma, line 96: Same as with DateTime, there could be multiple lists holding the same Period.

Ical.Net/DataTypes/PeriodListWrapperBase.cs
```diff
@@ -0,0 +1,109 @@
...
+                && p.PeriodKind == period.PeriodKind
+                && p[0].StartTime.HasTime == period.StartTime.HasTime);
+    }
+
+    /// <summary>
+    /// Gets a flattened and ordered list of all distinct periods with
+    /// <see cref="PeriodKind.Period"/>, <see cref="PeriodKind.DateOnly"/> and <see cref="PeriodKind.DateTime"/>.
+    /// </summary>
+    internal IEnumerable<Period> GetAllPeriodsByKind(params PeriodKind[] periodKinds)
+        => ListOfPeriodList.SelectMany(pl => pl.Where(p => periodKinds.Contains(p.PeriodKind))).OrderedDistinct();
```
minichma: You can use `OrderedDistinct` only if the list it is applied on is ordered. But this would only be the case if the individual lists would be ordered and they would be merged using `OrderedMerge`. But do we need to apply a Distinct at all? If the values are duplicated, why not return them here? The method name suggests that 'all' are returned.

Ical.Net/DataTypes/RecurrenceDates.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        return this;
+    }
+
+    /// <summary>
+    /// Determines whether the list contains the <paramref name="period"/>.
+    /// </summary>
+    public bool Contains(Period period)
+    {
+        var periodList = GetPeriodList(period);
```
minichma, line 80: See above - GetPriodList doesn't necessarily return the right list.

Ical.Net/DataTypes/RecurrenceDates.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        return periodList?
+            .FirstOrDefault(p => Equals(p, period)) != null;
+    }
+
+    /// <inheritdoc cref="PeriodList.Remove"/>
+    public bool Remove(Period period)
+    {
+        var periodList = GetPeriodList(period);
```
minichma, line 89: see above, possibly not the right list

Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -138,15 +138,13 @@ calculation from the zoned start time to the zoned end time.
...
-            // "DTEND" nor "DURATION" property, the event’s duration is taken to
-            // be one day.
-            return DataTypes.Duration.FromDays(1);
-        }
+        // RFC 5545 3.6.1:
+        // For cases where a "VEVENT" calendar component
+        // specifies a "DTSTART" property with a DATE value type but no
+        // "DTEND" nor "DURATION" property, the event’s duration is taken to
+        // be one day.
+        // This is taken care of in the PeriodSerializer class,
```
minichma: Don't think we should remove this as the Period only comes into play during evaluation, but the CalendarEvent's duration could be queried independently. Either we remove the method altogether or we return a precise value. We could consolidate the calculation though. I would also not consider this a 'magic number' though. It's just the value that's specified in the RFC.

We have the same kind of logic in multiple places now, so we could consolidate it in a utility class.

Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -287,29 +285,18 @@ protected bool Equals(CalendarEvent? other)
...
-        if (exDates.Any(exDate => !exDate.Value.OrderBy(d => d).SequenceEqual(otherExDates[exDate.Key].OrderBy(d => d))))
-        {
-            return false;
-        }
-
-        var rDates = PeriodList.GetGroupedPeriods(RecurrenceDates);
-        var otherRDates = PeriodList.GetGroupedPeriods(other.RecurrenceDates);
-        if (rDates.Keys.Count != otherRDates.Keys.Count || !rDates.Keys.OrderBy(k => k).SequenceEqual(otherRDates.Keys.OrderBy(k => k)))
+        // exDates and otherExDates are filled with a sorted list of distinct periods
+        var exDates = ExceptionDates.GetAllPeriodsByKind(PeriodKind.Period, PeriodKind.DateOnly, PeriodKind.DateTime).ToList();
```
minichma, line 291: Do we need an implementation of Equals at all? Not sure.

minichma (commented):
@axunonb Had a rather brief look to the most relevant portions and left a few comments again, maybe some duplicated. Please see, what you can use.

Ical.Net/DataTypes/Period.cs
axunonb: Removed

Ical.Net/DataTypes/Period.cs
axunonb: Changed

Ical.Net/DataTypes/Period.cs
axunonb: Must be `return endTime?.GreaterThan(dt) != false;`

Ical.Net/DataTypes/Period.cs
axunonb: agree

Ical.Net/DataTypes/Period.cs
axunonb: Oh yes, you remember the discussion about `CalDateTime` equality.
I opened a [discussion](https://github.com/ical-org/ical.net/discussions/606#discussioncomment-11867080) for that topic.
We can't use Exact Equivalence (same `IsFloating` and same `Value` and same `HasTime` and same `TzId`) here, because the timezone of 2 Periods may be different. and`.AsUtc` is a must.

Isn't the `CompareTo` method specific to each component, that the all callers should "just use".

Ical.Net/DataTypes/PeriodListWrapperBase.cs
axunonb: Yes - RTD., thanks. When attempting to add a duplicate, it just gets discarded. For consistency, values from deserialization should also be unique.

Ical.Net/CalendarComponents/CalendarEvent.cs
axunonb: Valid point - reverted

Ical.Net/CalendarComponents/CalendarEvent.cs
axunonb: I guess, we don't for `CalendarEvent`. Should go into another PR.

axunonb:
@minichma Thanks for having another look at the PR. As always: very helpful. See the comments for details.

minichma:
@axunonb Your are welcome :)

PR #689, opened 2025-01-08
Set Git line endings to autocrlf

axunonb:
Necessary for using C# raw strings

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -21,6 +21,8 @@ jobs:
       uses: actions/checkout@v4
       with:
         fetch-depth: 0  # Fetch all history for all tags and branches
+    - name: Set Git config for line endings
+      run: git config --global core.autocrlf true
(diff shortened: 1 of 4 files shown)


axunonb:
@minichma  This setting that I use with AppVeyor in another project still does not work with raw strings on GitHub.
For now we have to use regular strings, which are not so nice to read. I'll have a deeper look later.

axunonb:
The line endings issue only pops up when comparing the ical serialization result with a C# Raw String. This is because line ending is defined as CRLF for iCalendar. CRLF is the default for *.cs files on Windows, but LF on Linux. We don't use this too often, so I think we can live with using regular strings like this one in such cases:
```c#
Does.Contain(
    "EXDATE;VALUE=DATE:20250101,20250102\r\n" +
    "EXDATE:20250102T140000Z\r\n" +
    "EXDATE;TZID=Europe/Berlin:20250101T101112,20250101T101113"));
```

PR #695, opened 2025-01-18
Improve line folding and serialization performance

### LINKED ISSUE #693: Error in line folding algorithm
According to [Section 3.1 of RFC-5545](https://datatracker.ietf.org/doc/html/rfc5545#section-3.1) line:  
>> Lines of text SHOULD NOT be longer than 75 **octets**, excluding the line break.

Octet is not a character in `C#` cause of `C# char` is 2 byte (16 bit = 2 octets) length, but 1 octet is 1 byte.  
So, you must firstly get the byte representation of a string, say with utf8 encoder, after that it's allowed to do the folding.  
You can also pick up another strategy except one you choose now.

axunonb:
- Update line folding to use StringBuilder and ArrayPool<char> for better performance and memory efficiency.
- Ensure compliance with RFC 5545 by handling multi-byte characters correctly when folding lines
- Add System.Buffers package for netstandard2.0 target framework.

Resolves #693

### DIFF
--- a/Ical.Net/Serialization/PropertySerializer.cs
+++ b/Ical.Net/Serialization/PropertySerializer.cs
@@ -39,65 +39,83 @@ public class PropertySerializer : SerializerBase
         var result = new StringBuilder();
         foreach (var v in prop.Values.Where(value => value != null))
         {
-            // Get a serializer to serialize the property's value.
-            // If we can't serialize the property's value, the next step is worthless anyway.
-            var valueSerializer = sf.Build(v.GetType(), SerializationContext) as IStringSerializer;
-
-            // Iterate through each value to be serialized,
-            // and give it a property (with parameters).
-            // FIXME: this isn't always the way this is accomplished.
-            // Multiple values can often be serialized within the
-            // same property.  How should we fix this?
-
-            // NOTE:
-            // We Serialize the property's value first, as during
-            // serialization it may modify our parameters.
-            // FIXME: the "parameter modification" operation should
-            // be separated from serialization. Perhaps something
-            // like PreSerialize(), etc.
-            var value = valueSerializer?.SerializeToString(v);
-
-            // Get the list of parameters we'll be serializing
-            var parameterList = prop.Parameters;
-            if (v is ICalendarDataType)
-            {
-                parameterList = ((ICalendarDataType) v).Parameters;
-            }
-
-            // The TZID property of an RDATE/EXDATE collection is owned by the PeriodList that contains it. 
-            // It is allowed to have multiple EXDATE or RDATE collections, each with a different TZID.
-            // Using RecurrenceDate and ExceptionDate classes ensures, that all Periods in the
-            // PeriodList have the same TZID and PeriodKind, which allows PeriodList be serialized in one go.
-            // Here, to determine the timezone, we can safely use the first Period's timezone.
-            if (v is PeriodList periodList && periodList[0].TzId != null && periodList[0].TzId != "UTC" &&
-                parameterList.All(p => string.Equals("TZID", p.Value, StringComparison.OrdinalIgnoreCase)))
-            {
-                parameterList.Set("TZID", periodList[0].TzId);
-            }
-
-            var sb = new StringBuilder();
-            sb.Append(prop.Name);
-            if (parameterList.Any())
-            {
-                // Get a serializer for parameters
-                var parameterSerializer = sf.Build(typeof(CalendarParameter), SerializationContext) as IStringSerializer;
-                if (parameterSerializer != null)
-                {
-                    // Serialize each parameter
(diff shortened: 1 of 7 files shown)


Ical.Net/Serialization/PropertySerializer.cs
```diff
@@ -39,65 +39,83 @@ public PropertySerializer(SerializationContext ctx) : base(ctx) { }
         var result = new StringBuilder();
         foreach (var v in prop.Values.Where(value => value != null))
         {
```
axunonb, line 41: Reducing cognitive complexity by splitting up this method

Ical.Net/Serialization/PropertySerializer.cs
```diff
@@ -39,65 +39,83 @@ public PropertySerializer(SerializationContext ctx) : base(ctx) { }
...
-        return result.ToString();
+        result.FoldLines(sb.ToString());
+    }
+
+    private static IParameterCollection GetParameterList(ICalendarProperty prop, object value)
+        => value is ICalendarDataType dataType ? dataType.Parameters : prop.Parameters;
+
+    private static void UpdateTzId(IParameterCollection parameterList, PeriodList periodList)
+    {
+        if (periodList[0].TzId != null && periodList[0].TzId != "UTC" &&
```
minichma: Just remember that PeriodList has a `.Parameters` property, that can have a `TZID` property of its own. Do we deal with/validate that in any way?

Ical.Net/Serialization/PropertySerializer.cs
```diff
@@ -39,65 +39,83 @@ public PropertySerializer(SerializationContext ctx) : base(ctx) { }
...
+            parameterList.All(p => string.Equals("TZID", p.Value, StringComparison.OrdinalIgnoreCase)))
```
minichma: Shouldn't htis be `.Any()`? But even then, why would we set the TZID only if it already exists? Shouldn't we set it in any case except if floating or UTC?

Ical.Net/Serialization/PropertySerializer.cs
```diff
@@ -39,65 +39,83 @@ public PropertySerializer(SerializationContext ctx) : base(ctx) { }
...
+        {
+            parameterList.Set("TZID", periodList[0].TzId);
```
minichma: A serializer is usually expected not to modify the values it operates on.

minichma (commented):
Basically LGTM. One finding would be that I think it would be desirable for a serializer to treat the values it operates on as immutable but we modify the TZID property of the period list. Not sure we can change that. I know that's not a new pattern though.

axunonb:
Excellent, thanks for pointing this out. 
The mentioned code became fully redundant with #684: All timezones of a `PeriodList` are always the same, and there's nothing left to care for in `PropertySerializer` when serializing 👍
https://github.com/ical-org/ical.net/blob/366b7056a9276167be0105e537d7bc9ff0c52c88/Ical.Net/Serialization/DataTypes/PeriodListSerializer.cs#L42-L46

minichma:
> The mentioned code became fully redundant with #684: All timezones of a `PeriodList` are always the same, and there's nothing left to care for in `PropertySerializer` when serializing 👍

Wonderful! It would be good at some point to get rid of the redundant data we have in this regard now. I.e. we have the `Properties` collection which can hold a `TZID` proprety, as well as the explicit TzId fields in `Period.StartDate`.

Ical.Net/Utility/TextUtil.cs
```diff
@@ -4,32 +4,73 @@
...
-            return trimmed + SerializationConstants.LineBreak;
+            result.Append(inputLine);
+            result.Append(SerializationConstants.LineBreak);
+            return;
         }
 
-        const int takeLimit = 74;
+        // Use ArrayPool<char> to rent arrays for processing
+        // Cannot use Span<char> and stackalloc because netstandard2.0 does not support it
+        var arrayPool = ArrayPool<char>.Shared;
```
minichma: I see the performance optimization the usage of the ArrayPool could bring but I would usually try to avoid using a global pool here. It increases complexity and could have unexpected side effects. Had similar situations (in other languages though) where some memory sanitizer suddenly complained that the globally allocated memory changed after calling a certain method just because of some global cache. This might not be the case here but given the quite small effect we can expect from the optimization (compared to just allocating a new array) and given the fact that we don't know how/where users of the lib will will use it, I would rather avoid it.

Question is, whether its needed at all. Why not directly append to the result?

minichma (commented):
LGTM, just another very small comment

minichma (dismissed):

Ical.Net/Utility/TextUtil.cs
axunonb: KISS, replaced it

PR #698, opened 2025-01-18
GHITHUB_TOKEN write access

axunonb:


axunonb:
@minichma  Finally the [daily builds](https://github.com/ical-org/ical.net/pkgs/nuget/Ical.Net) on demand work.
The `GITHUB_TOKEN` had read access on the organization level, which cannot be overridden for the repo. Changed access to write for ical-org.

PR #705, opened 2025-01-22
Remove interface `IDateTime` and replace any use with `CalDateTime`

minichma:
The changes in this PR are in based on the discussion in https://github.com/ical-org/ical.net/pull/705#issuecomment-2629392007, where we discussed whether arithmetics and timezone-related operations on should be moved out of `CalDateTime` in favor of a more flexible approach where we could even configure different timezone databases in a future iteration. We decided against it for reasons of available resources, so arithmetics are kept as part of `CalDateTime`. Based on that the PR implements several improvements, most notably:

* Remove the interface `IDateTime` in favor of using `CalDateTime` directly.
* Simplify the type hierarchy of `CalDateTime`, i.e. stop it inheriting from `EncodableDataType`
* Get rid of `CalDateTime.AssociatedObject`


Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -23,7 +23,7 @@ namespace Ical.Net.DataTypes;
 /// This is because RFC 5545, Section 3.3.5, does not allow for fractional seconds.
 /// </remarks>
 /// </summary>
-public sealed class CalDateTime : EncodableDataType, IComparable<CalDateTime>, IFormattable
+public sealed class CalDateTime : IComparable<CalDateTime>, IFormattable
```
axunonb, line 26: ... or even struct? (not in this PR probably)

minichma: Yes, struct would certainly be good. But we'd have to deal with all the nullability stuff first, which is not perfectly straight forward. Next step could then be to do all that also for `Period`, `PeriodList` and maybe others.

Ical.Net/Utility/DateUtil.cs
```diff
@@ -15,7 +15,7 @@ namespace Ical.Net.Utility;
 
 internal static class DateUtil
 {
-    public static DateTime GetSimpleDateTimeData(IDateTime dt)
+    public static DateTime GetSimpleDateTimeData(CalDateTime dt)
```
axunonb: Should we remove this method? Using `dt.Value` is shorter and more clear

axunonb:
This is PR is still in progress, right? Just asking because I'm assigned to review.

minichma:
Yes, it is still work in progress. I'm not quiet convinced, this is the right approach. Will try to comment more this weekend.

minichma:
@axunonb I apologize for the delay, don't find sufficient time recently. I continued the discussion in https://github.com/ical-org/ical.net/discussions/606#discussioncomment-12031951, as this is not really specific to this PR. Please let me know what you think!

minichma:
Based on the discussion in https://github.com/ical-org/ical.net/pull/705#issuecomment-2629392007 we decided to proceed, so the PR should be ready for review.

axunonb:
`public CalDateTime() { }` should be internal or even private - nothing for the public API

Ical.Net.Tests/DeserializationTests.cs
```diff
@@ -107,7 +107,7 @@ public void Bug2033495()
     }
 
     /// <summary>
-    /// Tests bug #2938007 - involving the HasTime property in IDateTime values.
+    /// Tests bug #2938007 - involving the HasTime property in CalDateTime values.
```
axunonb, line 110: Summary could be removed?

Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -54,9 +54,9 @@ private void SetProperty<T>(string group, T value)
     /// will be extrapolated.
     /// </note>
     /// </summary>
-    public virtual IDateTime? DtEnd
+    public virtual CalDateTime? DtEnd
```
axunonb, line 57: Should we now remove the aliases End => DtEnd and Start => DtStart and just use the more intuitive **End** and **Start** everywhere?
(The underlying Properties could stay unchanged)

Ical.Net/CalendarComponents/IUniqueComponent.cs
```diff
@@ -15,7 +15,7 @@ public interface IUniqueComponent : ICalendarComponent
 
     IList<Attendee> Attendees { get; set; }
     IList<string> Comments { get; set; }
-    IDateTime DtStamp { get; set; }
+    CalDateTime DtStamp { get; set; }
```
axunonb, line 18: or just **Stamp**?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -369,38 +373,60 @@ public DateTime Value
     /// </summary>
     public string? TzId => _tzId;
 
-    /// <inheritdoc cref="TzId"/>
+    /// <summary>
+    /// Gets the timezone name this time is in, if it references a timezone.
+    /// </summary>
     /// <remarks>This is an alias for <see cref="TzId"/></remarks>
     public string? TimeZoneName => TzId;
```
axunonb, line 380: Remove the alias?

Ical.Net/Utility/DateUtil.cs
```diff
@@ -16,9 +16,6 @@ namespace Ical.Net.Utility;
 
 internal static class DateUtil
 {
-    public static DateTime GetSimpleDateTimeData(IDateTime dt)
```
axunonb, line 19: 👍

Ical.Net/VTimeZoneInfo.cs
```diff
@@ -133,15 +133,15 @@ public virtual IList<string> TimeZoneNames
         set => Properties.Set("TZNAME", value);
     }
 
-    public virtual IDateTime DtStart
+    public virtual CalDateTime DtStart
```
axunonb, line 136: There are many of these aliases, so maybe rather address remove the the Dt... properties in a separate PR

axunonb (approved):
Very nice. See few minor comments.

PR #706, opened 2025-01-25
Enable NRT for classes in namespace Ical.Net.Utility 

axunonb:


Ical.Net/Utility/DateUtil.cs
```diff
@@ -63,8 +62,11 @@ private static Dictionary<string, string> InitializeWindowsMappings()
     /// <param name="tzId">A BCL, IANA, or serialization time zone identifier</param>
     /// <param name="useLocalIfNotFound">If true, this method will return the system local time zone if tzId doesn't match a known time zone identifier.
     /// Otherwise, it will throw an exception.</param>
+    /// <exception cref="ArgumentException">Unrecognized time zone id</exception>
     public static DateTimeZone GetZone(string tzId, bool useLocalIfNotFound = true)
     {
+        var exMsg = $"Unrecognized time zone id {tzId}";
+
         if (string.IsNullOrWhiteSpace(tzId))
```
axunonb, line 70: We removed the implicit usage of the local time zone in other parts, didn't we? Hope, this is the only one left.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -634,7 +634,7 @@ public int CompareTo(IDateTime? dt)
     public string ToString(string? format, IFormatProvider? formatProvider)
     {
         formatProvider ??= CultureInfo.InvariantCulture;
-        var dateTimeOffset = DateUtil.ToZonedDateTimeLeniently(Value, _tzId).ToDateTimeOffset();
+        var dateTimeOffset = DateUtil.ToZonedDateTimeLeniently(Value, _tzId ?? string.Empty).ToDateTimeOffset();
```
minichma, line 637: Shouldn't we handle floating separately? Not sure how the method behaves if we pass an empty string.

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -66,18 +62,21 @@ public static bool Equals<T>(IEnumerable<T> left, IEnumerable<T> right, bool ord
             return left.SequenceEqual(right);
         }
 
+        // No multiple enumerations
+        var rightArray = right as T[] ?? right.ToArray();
```
minichma, line 66: Eager iteration would only help if right/left would be generators, which doesn't seem to be the case in any of today's uses, otherwise it would rather add overhead.

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -66,18 +62,21 @@ public static bool Equals<T>(IEnumerable<T> left, IEnumerable<T> right, bool ord
...
+            return leftArray.OrderBy(l => l).SequenceEqual(rightArray.OrderBy(r => r));
         }
         catch (Exception)
         {
             //It's not possible to sort some collections of things (like Calendars) in any meaningful way. Properties can be null, and there's no natural
             //ordering for the contents therein. In cases like that, the best we can do is treat them like sets, and compare them. We don't maintain
             //fidelity with respect to duplicates, but it seems better than doing nothing
-            var leftSet = new HashSet<T>(left);
-            var rightSet = new HashSet<T>(right);
+            var leftSet = new HashSet<T>(leftArray);
```
minichma, line 78: I think this whole way of comparing isn't too helpful. We discussed this in another ticket, that `Equals` might not be too helpful for many of the types, e.g. Calendar. But that's a topic for a different PR.

minichma (approved):

Ical.Net/Utility/CollectionHelpers.cs
axunonb: Yes, I wanted to stick to NRT

PR #719, opened 2025-02-10
Make parameterless CTOR of CalDateTime private

axunonb:
Reasoning: Should not be part of the public API


minichma (dismissed):
Very nice! Could the constructor even be made private?

axunonb:
> Very nice! Could the constructor even be made private?

Indeed it could. "At little as possible" would then mean `private`. Should we prefer that?

minichma:
> > Very nice! Could the constructor even be made private?
> 
> Indeed it could. "At little as possible" would then mean `private`. Should we prefer that?

If it doesn't cause any issues with the existing code base, then I'd say yes.

PR #723, opened 2025-02-11
Type of RecurrencePattern.Until is CalDateTime

### LINKED ISSUE #588: RecurrencePattern.Until DateTime property should be a CalDateTime

### LINKED ISSUE #653: `RecurrencePattern.Until` should be of type `IDateTime` instead of `DateTime`
Reasoning:
In the `RecurrencePatternSerializer` the `RecurrencePattern.Until` is necessarily converted to an `IDateTime` for serialization.
That's why it would be better to store it as an `IDateTime` from the beginning, instead of relying on `RecurrencePattern.Until.Kind`.

```C#
if (recur.Until != DateTime.MinValue)
{
    var serializer = factory.Build(typeof(IDateTime), SerializationContext) as IStringSerializer;
    if (serializer != null)
    {
        IDateTime until = new CalDateTime(recur.Until); // <<<<<<<<< The DateTimeKind is important here
        until.HasTime = true;
        values.Add("UNTIL=" + serializer.SerializeToString(until));
    }
}
```

axunonb:
Changed type of `RecurrencePattern.Until` from `DateTime` to `CalDateTime`
Change `ArgumentException` to `ArgumentOutOfRangeException` in `RecurrencePatternSerializer`

Resolves #588 
Resolves #653

### DIFF
--- a/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
@@ -31,7 +31,7 @@ public class RecurrencePatternEvaluator : Evaluator
         // Convert the UNTIL value to one that matches the same time information as the reference date
         if (r.Until is not null)
         {
-            r.Until = MatchTimeZone(referenceDate, r.Until.Value);
+            r.Until = MatchTimeZone(referenceDate, r.Until);
         }
 
         if (referenceDate.HasTime)
@@ -886,7 +886,7 @@ public class RecurrencePatternEvaluator : Evaluator
         return periodQuery;
     }
 
-    private static DateTime MatchTimeZone(CalDateTime reference, DateTime until)
+    private static CalDateTime MatchTimeZone(CalDateTime reference, CalDateTime until)
     {
         /*
            The value of the "UNTIL" rule part MUST have the same value type as the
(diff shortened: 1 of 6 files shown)


Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -908,17 +908,17 @@ specified as a date with UTC time.
...
+        var untilCalDt = new CalDateTime(until.Value, untilTzId, reference.HasTime);
 
         // If 'reference' is floating, then 'until' is floating, too
         return reference.TzId is null
-            ? untilCalDt.Value
-            // convert to the reference timezone
-            : untilCalDt.ToTimeZone(reference.TzId).Value;
+            ? untilCalDt
+            // convert to the reference timezone and convert the value to Floating
+            : untilCalDt.ToTimeZone(reference.TzId).Value.AsCalDateTime();
```
minichma: We should have a more elegant way of converting between floating and non-floating. Can't we write `.ToTimeZone(null)`? I feel that we should actually have something like `CalDateTime.AsFloating()` and `CalDateTime.WithZoned(tzId)` (which is not a topic of this PR and which is in contrast to my comments in a previous discussion - I know)

minichma (dismissed):

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: OMG, of course, this is implemented already: `ToTimeZone(reference.TzId).ToTimeZone(null)`, as simple as that 👍

PR #726, opened 2025-02-12
Fix: Derive correct file and assembly version from package version

axunonb:


axunonb:
@minichma Sorry, I noticed too late, the `-pre.xx` breaks the Action script. Should work now. The package also gets attached as an artifact to each Action run.
Till 26 Feb I'm in another timezone, with possible web access limitations (in case you wonder about missing reactions).

PR #732, opened 2025-02-15
Make the time zone resolver plugable

minichma:
This PR implements several improvements related to resolving time zones:

* Fix test cases so they use well defined and known time zone IDs, so they don't fall back to the system's local TZ
* Stop falling back to the local TZ if the specified TZ is unknown. Its more than likely that in most cases the fallback is not what is expected/desired by the user.
* Make the time zone resolving functionality plugable


minichma:
Relates to the discussion in https://github.com/ical-org/ical.net/discussions/606#discussioncomment-12031951

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -2291,7 +2291,7 @@ public void Yearly1()
     public void Bug2912657()
     {
         var iCal = Calendar.Load(IcsFiles.Bug2912657);
-        var localTzid = iCal.TimeZones[0].TzId;
+        var localTzid = iCal.Events.First().Start.TzId;
```
axunonb, line 2294: 👍

Ical.Net/DefaultTimeZoneResolver.cs
```diff
@@ -0,0 +1,96 @@
...
+        .ToDictionary(k => k.Key, v => v.Value, StringComparer.OrdinalIgnoreCase);
+
+    private static readonly Lazy<Dictionary<string, string>> _windowsMapping
+        = new Lazy<Dictionary<string, string>>(InitializeWindowsMappings, LazyThreadSafetyMode.PublicationOnly);
+
+    /// <summary>
+    /// Use this method to turn a raw string into a NodaTime DateTimeZone. It searches all time zone providers (IANA, BCL, serialization, etc) to see if
+    /// the string matches. If it doesn't, it walks each provider, and checks to see if the time zone the provider knows about is contained within the
+    /// target time zone string. Some older icalendar programs would generate nonstandard time zone strings, and this secondary check works around
+    /// that.
```
axunonb, line 28: 👍

axunonb (approved):
Great idea!

Ical.Net/DefaultTimeZoneResolver.cs
minichma: The code in this method is copied (almost) 1:1 from the original in `DateUtil`.

minichma:
> Great idea!

Thanks and thanks for the review! Having global static parameters is always a somewhat poor design but in this case it seems to be the best short term solution without refactoring the whole evaluation code. At least for our project its a viable workaround. Could/should be improved in v6.

PR #738, opened 2025-02-23
Fix incorrect handling of UNTIL if falling into DST change and some related improvements

### LINKED ISSUE #736: `UNTIL` not handled correctly if it falls into a DST change
**Describe the bug**
`UNTIL` not handled correctly if it falls into a DST change.

The following VEVENT should produce two occurrences, but it produces only one:

```ics
DTSTART;TZID=Europe/Amsterdam:20201024T023000
DURATION:PT5M
RRULE:FREQ=DAILY;UNTIL=20201025T010000Z
```

The expected occurrences are at `20201024T023000` and `20201025T023000` but the latter is not returned. I.e. `20201025T023000` exists twice in CET due to a DST change at 3:00 back to 2:00.

- .NET version: .net 8.0

minichma:
Reproduce and fix #736. I.e. if `UNTIL` falls into a DST change with ambiguity, then recurrences could have missed, which is fixed with this PR.

Other improvements include:
* Remove implicit conversion operator from `System.DateTime` to `CalDateTime`, which could cause confusion and was a source of bugs.
* Remove `AsUtc` from `CalDateTime.GetHashCode()` to avoid timezone calculations
* Avoid UTC conversion when comparing two `CalDateTime` if they have the same `TzId`

### DIFF
--- a/Ical.Net.Benchmarks/OccurencePerfTests.cs
+++ b/Ical.Net.Benchmarks/OccurencePerfTests.cs
@@ -60,7 +60,7 @@ public class OccurencePerfTests
             {
                 var rrule = new RecurrencePattern(FrequencyType.Daily, 1)
                 {
-                    Until = startTime.AddDays(10),
+                    Until = new CalDateTime(startTime.AddDays(10)),
                 };
 
                 var e = new CalendarEvent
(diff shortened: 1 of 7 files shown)


axunonb (approved):
Nice, thanks. Sorry for the delay.

PR #743, opened 2025-03-02
Fix: Minor NRT warnings with Recurrence

axunonb:
Fix: Minor NRT warnings with Recurrence
Enable NRT for RecurrenceTests

### DIFF
--- a/Ical.Net/Utility/CollectionHelpers.cs
+++ b/Ical.Net/Utility/CollectionHelpers.cs
@@ -261,7 +261,7 @@ internal static class CollectionHelpers
 
         foreach (var item in items)
         {
-            if (first || !comparer.Equals(prev, item))
+            if (first || !comparer.Equals(prev!, item))
                 yield return item;
 
             prev = item;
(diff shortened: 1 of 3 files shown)


Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -261,7 +261,7 @@ public static IEnumerable<T> OrderedDistinct<T>(this IEnumerable<T> items, IEqua
 
         foreach (var item in items)
         {
-            if (first || !comparer.Equals(prev, item))
+            if (first || (!EqualityComparer<T>.Default.Equals(prev, default) && !comparer.Equals(prev, item)))
```
minichma: Shouldn't use an additional comparer here for multiple reasons
* involves additional comparisons, which even involves time zone calculations in case of CalDateTime
* The passed comparer might have special handling implemented for `default`, which might be in conflict with the default comparer
* The default comparer might not even be able to compare instances of T properly (not so likely though)
* Not required from a functional standpoint, only introduced to make the compiler happy.

Would rather go for `comparer.Equals(prev!, item)` or suppressing the warning.

minichma (approved):

PR #748, opened 2025-03-06
Update `PRODID` and `VERSION` property handling

### LINKED ISSUE #531: Setting own ProductId or Version has no effect
Setting a own ProductId or Version on the Calendar has no effect. The ProductId and Version properties are replaced on the serialization with the library meta data.

```
var calendar = new Calendar
{
    ProductId = "MyCalendarTool",
    Version ="1.0"
};
```

https://github.com/rianjs/ical.net/blob/6c03c42bd9e040622ffaa240be856dc531a20823/src/Ical.Net/Serialization/CalendarSerializer.cs#L25-L38

axunonb:
Issue: `Calendar` has setters for `ProductId` and `Version` which are overridden with fixed values when serializing.

- When creating a new `Calendar` instance, `ProductId` and `Version` contain default values
- When Deserializing an iCalendar, `ProductId` and `Version` will be taken from the input
- `ProductId` and `Version` can be overridden by user code. An attempt to set as an empty string will throw.
- Update the default `PRODID` property `LibraryMetadata.ProdId` to include the ical.net assembly version. Example: "PRODID:-//github.com/ical-org/ical.net//NONSGML ical.net 5.4.3//EN"
- Modified `CalendarSerializer.SerializeToString` so that the `ProdId` or `Version` set by users do not get overridden
- Add an xmldoc description about the purpose of `ProdId` and `Version`, and about the risks when modified
- Add unit tests

Resolves #531


axunonb:
@minichma TBD: Add the current ical.net version to the PRODID

minichma:
The RFC says the PRODID should contain

> Any text that describes the product and version

so, yes, including the version would probably be a good idea.

axunonb:
@minichma  RFC5545 says that PRODID and VERSION are mandatory. This may be the reason for the current implementation (always set the defaults when serializing).
But given this ics with missing PRODID and VERSION to deserialize:
```
BEGIN:VCALENDAR
BEGIN:VEVENT
DTSTART:20070406T230000Z
DTEND:20070407T010000Z
END:VEVENT
END:VCALENDAR
```
TBD:
Options
- Make it compliant by adding missing values: Then we don't see the original values.
- Leave it invalid. It's up to the user to fix, if needed before serializing (my preferred option, - breaking change to previous versions)
- Leave the properties unchanged (invalid), but replace empty PRODID or VERSION when serializing (non-braking, but kind of unexpected magic)

minichma:
> Leave it invalid. It's up to the user to fix, if needed before serializing (my preferred option, - breaking change to previous versions)

My preferred option too. Only populate ProdId and Version when creating a new calendar, but not when deserializing an existing one.

Ical.Net/Calendar.cs
```diff
@@ -143,20 +147,54 @@ public override int GetHashCode()
...
+    /// <para/>
+    /// The default value does not apply to deserialized objects.
+    /// </summary>
     public virtual string Version
     {
         get => Properties.Get<string>("VERSION");
-        set => Properties.Set("VERSION", value);
+        set
+        {
+            if (string.IsNullOrEmpty(value))
```
minichma: Just a nit - not sure we need a special handling for null and empty here. Wouldn't know why for instance "XXX" would be any better than "". So for simplicity and to reduce the work f or CodeCov (aka the number of branches) I'd rather avoid such checks where not explicitly necessary.

Ical.Net/Calendar.cs
```diff
@@ -143,20 +147,54 @@ public override int GetHashCode()
...
+    /// <para/>
+    /// The default value does not apply to deserialized objects.
+    /// </summary>
     public virtual string ProductId
     {
         get => Properties.Get<string>("PRODID");
-        set => Properties.Set("PRODID", value);
+        set
+        {
+            if (string.IsNullOrEmpty(value))
```
minichma: Same as above - wouldn't see any requirement for special handling of null and empty.

minichma (dismissed):

Ical.Net/Calendar.cs
axunonb: The intention was to avoid users create invalid iCalendars. `VERSION` and `PRODID` must be set. But yes, xmldocs clearly say "know what you are doing". So I follow your comment regarding KISS.

PR #750, opened 2025-03-06
Evaluation: Make `MaxIncrementCount` configurable

### LINKED ISSUE #747: Introduce `EvaluationOptions` for params like `MaxIncrementCount`
Today we the param `MaxIncrementCount` is hard coded in `RecurrencePatternEvaluator`. It limits the number of iterations that will be calculated by the class without returning any recurrence, before aborting the calculations.

The implementation has two major shortcomings:
1. It is not configurable
2. The calculation aborts silently if the limit is hit.

Ad 1)
It would be good to introduce something like `EvaluationOptions` that is passed to `Calendar.GetOccurrences` et al. For now the class should hold only the `MaxIncrementCount` param that should replace the hard-coded value. In the future it could be extended by additional params. The new param to `GetOccurrences()` would default to null, in which case default values would apply. The default value is TBD, I would suggest to have no limit as the default.

Ad 2)
If the limit is hit, calculation stops silently, which prevents users from detecting the case. The default behavior should be that a specific exception is raised, maybe something like `IcalEvaluationLimitExceededException` or alternatively some more generic `IcalEvaluationException` holding a more detailed error code as a property. Basically the behavior (fail/stop silently) could also be configured via `EvaluationOptions` but that doesn't seem to be too helpful as users can easily catch the exception and implement ignore it at the call site if desired.

In the future the `EvaluationOptions` could hold additional properties
- the TZDB to be used
- ...

I think it would be good to introduce the `EvaluationOptions` still in v5 so we have an extension point that allows us adding more in the future, without breaking the API.

minichma:
Several improvements:
* Remove the seemingly unneeded `includeReferenceDateInResults` from `Evaluate()` et al as it shouldn't be needed as the expected behavior is pretty well defined in the RFC. Only exception: What to happen if an RRULE doesn't match DTSTART, in which case we don't include `DTSTART` as we already didn't before.
* Introduce new `EvaluationOptions` type and pass it to `Calendar.GetOccurrences()` and related
* Replace the hard-coded `RecurrencePatternEvaluator.MaxIncrementCount` with a new configuration property `EvaluationOptions.MaxIncrementCount`.
* If the limit is hit, raise an `MaxIncrementsExceededEvaluationException` rather than stopping evaluation silently.

Fixes #747


minichma:
@axunonb Please have a look - I think its an improvement. The issues sonarcube is complaining about are legacy, so nothing to fix in this PR. The overall code coverage increase, so should be ok from that standpoint too.

TBD: The default for `MaxIncrementCount` changed from previously 1000 to indefinite. What do you think? (Should probably be mentioned in the release notes.)

Constructing the options instance might be better to be done via a Builder pattern, so we'd be more flexible when extending it in the future without risking breaking changes. We could also default to some default instance rather than passing `null` as the default but this would require some extra thoughts that I don't really have the time for right now. A default instance should probably be immutable, so the type as a whole should be immutable, which again would suggest using a Builder pattern. TBD

Ical.Net/CalendarComponents/CalendarEvent.cs
```diff
@@ -251,9 +251,6 @@ private void Initialize()
     /// <returns>True if the event has not been cancelled, False otherwise.</returns>
     public virtual bool IsActive => !string.Equals(Status, EventStatus.Cancelled, EventStatus.Comparison);
 
-    /// <inheritdoc/>
-    protected override bool EvaluationIncludesReferenceDate => true;
-
```
axunonb, line 256: I stumbled across that too, it makes sense

Ical.Net/Evaluation/EvaluationException.cs
```diff
@@ -0,0 +1,14 @@
...
+//
+
+using System;
+
+namespace Ical.Net.Evaluation;
+
+/// <summary>
+/// Represents an exception that may occur during calendar evaluation.
+/// </summary>
+public class EvaluationException : Exception
```
axunonb, line 13: 👍

Ical.Net/Evaluation/MaxIncrementsExceededEvaluationException.cs
```diff
@@ -0,0 +1,14 @@
...
+//
+
+namespace Ical.Net.Evaluation;
+
+/// <summary>
+/// Represents an exception will be raised during calendar evaluation when the maximum number of
+/// increments is exceeded.
+/// </summary>
+/// <seealso cref="EvaluationOptions.MaxIncrementCount"/>
+public class MaxIncrementsExceededEvaluationException : EvaluationException
```
axunonb: 👍

Ical.Net/Evaluation/PeriodListEvaluator.cs
```diff
@@ -19,16 +19,10 @@ public PeriodListEvaluator(PeriodList rdt)
         _mPeriodList = rdt;
     }
 
-    public override IEnumerable<Period> Evaluate(CalDateTime referenceDate, CalDateTime? periodStart, CalDateTime? periodEnd, bool includeReferenceDateInResults)
+    public override IEnumerable<Period> Evaluate(CalDateTime referenceDate, CalDateTime? periodStart, CalDateTime? periodEnd, EvaluationOptions options)
     {
```
axunonb: No coverage, wondering when this method would get called

Ical.Net/Evaluation/EvaluationOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// The maximum number of increments to evaluate without finding a recurrence before
+    /// evaluation is stopped exceptionally. If null, the evaluation will continue indefinitely.
+    /// </summary>
+    /// <remarks>
+    /// This option only applies to the evaluation of RecurrencePatterns.
+    ///
+    /// If the specified number of increments is exceeded without finding a recurrence, an
+    /// exception of type <see cref="Ical.Net.Evaluation.MaxIncrementsExceededEvaluationException"/> will be thrown.
+    /// </remarks>
+    public int? MaxIncrementCount { get; set; }
```
axunonb: As it's not a counter, would sth. like `NoPatternMatchLimit` be more intuitive?
Likewise sth. like `NoPatternMatchLimitExceededException`?
and TestMaxIncrementCount method

Ical.Net/Evaluation/RecurringEvaluator.cs
```diff
@@ -41,20 +41,13 @@ protected IEnumerable<Period> EvaluateRRule(CalDateTime referenceDate, CalDateTi
...
+            return ruleEvaluator.Evaluate(referenceDate, periodStart, periodEnd, options);
         })
             // Enumerate the outer sequence (not the inner sequences of periods themselves) now to ensure
             // the initialization code is run, including validation and error handling.
             // This way we receive validation errors early, not only when enumeration starts.
             .ToList(); //NOSONAR - deliberately enumerate here
 
-
-        //Only add referenceDate if there are no RecurrenceRules defined
-        if (includeReferenceDateInResults && (Recurrable.RecurrenceRules == null || !Recurrable.RecurrenceRules.Any()))
```
axunonb, line 53: Also had no coverage before

axunonb:
Looks great!

> TBD: The default for MaxIncrementCount changed from previously 1000 to indefinite.

We already had exchanged some thoughts about indefinite, so let's use it as the default, but add an explicit hints in our xmldoc about the risk. Same with release notes, sure.

> options instance might be better to be done via a Builder pattern

Balancing benefit and related effort the current way should be appropriate.

> `MaxIncrementCount `

Pls. consider the comments above.

Ical.Net/Evaluation/PeriodListEvaluator.cs
minichma: Within Ical.Net not at all I assume, because the only place where it *could* be used is `RecurringEvaluator`, where we use `Recurrable.RecurrenceDates.GetAllPeriodsByKind()`. Remove?!

Ical.Net/Evaluation/EvaluationOptions.cs
minichma: Certainly better!

Ical.Net/Evaluation/RecurringEvaluator.cs
minichma: Yeah, because `EvaluateRRule` would be called if no recurrence rules existed. 😄

minichma:
@axunonb Thought a little about the naming, now choose `MaxUnmatchedIncrementsLimit` and `EvaluationLimitExceededException`, what do you think?

Ical.Net/Evaluation/EvaluationOptions.cs
axunonb: Your new naming is the best

axunonb (approved):
Perfectr, thank you!

Issue #749, opened 2025-03-06
Unrecognized time zone id Customized Time Zone


jakoss:
**Describe the bug**
I'm trying to parse attached ics file, but i have an error: Unrecognized time zone id Customized Time Zone.

I'm pretty sure that the entries with this Customized Time Zone are entries created by accepting invitation from Google Meet from outlook application.

**To Reproduce**
Try to parse attached file
```csharp
var calendarString = File.ReadAllText(@"<your path>\calendar.ics");
var calendar = Calendar.Load(calendarString);
var occurrences = calendar.GetOccurrences(DateTime.UtcNow, DateTime.UtcNow).ToList(); // crashes here
```

1. iCalendar Data
Attached in file [calendar.zip](https://github.com/user-attachments/files/19110133/calendar.zip)

**Expected behavior**
I know there is no "proper" solution to handle this, but for now the only thing i can do is to do enumeration manually and handle the exception there which is not ideal. I'd love to have some solution, maybe just skip such entries but return everything else?

**Environment (please complete the following information):**
- OS: Windows 11
- .NET version: .NET 8
- ical.net version: 5.0.0-pre.41

jakoss:
Sorry, my mistake. It was Dump() function in LINQPad that triggered the exception. Without dumping event object i can handle this in place properly

axunonb:
@minichma Seems to be broken starting with #732

minichma:
> Seems to be broken starting with https://github.com/ical-org/ical.net/pull/732

Yes, it was an explicit decision to stop falling back to the local timezone if the TZ is unknown. From the PR description:

> * Stop falling back to the local TZ if the specified TZ is unknown. Its more than likely that in most cases the fallback is not what is expected/desired by the user.

I'm pretty sure that also in this case the fallback to the local tz is not what is desired. Actually if I understand @jakoss correctly, they just want it to work 'some way'.

Possible solutions and workarounds:
* The right solution would actually be to consider the TZ definition for the unknown TZ within the ics file, but that's not something we have in place today.
* Keep falling back to local tz - probably not a good idea
* Fall back to some other random TZ like UTC or assume floating time - probably also not a good idea, as it simply isn't correct
* Make the fallback configurable via the new `EvaluationOptions` - better
* Make the whole TZ resolution configurable via `EvaluationOptions` - much better, but not too straight forward for now
* Introduce some configuration that allows skipping values that can't be resolved - should be doable
[...]

PR #753, opened 2025-03-10
Remove `IServiceProvider`

minichma:
The concept of the `IServiceProvider` seems overly complex and makes reasoning on the actual control flow hard. It is used for managing `IEvaluator` instances and for serialization. Remove the use for `IEvaluator` for now, which already allows removing it from the model types. Keep it for serialization for now and remove it in a subsequent PR.


minichma:
@axunonb This PR is a tiny step towards decoupling evaluation from the model, following the discussion in https://github.com/ical-org/ical.net/issues/749#issuecomment-2710909941.

axunonb (approved):
> makes reasoning on the actual control flow hard

Absolutely, love it!

Issue #740, opened 2025-02-24
GetOccurrences startTime and endTime inclusive or exclusive?


RemcoBlok:
I would like to understand how the startTime and endTime parameters for the `Calendar.GetOccurrences<CalendarEvent>()` methods apply to the Occurrence.Period in v5. Is the startTime inclusive and the endTime exclusive?  In v5 I think endTime is exclusive. In v4 I thought endTime was inclusive. In v4 you had to subtract one tick from the endTime to get exclusive behaviour.

Is it `startTime <= Occurence.Period.StartTime && Occurence.Period.StartTime < endTime`? 
Or is it `startTime <= Occurence.Period.StartTime && Occurence.Period.StartTime <= endTime`? 

To find out how it works I created some tests and observed some behaviour I do not understand.

The following test passes and works as I expected.

```cs
        [Fact]
        public void GetOccurrencesTest1()
        {
            // Arrange
            string timeZoneId = TimeZoneInfo.Local.Id;
            
            CalendarEvent myEvent = new CalendarEvent();
            myEvent.Start = new CalDateTime(2024, 10, 19, 0, 0, 0, timeZoneId);
            myEvent.End = myEvent.Start.AddDays(1);
            
            Calendar calendar = new Calendar();
            calendar.Events.Add(myEvent);

            // Act
            CalDateTime startTime = new CalDateTime(2024, 10, 19, 0, 0, 0, timeZoneId);
            CalDateTime endTime = startTime.AddDays(1);
            HashSet<Occurrence> occurrences = calendar.GetOccurrences<CalendarEvent>(startTime, endTime).ToHashSet();

            // Assert
            Assert.Single(occurrences);
        }
```

The following test also passes and suggests we only need to match the startTime and endTime to the Occurrence.Period.StartTime. (I use an AddTicks() extension method)

```cs
        [Fact]
        public void GetOccurrencesTest2()
        {
            // Arrange
            string timeZoneId = TimeZoneInfo.Local.Id;

            CalendarEvent myEvent = new CalendarEvent();
            myEvent.Start = new CalDateTime(2024, 10, 19, 0, 0, 0, timeZoneId);
            myEvent.End = myEvent.Start.AddDays(1);

            Calendar calendar = new Calendar();
            calendar.Events.Add(myEvent);

            // Act
            CalDateTime startTime = new CalDateTime(2024, 10, 19, 0, 0, 0, timeZoneId);
            CalDateTime endTime = startTime.AddTicks(1);
[...]

PR #754, opened 2025-03-12
Add `Ical.Net.TimeZone` namespace and related classes

axunonb:
Introduced new namespace `Ical.Net.TimeZone` containing:
* `TimeZoneResolver` (just moved)
* `VTimeZoneExtensions`: Allow to convert `VTimeZone`s to `DateZimeZone` (explicitly to internal `CalDateTimeZone`) keeping the calendar component lean
* `CalDateTimeZone` inheriting form `DateZimeZone` to handle timezones created from `VTimeZone`s
* `CalTimeZoneProvider` to be used with `TimeZoneResolver` for timezones included in VTIMEZONE iCalendar components


axunonb:
@minichma Please take a look at `VTimeZoneTest.Occurrences_WithinUndefined_TzIntervals()`:
We do get occurrences for dates where no timezone interval exists. The periods, however, resolve to wrong values returned from `NodaTime`. This is to be expected, and the behavior is in line with RFC5545 ("undefined"). Consequently this test should be removed. I feel this not a good behavior, though?

minichma:
@axunonb I will be on vacation until Sunday. Will have a look when I'm back.

Ical.Net.Tests/Calendars/Recurrence/VTimeZone3.ics
```diff
@@ -0,0 +1,28 @@
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Sample Corp//NONSGML Event//EN
+BEGIN:VTIMEZONE
+TZID:July August Timezone
+X-LIC-LOCATION:July August Timezone
+BEGIN:STANDARD
+DTSTART:20250701T000000
+DTEND:20250731T000000
```
minichma: `DTEND` is not allowed in `STANDARD` or `DAYLIGHT` objects.

Ical.Net.Tests/Calendars/Recurrence/VTimeZone3.ics
```diff
@@ -0,0 +1,28 @@
...
+BEGIN:DAYLIGHT
+DTSTART:20250801T000000
+DTEND:20250830T000000
+TZOFFSETFROM:+0100
+TZOFFSETTO:-0100
+TZNAME:August -1
+END:DAYLIGHT
+END:VTIMEZONE
+BEGIN:VEVENT
+DTSTART;TZID=July August Timezone:20240601T090000
```
minichma: There are two `DTSTART`, which is illegal.

Ical.Net/VTimeZoneProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        return provider;
+    }
+
+    private VTimeZoneProvider()
+    {
+        _customTimeZones = new();
+    }
+
+    private static bool TryCreateIntervals(VTimeZone vTimeZone, out string tzId, out List<ZoneInterval> intervals)
```
minichma: When evaluating timezone intervals, RRULEs and RDATEs need to be considered. I suggest using `TimeZoneEvaluator` for that purpose.

minichma (commented):
I had a very brief look at `Occurrences_WithinUndefined_TzIntervals()`. I assume, a TZ-related test for intervals where no TZ is defined only makes sense if we very clearly define what should happen in such case (which would most likely throwing an exception IMHO). But the more fundamental topic seems to be that `VTimeZoneProvider` is not fully implemented. I.e. it needs to consider RRULEs and RDATEs. Recommend to use `TimeZoneEvaluator` for that purpose.

I also recommend adding some tests that compare the time zone definitions from NodaTime to those generated by `VTimeZoneProvider`. The related TZ definitions could be taken [from here](https://github.com/libical/tzdbics/tree/main/zoneinfo). It would be best to compare *all* defined timezones, but then we'd have to use the exact same version of TZDB, which would require additional maintenance efforts in the future, so maybe just compare just some selected TZ that are unlikely to have their definitions like `Europe/Paris`, `America/New_York` or the like.

axunonb:
@minichma This WIP is a rewrite of the initial PR for handling custom timezones, incorporating your recommendations. Unit tests using _America_New_York.ics_ and _America_Los_Angeles.ics_ from _libical_ currently produce results consistent with NodaTime. Additional test cases are still needed. Do you have any feedback on this early preview? 
I'm still considering the final implementation—ideally, timezones embedded in the iCalendar would work seamlessly...

Ical.Net/TimeZone/VTimeZoneExtensions.cs
```diff
@@ -0,0 +1,189 @@
...
+        foreach (var tzi in timeZoneInfos)
+        {
+            var zoneTransitionsCount = 0;
+            var isDaylight = tzi.Name?.Equals(Components.Daylight, StringComparison.OrdinalIgnoreCase) == true;
+            var tzName = tzi.TimeZoneName ?? (isDaylight ? "DST" : "STD");
+            var offsetFrom = tzi.OffsetFrom != null ?  Offset.FromTimeSpan(tzi.OffsetFrom.Offset) : Offset.Zero;
+            var offsetTo = tzi.OffsetTo != null ? Offset.FromTimeSpan(tzi.OffsetTo.Offset) : Offset.Zero;
+            var savings = isDaylight ? (offsetTo - offsetFrom) : Offset.Zero;
+
+            foreach (var rrule in tzi.RecurrenceRules)
```
minichma: The whole evaluation logic is implemented in `TimeZoneInfoEvaluator` (which doesn't add any functionality over `RecurringEvaluator` and could therefore be removed), so I recommend reusing that.

axunonb: Oh yes, sure - sorry

Ical.Net/TimeZone/CalTimeZoneProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+    private readonly Dictionary<string, DateTimeZone> _timeZones = new();
+
+    /// <summary>
+    /// Adds a <see cref="DateTimeZone"/> to the provider if it does not already exist.
+    /// </summary>
+    /// <param name="timeZone">The <see cref="DateTimeZone"/> to add.</param>
+    private void Add(DateTimeZone timeZone)
+    {
+#if NET6_0_OR_GREATER
+        _timeZones.TryAdd(timeZone.Id, timeZone);
```
minichma, line 32: If adding is tolerant regarding duplicates, we should follow the .net conventions and name the method (and the calling ones) `bool TryAdd(...)`. Otherwise we should throw.

Ical.Net/TimeZone/CalTimeZoneProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// <summary>
+    /// Converts all <see cref="VTimeZone"/>s to <see cref="DateTimeZone"/>s
+    /// and adds them to the provider.
+    /// <para/>
+    /// The <see cref="VTimeZone.TzId"/> is used as the <see cref="DateTimeZone.Id"/>.
+    /// </summary>
+    /// <param name="timeZones"></param>
+    /// <param name="untilYear">The year to which timezone transitions will be calculated.</param>
+    public void AddRangeFrom(ICollection<VTimeZone> timeZones, int untilYear)
```
minichma: Not sure we need the `untilYear` here. We could just enumerate as needed.

Ical.Net/TimeZone/CalTimeZoneProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+/// </summary>
+public class CalTimeZoneProvider : IDateTimeZoneProvider
+{
+    private readonly Dictionary<string, DateTimeZone> _timeZones = new();
+
+    /// <summary>
+    /// Adds a <see cref="DateTimeZone"/> to the provider if it does not already exist.
+    /// </summary>
+    /// <param name="timeZone">The <see cref="DateTimeZone"/> to add.</param>
+    private void Add(DateTimeZone timeZone)
```
minichma, line 29: Most often its preferable to make classes like this immutable. I.e. pass all the required details in the constructor. This simplifies the interface and complexity of the class significantly. I wouldn't see a reason why mutability would be required.

Ical.Net/TimeZone/CalTimeZoneProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// <summary>
+    /// Clears all time zones from the provider.
+    /// </summary>
+    public void Clear() => _timeZones.Clear();
+
+    /// <inheritdoc/>
+    public DateTimeZone? GetZoneOrNull(string id)
+    {
+        // Check for fixed-offset timezones, as described for IDateTimeZoneProvider
+        if (id.Equals(DateTimeZone.Utc.Id, StringComparison.OrdinalIgnoreCase))
```
minichma, line 80: Hm, why would we add special treatment for UTC? This class is dedicated to deal with time zones based on VTimeZone. Wouldn't artificially add UTC here.

Ical.Net/TimeZone/CalTimeZoneProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+            return DateTimeZone.Utc;
+
+        if (id.StartsWith(DateTimeZone.Utc.Id, StringComparison.OrdinalIgnoreCase))
```
minichma, line 83: Same here, this has nothing to do with VTimeZone. If we think, additional functionality is needed, this should be added outside of this class, e.g. by providing some utility function that allows combining multiple providers, where one could be an instance of this class and the other could deal with UTC. But without having thought about that in detail, I don't think this is something this project should deal with.

Ical.Net/TimeZone/CalTimeZoneProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+            var parseResult = NodaTime.Text.OffsetPattern.GeneralInvariant.Parse(id.Substring(3));
+            if (parseResult.Success)
+                return DateTimeZone.ForOffset(parseResult.Value);
+        }
+
+        return _timeZones.TryGetValue(id, out var zone) ? zone : null;
+    }
+
+    /// <inheritdoc/>
+    public string VersionId => LibraryMetadata.AssemblyFileVersion;
```
minichma, line 94: Hm, the VersionId seems to refer to the version of the data, not to the version of the library. E.g. 
`NodaTime.DateTimeZoneProviders.Tzdb.VersionId` returns `TZDB: 2024b (mapping: $Revision$)` (not sure, what the `$Revision$` string stands for though)

However, there is corresponding version field in VCALENDAR, so maybe just return the PRODID, pass the version from outside (so the caller would know what to specify) or just leave it as is.

Ical.Net/TimeZone/CalDateTimeZone.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// <inheritdoc/>
+    public override ZoneInterval GetZoneInterval(Instant instant)
+    {
+        // instant is a NodaTime internal LocalInstant
+        // that is created from LocalDateTime
+
+        // Handle instants before the first zone interval
+        if (_zoneIntervals.Count == 0 || instant < _zoneIntervals[0].Start)
+            return CreateZoneInterval(
```
minichma, line 50: If we don't have any time zone information, we shouldn't guess here. It's completely unclear, what the reason is for no time zone info being present. It could easily be that the time zone info has intentionally been limited to a certain time range, in which case returning some fallback based on heuristics will most likely be wrong. Just return either null or throw.

Ical.Net/TimeZone/CalDateTimeZone.cs
```diff
@@ -0,0 +1,101 @@
...
+                high = mid - 1;
+            else if (instant >= interval.End)
+                low = mid + 1;
+            else
+                return interval;
+        }
+
+        // If we get here, the instant is *after all defined intervals*
+        // Return a default interval (this case shouldn't happen with proper data)
+        return CreateZoneInterval(
```
minichma, line 77: Same as above - if we don't know the actual data, throw or return null.

Ical.Net/TimeZone/CalDateTimeZone.cs
```diff
@@ -0,0 +1,101 @@
...
+            return CreateZoneInterval(
+                "LMT", // Before standardized time zones used for "Local Mean Time" :)
+                Instant.MinValue, // makes zoneInterval.localStart being equal to StartOfTime
+                _zoneIntervals.Count > 0 ? _zoneIntervals[0].Start : Instant.MaxValue,
+                Offset.Zero,
+                Offset.Zero,
+                Id);
+
+        // Binary search for the matching interval
+        var low = 0;
```
minichma, line 59: Isn't this already implemented in some NodaTime type we could reuse?

Ical.Net/TimeZone/VTimeZoneExtensions.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            var isDaylight = tzi.Name?.Equals(Components.Daylight, StringComparison.OrdinalIgnoreCase) == true;
+            var tzName = tzi.TimeZoneName ?? (isDaylight ? "DST" : "STD");
+            var offsetFrom = tzi.OffsetFrom != null ? Offset.FromTimeSpan(tzi.OffsetFrom.Offset) : Offset.Zero;
+            var offsetTo = tzi.OffsetTo != null ? Offset.FromTimeSpan(tzi.OffsetTo.Offset) : Offset.Zero;
+            var savings = isDaylight ? (offsetTo - offsetFrom) : Offset.Zero;
+
+            var tzEvaluator = new TimeZoneInfoEvaluator(tzi);
+            var start = tzi.Start!; // Start is not null due to the filter above
+            var end = new CalDateTime(untilYear, 12, 13, 23, 59, 59);
```
minichma: 31 rather than 13

Ical.Net/TimeZone/VTimeZoneExtensions.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Will be set in <see cref="CreateIntervalsFromTransitions"/>
+        /// </summary>
+        public Instant? End { get; set; } = end;
+        public Offset Offset { get; } = offset;
+        public Offset Savings { get; } = savings;
+    }
+
+    private static bool TryCreateIntervals(VTimeZone vTimeZone, int untilYear, out string tzId,
```
minichma: Try to avoid the need for `untilYear` altogether? Alternatively just combine the individual IEnumerables to a single one using `.OrderedMerge()` and store the corresponding `IEnumerator`. When more data is needed, just continue the enumeration?

minichma (commented):
I didn't have the time for a proper review, especially in VTimeZoneExtensions, just a few thoughts for now.

Ical.Net/TimeZone/VTimeZoneExtensions.cs
minichma: Just noted that we have some inconsistency regarding periodEnd being inclusive or exclusive. I.e. in https://github.com/ical-org/ical.net/blob/c7faa31892b665783220c8927e5d97032332a1c3/Ical.Net/Evaluation/RecurrenceUtil.cs#L44 `periodEnd` is handled exclusive (which I'd consider correct), while in https://github.com/ical-org/ical.net/blob/c7faa31892b665783220c8927e5d97032332a1c3/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L184 it is considered inclusive (which should be fixed).

(see #740)

axunonb: Hm, if we rewrite the existing condition
```cs
if (candidate < periodStart || candidate >= periodEnd)
{
    if (periodStart != periodEnd || candidate > periodEnd)
    {
        continue;
    }
}
```
isn't `periodEnd` **exclusive** then, except for `periodStart == periodEnd`? (same as in RecurrenceUtil)?

minichma: If I read it correctly, in this case periodEnd would still be treated inclusive and periodStart would be mostly ignored. 

However, I feel the whole start/end evaluation is quite broken. If I'm right, `RecurringEvaluator` doesn't filter RDATEs at all, that's only done by RecurrenceUtil. So if you use the Evaluator directly as done in this PR, this might cause inconsistencies.

I'd recommend fixing this start/end topic first. Question is how to design the behavior. Actually its usually always a good idea and I'd heavily recommend to have the upper bound exclusive to avoid having to deal with rounding issues, epsilons, 23:59:59, .AddSecond(-1), etc. On the other hand the extra handling of start==end is also less than ideal. This special handling increases complexity quite significantly, as the caller always needs to consider this special case.  But having `periodEnd` being exclusive AND avoiding the start==end special case makes it harder for the caller to ask for recurrences at a certain instant. An elegant solution would be to remove the `periodEnd` param altogether. Now that `Evaluate` and `GetOccurrences` return a generator IEnumerable, the callers can easily apply something like `TakeWhile(p => p.StartDate <= endDate)` themselves. Would also make quite some code significantly simpler. This would also come very handy, should we consider extending the evaluation functionality by something like evaluating  backwards in the future. What do you think?

minichma: Just a very quick (untested, unreviewed) sketch: https://github.com/minichma/ical.net/tree/work/minichma/wip/remove_period_end - nice how the complex limiting expressions go away :-)

axunonb: >  the extra handling of start==end is also less than ideal

Less than ideal is a cautious description for something we counter in several places with different, complex code

> consider extending the evaluation functionality by something like evaluating backwards in the future. 

Wow, perfect! And `periodEnd` or `untilYear` just fly into the past. Hope we get this implemented.

axunonb:
Obsolete after latest commits to version/6.0

Issue #576, opened 2023-07-04
Natural Language de/serialization


paviad:
A very desirable feature of this library, and which exists in similar libraries in other languages (e.g. [Typescript's Equivalent](https://github.com/jakubroztocil/rrule) is the ability to serialize and deserialize to and from natural language, for example:

`Every day at 10, 12 and 17` should deserialize into `FREQ=DAILY;BYHOUR=10,12,17`

and vice versa.

I think this should be implemented as a separate (and optional) NuGet package.

paviad:
I also have made an Antlr4 parser - I think it is better, easier to reason about the parsing (and also to maintain/modify)

axunonb:
@minichma There are 2 different aspects mentioned:
1. Natural language: Not a high priority feature, and should go into an extension package
2. Replace Regex with a Parser: A valid point regarding performance, memory allocations, maintainability. At NCalc we're using [Parlot](https://github.com/sebastienros/parlot) - unfortunately no experience on my side.

How to deal with this issue?

minichma:
Ad 1) I'm really not the right one to comment on that as it is strongly tied to the future strategy of this project. At this time I feel that addressing existing issues would be a number one priority and is enough work for a while. So adding new features would not be too helpful for that as someone will have to work on it and, more notably, maintain it in the future. But this could also be maintained in a separate repo. @paviad would it be an option for you to publish this as an independent package in a separate repo?

Ad 2) Are we talking about a parser for the new feature (natural language) or for ical.net as a whole. @paviad I assume the first, right? Then its tied to 1. The ical.net parser seems to be rather stable, but if an alternative existed in the form of a PR, I would assume it would be welcome.

paviad:
Can I use the name *Ical.Net.NaturalLanguage* ?

axunonb:
@paviad The name would be fine - maybe a bit long. Sth. shorter like `ICal.Net.Phrases`...? This was a co-pilot suggestion :)
If you like, we can create the repo under `ical-org` (GitHub organization) to strengthen the "brand" with you as maintainer or admin.

axunonb:
@minichma 
>  I feel that addressing existing issues would be a number one priority and is enough work for a while

Fully agree

> the future strategy of this project

Making ICal.Net the best in class for dot net - in terms of usability, performance and RFC compliance / compatibility

minichma:
> Making ICal.Net the best in class for dot net - in terms of usability, performance and RFC compliance / compatibility

[...]

PR #758, opened 2025-03-30
Enhance `RecurrencePatternSerializer`

### LINKED ISSUE #755: `RecurrencePatternSerializer` expects that the `FREQ`` is the first in Recurrence Rules
The RecurrencePattern class when parsing RRule should return logically the same RRule string. However, I see when the INTERVAL is specified before the FREQ in the RRule the RecurrencePattern ignores INTERVAL and obviously could not return it through the ToString().

**To Reproduce**

This was executed in an immediate window.
```
new RecurrencePattern("RRULE:FREQ=DAILY;INTERVAL=2;UNTIL=20250430T000000Z").ToString()
// output: "FREQ=DAILY;INTERVAL=2;UNTIL=20250430T000000Z"

new RecurrencePattern("RRULE:INTERVAL=2;FREQ=DAILY;UNTIL=20250430T000000Z").ToString()
// output: "FREQ=DAILY;UNTIL=20250430T000000Z" // faulty - does not have INTERVAL
```

**Expected behavior**
The INTERVAL value should be set in RecurrencePattern object and returned through ToSTtring() when the INTERVAL is specified before the FREQ in RRule string.

**Environment**
- Windows
- .NET 8
- ical.net 4.3.1

axunonb:
- RRULE parts are now processed in any order, which is compliant with RFC5545 section 3.3.10. Resolves #755
- Disallowed recurrence pattern parts will throw `ArgumentOutOfRangeException`s
- Refactor methods with high cognitive complexity
- Remove deserialization of natural language input

### DIFF
--- a/Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
+++ b/Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
@@ -8,7 +8,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text.RegularExpressions;
 using Ical.Net.DataTypes;
 
 namespace Ical.Net.Serialization.DataTypes;
@@ -176,12 +175,6 @@ public class RecurrencePatternSerializer : EncodableDataTypeSerializer
         return Encode(recur, string.Join(";", values));
     }
 
-    private const RegexOptions CiCompiled = RegexOptions.IgnoreCase | RegexOptions.Compiled;
-
-    internal static readonly Regex FrequencyPattern =
-        new Regex(@"FREQ=(SECONDLY|MINUTELY|HOURLY|DAILY|WEEKLY|MONTHLY|YEARLY);?(.*)", CiCompiled,
-            RegexDefaults.Timeout);
-
     /// <summary>
     /// Deserializes an RRULE value string into an <see cref="RecurrencePattern"/> object.
     /// <para/>
@@ -211,20 +204,17 @@ public class RecurrencePatternSerializer : EncodableDataTypeSerializer
         // Decode the value, if necessary
         value = Decode(r, value);
 
-        return TryDeserializeFrequencyPattern(value, r, factory)
-            ? r
-            : null;
+        DeserializePattern(value, r, factory);
+        return r;
     }
 
-    private bool TryDeserializeFrequencyPattern(string value, RecurrencePattern r, ISerializerFactory factory)
+    /// <summary>
+    /// Deserializes the recurrence rule.
+    /// </summary>
+    /// <exception cref="ArgumentOutOfRangeException">Always throws on failure.</exception>
+    private void DeserializePattern(string value, RecurrencePattern r, ISerializerFactory factory)
     {
-        // We can't proceed if the FREQ part is missing
-        var match = FrequencyPattern.Match(value);
-        if (!match.Success)
-        {
-            return false;
-        }
-
+        var freqPartExists = false;
         var keywordPairs = value.Split(';');
         foreach (var keywordPair in keywordPairs)
         {
@@ -241,13 +231,21 @@ public class RecurrencePatternSerializer : EncodableDataTypeSerializer
                     $"The recurrence rule part '{keywordPair}' is invalid.");
             }
 
+            if (keyValues[0].Equals("FREQ", StringComparison.OrdinalIgnoreCase))
+            {
+                freqPartExists = true;
+            }
+
             ProcessKeyValuePair(keyValues[0].ToLower(), keyValues[1], r, factory);
         }
 
+        if (!freqPartExists || r.Frequency == FrequencyType.None)
+        {
+            throw new ArgumentOutOfRangeException(nameof(value),
+                "The recurrence rule must specify a FREQ part that is not NONE.");
+        }
         CheckMutuallyExclusive("COUNT", "UNTIL", r.Count, r.Until);
         CheckRanges(r);
-
-        return true;
     }
 
     private void ProcessKeyValuePair(string key, string value, RecurrencePattern r, ISerializerFactory factory)
@@ -310,6 +308,10 @@ public class RecurrencePatternSerializer : EncodableDataTypeSerializer
             case "wkst":
                 r.FirstDayOfWeek = GetDayOfWeek(value);
                 break;
+
+            default:
+                throw new ArgumentOutOfRangeException(nameof(key),
+                    $"The recurrence rule part '{key}' is not supported.");
         }
     }
 
--- a/Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
+++ b/Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
@@ -176,30 +176,54 @@ public class RecurrencePatternSerializer : EncodableDataTypeSerializer
         return Encode(recur, string.Join(";", values));
     }
 
-    // Compiling here is a one-time penalty of about 80ms
+    // Compiling here is a one-time penalty
     private const RegexOptions _ciCompiled = RegexOptions.IgnoreCase | RegexOptions.Compiled;
 
-    internal static readonly Regex OtherInterval =
-        new Regex(@"every\s+(?<Interval>other|\d+)?\w{0,2}\s*(?<Freq>second|minute|hour|day|week|month|year)s?,?\s*(?<More>.+)", _ciCompiled, RegexDefaults.Timeout);
+    internal static readonly Regex NaturalLanguagePattern =
+        new Regex(
+            @"every\s+(?<Interval>other|\d+)?\w{0,2}\s*(?<Freq>second|minute|hour|day|week|month|year)s?,?\s*(?<More>.+)",
+            _ciCompiled, RegexDefaults.Timeout);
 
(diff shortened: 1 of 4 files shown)


axunonb:
@minichma TBD:
`RecurrencePatternSerializer.TryDeserializeNaturalLanguagePattern` can process kind of natural language input. This ability is not in use and fully untested. Also this is part of the regular deserialization and gets called when the `FREQ` part is missing. I'm in doubt whether this functionality brings any added value. Should we remove it completely or at least use a different entry point for this kind of parsing?

BTW: Just noticed, we can ask Copilot for a review, too :)

minichma:
Didn't look into the actual implementation so far, but generally speaking I think the standard serializers/deserializers should stick to the RFC. So when parsing a calendar from any other format, this should happen fully intentional (not as a fallback from the standard parser) and independently. I therefore agree that it should rather not be an integral part of standard deserialization.

If this included as part of Ical.Net, it should be implemented in separate classes that could be used independently. The functionality reminds me of https://github.com/ical-org/ical.net/issues/576. Could this be combined?

axunonb:
>  The functionality reminds me of https://github.com/ical-org/ical.net/issues/576. Could this be combined?

Yes, it has some similarities. I wonder whether natural langue for recurrences brings much advantage. End users would be better off with a good GUI, while devops might prefer a builder with a fluent API.

Anyway, the natural language part is undocumented, untested, and can never be all fallback for RFC-compliance. I'll remove it.

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
-    internal static readonly Regex Time = new Regex(@"at\s+(?<Hour>\d{1,2})(:(?<Minute>\d{2})((:|\.)(?<Second>\d{2}))?)?\s*(?<Meridian>(a|p)m?)?",
-        _ciCompiled, RegexDefaults.Timeout);
-
-    internal static readonly Regex RecurUntil = new Regex(@"^\s*until\s+(?<DateTime>.+)$", _ciCompiled, RegexDefaults.Timeout);
-
-    internal static readonly Regex SpecificRecurrenceCount = new Regex(@"^\s*for\s+(?<Count>\d+)\s+occurrences\s*$", _ciCompiled, RegexDefaults.Timeout);
-
+    private const RegexOptions CiCompiled = RegexOptions.IgnoreCase | RegexOptions.Compiled;
+
+    internal static readonly Regex FrequencyPattern =
```
minichma: make private?

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
+        new Regex(@"FREQ=(SECONDLY|MINUTELY|HOURLY|DAILY|WEEKLY|MONTHLY|YEARLY);?(.*)", CiCompiled,
+            RegexDefaults.Timeout);
```
minichma: I know, this is legacy, but I think, we shouldn't specify a timeout in Regex. Timeouts in Regex are good if the pattern is user-specified, but in our case there is no reason to expect any bad behavior. On the other hand, the timeout could easily cause unexpected behavior. E.g. on a server running close to its limits, a timeout of 200 ms could easily be hit also on well-formed input.

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
minichma: Do we actually need this regex at all (anymore)? The only thing it really does now is validating that a RRULE contains a `FREQ=xxx` string (e.g. `thisIsNotFREQ=DAILY!` would work), but that's really not too helpful as either all parts of the string would have to be validated or nothing. The real validation happens throughout the parsing process.

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
 
         // Decode the value, if necessary
         value = Decode(r, value);
 
-        var match = AdverbFrequencies.Match(value);
-        if (match.Success)
-        {
-            // Parse the frequency type
-            r.Frequency = (FrequencyType) Enum.Parse(typeof(FrequencyType), match.Groups[1].Value, true);
+        return TryDeserializeFrequencyPattern(value, r, factory)
```
minichma: If the pattern can't be parsed, why would we return null? Shouldn't we raise an exception?

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
-                }
+                throw new ArgumentOutOfRangeException(nameof(value),
+                    $"The recurrence rule part '{keywordPair}' is invalid.");
             }
+
+            ProcessKeyValuePair(keyValues[0].ToLower(), keyValues[1], r, factory);
         }
-        else
+
+        CheckMutuallyExclusive("COUNT", "UNTIL", r.Count, r.Until);
```
minichma, line 247: Check frequency is present here, rather than doing the regex upfront.

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
-            {
-                r.Interval = !int.TryParse(match.Groups["Interval"].Value, out var interval)
-                    ? 2
-                    : interval;
-            }
-            else
+            if (keywordPair.Length == 0)
             {
-                r.Interval = 1;
+                // ignore subsequent semi-colons
```
minichma, line 223: .. as well as preceding and duplicate ones

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
+            case "count":
+                r.Count = Convert.ToInt32(value);
+                break;
+
+            case "interval":
+                r.Interval = Convert.ToInt32(value);
+                break;
+
+            case "bysecond":
+                AddInt32Values(r.BySecond, value);
```
minichma, line 273: By adding the entry, we accept `BYSECOND` to appear more than once (same for all 'BY' parts). Legacy though.

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
+    }
+
+    private void ProcessKeyValuePair(string key, string value, RecurrencePattern r, ISerializerFactory factory)
+    {
+        switch (key)
         {
-            // Couldn't parse the object, return null!
-            r = null;
+            case "freq":
+                r.Frequency = (FrequencyType) Enum.Parse(typeof(FrequencyType), value, true);
```
minichma, line 256: I guess we should fail if `None` is specified.

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
```diff
@@ -176,280 +176,154 @@ private static void SerializeByValue(List<string> aggregate, IList<int> byValue,
...
+                break;
+
+            case "bysetpos":
+                AddInt32Values(r.BySetPosition, value);
+                break;
+
+            case "wkst":
+                r.FirstDayOfWeek = GetDayOfWeek(value);
+                break;
         }
```
minichma, line 315: Shouldn't we add a default case for unknown/invalid rule parts and fail here?

minichma (commented):
Great, much better! Just a few minor comment.

minichma (dismissed):

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
axunonb: Let's make the timeout 2000 ms, which is unlikely ever to be hit

axunonb:
Agree with your comments on existing code. Starting with a bug fix and ending up with a rewrite, while limited time is running by...

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
axunonb: Will change to always throw on failure

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
axunonb: all kinds - will add a unit test

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
axunonb: won't check

Ical.Net/Serialization/DataTypes/RecurrencePatternSerializer.cs
axunonb: Good point

minichma (approved):

PR #759, opened 2025-04-01
Evaluation: Avoid dependency on local culture settings.

minichma:
The use of `CultureInfo.CurrentCulture.Calendar` is one of the remaining dependencies on the local system settings, which is fixed with this PR.


axunonb (commented):
Makes sense, just 2 typos (besides Sonar recommendations)

axunonb (approved):

minichma:
Thanks for the review!

PR #762, opened 2025-04-07
Enable NRT for `AlarmOccurrence`,`Attachment`, `Attendee`, `CalendarDataType`, `Duration`, `ICalendarDataType`, `ICalendarParameterCollectionContainer`

axunonb:
Now < 100 files left for NRT :) - 30% done

### DIFF
--- a/Ical.Net.Tests/RecurrenceTests.cs
+++ b/Ical.Net.Tests/RecurrenceTests.cs
@@ -2672,7 +2672,7 @@ public class RecurrenceTests
 
         evt.RecurrenceRules.Add(pattern);
 
-        var occurrences = evt.GetOccurrences(new DateTime(2018, 1, 1), DateTime.MaxValue).ToList();
+        var occurrences = evt.GetOccurrences(new CalDateTime(2018, 1, 1), new CalDateTime(DateTime.MaxValue, false)).ToList();
         Assert.That(occurrences, Has.Count.EqualTo(10), "There should be 10 occurrences of this event.");
     }
 
@@ -2741,7 +2741,7 @@ public class RecurrenceTests
 
         evt.RecurrenceRules.Add(pattern);
 
-        var occurrences = evt.GetOccurrences(new DateTime(2011, 1, 1), new DateTime(2012, 1, 1)).ToList();
+        var occurrences = evt.GetOccurrences(new CalDateTime(2011, 1, 1), new CalDateTime(2012, 1, 1)).ToList();
         Assert.That(occurrences, Has.Count.EqualTo(10), "There should be 10 occurrences of this event, one for each month except February and December.");
     }
 
@@ -2782,7 +2782,7 @@ public class RecurrenceTests
         //Testing on both the first day and the next, results used to be different
         for (var i = 0; i <= 1; i++)
         {
-            var checkTime = DateTime.Parse("2019-01-04T08:00Z").ToUniversalTime();
+            var checkTime = new CalDateTime(2019, 01, 04, 08, 00, 00, CalDateTime.UtcTzId);
             checkTime = checkTime.AddDays(i);
             //Valid asking for the exact moment
             var occurrences = vEvent.GetOccurrences(checkTime, checkTime).ToList();
@@ -2811,7 +2811,7 @@ public class RecurrenceTests
             End = new CalDateTime(DateTime.Parse("2020-01-11T00:00")),
         };
 
-        var occurrences = vEvent.GetOccurrences(DateTime.Parse("2020-01-10T00:00"), DateTime.Parse("2020-01-11T00:00"));
+        var occurrences = vEvent.GetOccurrences(new CalDateTime(2020,01, 10, 0, 0, 0), new CalDateTime(2020, 01, 11, 00, 00, 00));
         Assert.That(occurrences.Count, Is.EqualTo(0));
     }
 
@@ -3023,7 +3023,7 @@ public class RecurrenceTests
 
         Assert.That(() =>
         {
-            _ = evt.GetOccurrences(DateTime.Today.AddDays(1), DateTime.Today.AddDays(2));
+            _ = evt.GetOccurrences(CalDateTime.Today.AddDays(1), CalDateTime.Today.AddDays(2));
         }, Throws.Exception, "An exception should be thrown when evaluating a recurrence with no specified FREQUENCY");
     }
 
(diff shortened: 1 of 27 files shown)


Ical.Net/DataTypes/Duration.cs
```diff
@@ -118,7 +119,8 @@ public static Duration FromSeconds(int seconds) =>
     /// Parses the specified value according to RFC 5545.
     /// </summary>
     /// <exception cref="System.FormatException">Thrown if the value is not a valid duration.</exception>
-    public static Duration Parse(string value) => (Duration)new DurationSerializer().Deserialize(new StringReader(value));
+    public static Duration? Parse(string value) =>
```
minichma: Unfortunately we are somewhat inconsistent here. According to the XML doc the method should throw if the passed string is not a valid duration. `DurationSerializer` throws in some cases and returns `null` in other. I'd generally suggest to throw and return a non-nullable Duration.

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -69,7 +69,10 @@ public override void CopyFrom(ICopyable obj)
 
         foreach (var p in list)
         {
-            Add(p.Copy<Period>());
+            if (p != null)
```
minichma: By not copying null values the copy will not be equal to the original. Maybe to do something like `Add(p?.Copy<Period>())`?

Ical.Net/DataTypes/PeriodList.cs
```diff
@@ -121,11 +124,11 @@ public void Insert(int index, Period item)
     {
         EnsureConsistentTimezoneAndPeriodKind(item);
         if (Periods.Contains(item)) return;
-        Periods.Insert(index, item);
+        Periods?.Insert(index, item);
```
minichma: This `?` hast just been removed one commit earlier.

Ical.Net/DataTypes/Attendee.cs
```diff
@@ -276,10 +276,10 @@ protected bool Equals(Attendee other) => Equals(SentBy, other.SentBy)
                                              && Rsvp == other.Rsvp
                                              && Equals(Value, other.Value)
                                              && Members.SequenceEqual(other.Members)
-                                             && DelegatedTo.SequenceEqual(other.DelegatedTo)
-                                             && DelegatedFrom.SequenceEqual(other.DelegatedFrom);
+                                             && (DelegatedTo?.SequenceEqual(other.DelegatedTo ?? Enumerable.Empty<string>()) ?? other.DelegatedTo == null)
```
minichma: This comparison is not commutative this way. I.e. if one is null and the other is empty, then the comparison is not symmetrical. Not sure though, whether an empty sequence should equal null. Usually they're considered different.

Ical.Net/DataTypes/Attachment.cs
```diff
@@ -93,15 +94,15 @@ public override void CopyFrom(ICopyable obj)
...
 
-    protected bool Equals(Attachment other)
+    protected bool Equals(Attachment? other)
     {
-        var firstPart = Equals(Uri, other.Uri) && ValueEncoding.Equals(other.ValueEncoding);
+        var firstPart = Equals(Uri, other?.Uri) && ValueEncoding.Equals(other?.ValueEncoding);
         return Data == null
             ? firstPart
-            : firstPart && Data.SequenceEqual(other.Data);
+            : firstPart && other?.Data != null && Data.SequenceEqual(other.Data);
```
minichma: This comparison is not correct. If `other?.Data == null`, then `this.Data` is ignored, even if it contains data.

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -1232,8 +1232,8 @@ public void WeekNoOrderingShouldNotMatter()
         var rpe1 = new RecurrencePatternEvaluator(new RecurrencePattern("FREQ=YEARLY;WKST=MO;BYDAY=MO;BYWEEKNO=1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31,33,35,37,39,41,43,45,47,49,51,53"));
         var rpe2 = new RecurrencePatternEvaluator(new RecurrencePattern("FREQ=YEARLY;WKST=MO;BYDAY=MO;BYWEEKNO=53,51,49,47,45,43,41,39,37,35,33,31,29,27,25,23,21,19,17,15,13,11,9,7,5,3,1"));
 
-        var recurringPeriods1 = rpe1.Evaluate(new CalDateTime(start), start, end, default).ToList();
-        var recurringPeriods2 = rpe2.Evaluate(new CalDateTime(start), start, end, default).ToList();
+        var recurringPeriods1 = rpe1.Evaluate(new CalDateTime(start), start, end, null).ToList();
```
minichma: Yes, better. I started with `default` because the options should have been a value type at first.

Ical.Net/DataTypes/AlarmOccurrence.cs
```diff
@@ -44,14 +45,21 @@ public AlarmOccurrence(Alarm a, CalDateTime dt, IRecurringComponent rc)
         Component = rc;
     }
 
-    public int CompareTo(AlarmOccurrence other) => Period.CompareTo(other.Period);
+    public int CompareTo(AlarmOccurrence? other)
+    {
+        if (other == null || other.Period == null)
```
minichma: What if `other.Period == null` and `this.Period == null`?

minichma (commented):
Very nice, just a few minor comments.

Ical.Net/DataTypes/PeriodList.cs
axunonb: Removed

axunonb:
Love your reviews, thanks!

Ical.Net/DataTypes/Duration.cs
```diff
@@ -119,8 +119,8 @@ public static Duration FromSeconds(int seconds) =>
     /// Parses the specified value according to RFC 5545.
     /// </summary>
     /// <exception cref="System.FormatException">Thrown if the value is not a valid duration.</exception>
-    public static Duration? Parse(string value) =>
-        (Duration?) new DurationSerializer().Deserialize(new StringReader(value));
+    public static Duration Parse(string value) =>
+        (Duration) new DurationSerializer().Deserialize(new StringReader(value))!; // throws if null
```
minichma, line 123: Actually it doesn't throw in all cases as of today (https://github.com/ical-org/ical.net/blob/3c4e95b876d6ea0ed108310df4f4dcc36ba30089/Ical.Net/Serialization/DataTypes/DurationSerializer.cs#L73) although it should.

minichma (dismissed):
Very nice!

PR #764, opened 2025-04-08
Enable NRT for `CalendarComponents`

axunonb:


Ical.Net/CalendarComponents/VTimeZone.cs
```diff
@@ -83,23 +84,25 @@ public static VTimeZone FromDateTimeZone(string tzId, DateTime earliestDateTimeT
...
         {
-            // first, get the latest standard and daylight intervals, find the oldest recurring date in both, set the RRULES for it, and create a VTimeZoneInfos out of them.
-            //standard
+            // first, get the latest standard and daylight intervals, find the oldest recurring date in both,
+            // set the RRULES for it, and create a VTimeZoneInfos out of them.
+
+            // Standard
             var standardIntervals = intervals.Where(x => x.Savings.ToTimeSpan() == new TimeSpan(0)).ToList();
-            var latestStandardInterval = standardIntervals.OrderByDescending(x => x.Start).FirstOrDefault();
+            var latestStandardInterval = standardIntervals.OrderByDescending(x => x.Start).First();
```
minichma: standardIntervals is not guaranteed to be non-empty here, so `.First()` might throw.

Ical.Net/CalendarComponents/VTimeZone.cs
```diff
@@ -293,8 +292,8 @@ public VTimeZone(string tzId) : this()
         Location = _nodaZone.Id;
     }
 
-    private DateTimeZone _nodaZone;
-    private string _tzId;
+    private DateTimeZone _nodaZone = DateTimeZone.Utc; // must initialize
```
minichma, line 302: In a future (v6) version, the Noda tz should not be part of this model object at all.

Ical.Net/Evaluation/TodoEvaluator.cs
```diff
@@ -67,7 +65,7 @@ private static void DetermineStartingRecurrence(IEnumerable<CalDateTime> rdate,
 
     private void DetermineStartingRecurrence(RecurrencePattern recur, ref CalDateTime referenceDateTime)
     {
-        if (recur.Count.HasValue)
+        if (recur.Count.HasValue && Todo.Start != null)
```
minichma: This whole `EvaluateToPreviousOccurrence` that this method belongs to, is pretty weird. Not sure, what the right behavior in this case is. Probably just return if Todo.Start is null. Going into the else branch if its null probably isn't correct.

minichma (dismissed):

Ical.Net/CalendarComponents/VTimeZone.cs
axunonb: Right, then we could just return early with the current `vTimeZone`

Ical.Net/Evaluation/TodoEvaluator.cs
axunonb: Yep, returning directly is better then.

Ical.Net/CalendarComponents/VTimeZone.cs
axunonb: Yes, agree. The initialization is a NRT thing, and shouldn't have side effects.

PR #771, opened 2025-04-14
Enable NRT for `ical.net` (top) namespace

axunonb:
Enable NRT for `ical.net` (top) namespace

### DIFF
--- a/Ical.Net/Calendar.cs
+++ b/Ical.Net/Calendar.cs
@@ -241,7 +241,7 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
         // These are the UID/RECURRENCE-ID combinations that replace other occurrences.
         var recurrenceIdsAndUids = this.Children.OfType<IRecurrable>()
             .Where(r => r.RecurrenceId != null)
-            .Select(r => new { ((IUniqueComponent) r).Uid, Dt = r.RecurrenceId!.Value })
+            .Select(r => new { (r as IUniqueComponent).Uid, Dt = r.RecurrenceId!.Value })
             .Where(r => r.Uid != null)
             .ToDictionary(x => x);
 
@@ -294,7 +294,7 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
             this.AddChild(cal);
             return (T) cal;
         }
-        throw new ArgumentException($"Creating {typeof(T)} failed.");
+        throw new ArgumentException($"Creating {typeof(T).FullName} failed.");
     }
 
     public virtual void MergeWith(IMergeable obj)
--- a/Ical.Net/Calendar.cs
+++ b/Ical.Net/Calendar.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license.
 //
 
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -20,7 +21,7 @@ namespace Ical.Net;
 
 public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, IMergeable
 {
-    public static Calendar Load(string iCalendarString)
+    public static Calendar? Load(string iCalendarString)
         => CalendarCollection.Load(new StringReader(iCalendarString)).SingleOrDefault();
 
     /// <summary>
@@ -28,11 +29,11 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
     /// </summary>
     /// <param name="s">The stream from which to load the <see cref="Calendar"/> object</param>
     /// <returns>An <see cref="Calendar"/> object</returns>
-    public static Calendar Load(Stream s)
+    public static Calendar? Load(Stream s)
         => CalendarCollection.Load(new StreamReader(s, Encoding.UTF8)).SingleOrDefault();
 
-    public static Calendar Load(TextReader tr)
-        => CalendarCollection.Load(tr)?.SingleOrDefault();
+    public static Calendar? Load(TextReader tr)
+        => CalendarCollection.Load(tr).SingleOrDefault();
 
     public static IList<T> Load<T>(Stream s, Encoding e)
         => Load<T>(new StreamReader(s, e));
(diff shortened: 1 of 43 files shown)


Ical.Net/Calendar.cs
```diff
@@ -225,12 +235,13 @@ public virtual IEnumerable<Occurrence> GetOccurrences(CalDateTime startTime = nu
...
     /// <param name="endTime">The ending date range</param>
-    public virtual IEnumerable<Occurrence> GetOccurrences<T>(CalDateTime startTime = null, CalDateTime endTime = null, EvaluationOptions options = default) where T : IRecurringComponent
+    /// <param name="options"></param>
+    public virtual IEnumerable<Occurrence> GetOccurrences<T>(CalDateTime? startTime = null, CalDateTime? endTime = null, EvaluationOptions? options = null) where T : IRecurringComponent
     {
         // These are the UID/RECURRENCE-ID combinations that replace other occurrences.
         var recurrenceIdsAndUids = this.Children.OfType<IRecurrable>()
             .Where(r => r.RecurrenceId != null)
-            .Select(r => new { (r as IUniqueComponent)?.Uid, Dt = r.RecurrenceId.Value })
+            .Select(r => new { ((IUniqueComponent) r).Uid, Dt = r.RecurrenceId!.Value })
```
minichma: What if r is not an IUniqueComponent? Previously we'd have returned null, now it would fail with an invalid cast.

Ical.Net/Calendar.cs
```diff
@@ -283,37 +294,33 @@ public T Create<T>() where T : ICalendarComponent
             this.AddChild(cal);
             return (T) cal;
         }
-        return default(T);
+        throw new ArgumentException($"Creating {typeof(T)} failed.");
```
minichma: Very nice! Not sure what Type.ToString() produces, but maybe specify something like `typeof(T).FullName`?

Ical.Net/CalendarCollection.cs
```diff
@@ -57,38 +57,41 @@ private IEnumerable<Occurrence> GetOccurrences(Func<Calendar, IEnumerable<Occurr
...
     }
 
-    public FreeBusy GetFreeBusy(FreeBusy freeBusyRequest)
+    public FreeBusy? GetFreeBusy(FreeBusy freeBusyRequest)
     {
-        return this.Aggregate<Calendar, FreeBusy>(null, (current, iCal) => CombineFreeBusy(current, iCal.GetFreeBusy(freeBusyRequest)));
+        return this.Aggregate<Calendar, FreeBusy?>(null, (current, iCal) =>
+        {
+            var freeBusy = iCal.GetFreeBusy(freeBusyRequest);
+            return current is null ? freeBusy : CombineFreeBusy(current, freeBusy!);
```
minichma: Hm, if `freeBusy` could be null here, we shouldn't apply `!`. Not sure, what the right behavior is but either `GetFreeBusy()` should return not-null or we should deal with `null`s here.

Ical.Net/CalendarCollection.cs
```diff
@@ -57,38 +57,41 @@ private IEnumerable<Occurrence> GetOccurrences(Func<Calendar, IEnumerable<Occurr
...
-        return this.Aggregate<Calendar, FreeBusy>(null, (current, iCal) => CombineFreeBusy(current, iCal.GetFreeBusy(fromInclusive, toExclusive)));
-    }
-
-    public FreeBusy GetFreeBusy(Organizer organizer, IEnumerable<Attendee> contacts, CalDateTime fromInclusive, CalDateTime toExclusive)
-    {
-        return this.Aggregate<Calendar, FreeBusy>(null, (current, iCal) => CombineFreeBusy(current, iCal.GetFreeBusy(organizer, contacts, fromInclusive, toExclusive)));
+        return this.Aggregate<Calendar, FreeBusy?>(null, (current, iCal) =>
+        {
+            var freeBusy = iCal.GetFreeBusy(organizer, contacts, fromInclusive, toExclusive);
+            return current is null ? freeBusy : CombineFreeBusy(current, freeBusy!);
```
minichma: `freeBusy!` - same as above

minichma (dismissed):
Very nice, a significant improvement!

A comment on making reviews somewhat more efficient: It would be helpful to keep changes apart in separate commits that can be reviewed independently. E.g. pure formatting changes (line breaks, spacing, ...), especially when applied by formatting tools, usually don't require the level of attention that functional changes do, so it would be helpful to have them in a separate commit that can be skimmed over very quickly. Functional changes, that are independent of each other could also be committed separately. Example:

* Commit: Pure formatting changes
* Commit: Unrelated functional change 1
* Commit: Unrelated functional change 2
* Commit: Preparatory functional change 1 required for implementing NRT
* Commit: Preparatory functional change 2 required for implementing NRT
* Commit: NRT
* Commit: Documentation changes
* ...

Not a big deal, just because time is very limited on all ends ;-)

Ical.Net/Calendar.cs
axunonb: `$"{typeof(T)}"` will indeed render the fully qualified name of the type T as a string

Ical.Net/Calendar.cs
```diff
@@ -241,7 +241,7 @@ public virtual IEnumerable<Occurrence> GetOccurrences<T>(CalDateTime? startTime
         // These are the UID/RECURRENCE-ID combinations that replace other occurrences.
         var recurrenceIdsAndUids = this.Children.OfType<IRecurrable>()
             .Where(r => r.RecurrenceId != null)
-            .Select(r => new { ((IUniqueComponent) r).Uid, Dt = r.RecurrenceId!.Value })
+            .Select(r => new { (r as IUniqueComponent).Uid, Dt = r.RecurrenceId!.Value })
```
minichma, line 244: This could throw with a npe now.

minichma (approved):
Should the remaining/new null reference warnings be considered?

axunonb:
> Should the remaining/new null reference warnings be considered?

Yes, they should or rather: must.
After all NRT related PRs are merged, a couple of more NRT warnings are to be expected.
So there will be another, hopefully final PR for NRT.

axunonb:
Thanks for your time and feedback.

PR #772, opened 2025-04-15
Enable NRT for `Serialization` namespace

axunonb:
* Now all library files have NRT enabled
* `DataTypeSerializerFactory` simplified
* `UtcOffsetSerializer.GetOffset(string)` simplified

### DIFF
--- a/Ical.Net/Serialization/DataTypes/AttachmentSerializer.cs
+++ b/Ical.Net/Serialization/DataTypes/AttachmentSerializer.cs
@@ -82,9 +82,8 @@ public class AttachmentSerializer : EncodableDataTypeSerializer
 
             // The default VALUE type for attachments is URI.
             // So, let's grab the URI by default.
-            var uriValue = Decode(a, attachment);
-
-            if (string.IsNullOrEmpty(uriValue)) return null;
+            // Note: Returned value can't be null here when args are not null.
+            var uriValue = Decode(a, attachment)!; 
 
             a.Uri = new Uri(uriValue, UriKind.RelativeOrAbsolute);
 
--- a/Ical.Net/Serialization/DataTypes/AttachmentSerializer.cs
+++ b/Ical.Net/Serialization/DataTypes/AttachmentSerializer.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license.
 //
 
+#nullable enable
 using System;
 using System.IO;
 using Ical.Net.DataTypes;
@@ -17,7 +18,7 @@ public class AttachmentSerializer : EncodableDataTypeSerializer
 
     public override Type TargetType => typeof(Attachment);
 
-    public override string SerializeToString(object obj)
+    public override string? SerializeToString(object? obj)
     {
         var a = obj as Attachment;
         if (a == null)
@@ -49,17 +50,22 @@ public class AttachmentSerializer : EncodableDataTypeSerializer
         return Encode(a, a.Data);
     }
 
-    public Attachment Deserialize(string attachment)
+    public Attachment? Deserialize(string attachment)
     {
         try
         {
             var a = CreateAndAssociate() as Attachment;
+            if (a == null) return null;
+
             // Decode the value, if necessary
             var data = DecodeData(a, attachment);
 
             // Get the currently-used encoding off the encoding stack.
             var encodingStack = GetService<EncodingStack>();
-            a.ValueEncoding = encodingStack.Current;
+            if (encodingStack != null)
+            {
+                a.ValueEncoding = encodingStack.Current;
+            }
 
             // Get the format of the attachment
             var valueType = a.GetValueType();
@@ -74,11 +80,13 @@ public class AttachmentSerializer : EncodableDataTypeSerializer
                 };
             }
 
-            // The default VALUE type for attachments is URI.  So, let's
-            // grab the URI by default.
+            // The default VALUE type for attachments is URI.
+            // So, let's grab the URI by default.
             var uriValue = Decode(a, attachment);
-            a.Uri = new Uri(uriValue, UriKind.RelativeOrAbsolute);
 
+            if (string.IsNullOrEmpty(uriValue)) return null;
+
+            a.Uri = new Uri(uriValue, UriKind.RelativeOrAbsolute);
 
             return a;
         }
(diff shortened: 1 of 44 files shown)


Ical.Net/Serialization/DataTypes/AttachmentSerializer.cs
```diff
@@ -74,11 +80,13 @@ public Attachment Deserialize(string attachment)
...
             }
 
-            // The default VALUE type for attachments is URI.  So, let's
-            // grab the URI by default.
+            // The default VALUE type for attachments is URI.
+            // So, let's grab the URI by default.
             var uriValue = Decode(a, attachment);
-            a.Uri = new Uri(uriValue, UriKind.RelativeOrAbsolute);
 
+            if (string.IsNullOrEmpty(uriValue)) return null;
```
minichma: Not sure, returning null is the best option here, is it? Anyhow, it's a functional change, right?

Ical.Net/Serialization/DataTypes/AttendeeSerializer.cs
```diff
@@ -17,37 +18,41 @@ public AttendeeSerializer(SerializationContext ctx) : base(ctx) { }
...
             : Encode(a, a.Value.OriginalString);
     }
 
-    public Attendee Deserialize(string attendee)
+    public Attendee? Deserialize(string attendee)
     {
         try
         {
-            var a = CreateAndAssociate() as Attendee;
+            if (CreateAndAssociate() is not Attendee a) return null;
```
minichma, line 33: rather throw?

Ical.Net/Serialization/DataTypes/AttendeeSerializer.cs
```diff
@@ -17,37 +18,41 @@ public AttendeeSerializer(SerializationContext ctx) : base(ctx) { }
...
+
             var uriString = Unescape(Decode(a, attendee));
 
+            if (uriString == null) return a;
```
minichma: Here the behavior is different to AttachmentSerializer, where we return `null`. I think, this way is better.

Ical.Net/Serialization/DataTypes/AttendeeSerializer.cs
```diff
@@ -17,37 +18,41 @@ public AttendeeSerializer(SerializationContext ctx) : base(ctx) { }
...
+            // Return null instead of throwing an exception
         }
 
         return null;
     }
 
-    public override object Deserialize(TextReader tr) => Deserialize(tr.ReadToEnd());
-}
\ No newline at end of file
+    public override object? Deserialize(TextReader? tr) => tr != null ? Deserialize(tr.ReadToEnd()) : null;
```
minichma, line 56: Why would it be allowed to pass null as the TextReader? Rather make it not nullable?

Ical.Net/Serialization/DataTypes/UtcOffsetSerializer.cs
```diff
@@ -30,43 +30,34 @@ public override string SerializeToString(object obj)
...
+        // Determine if the offset is negative
         var isNegative = rawOffset.StartsWith("-");
+        rawOffset = rawOffset.TrimStart('+', '-');
 
-        if (isNegative || rawOffset.StartsWith("+"))
-        {
-            rawOffset = rawOffset.Substring(1, rawOffset.Length - 1);
-        }
+        // Supported formats
+        var formats = new[] { "hhmmss", "hhmm", "hh" };
```
minichma: Could be declared as static property or field

Ical.Net/Serialization/DataTypes/WeekDaySerializer.cs
```diff
@@ -30,22 +31,33 @@ public override string SerializeToString(object obj)
         {
             value += ds.Offset;
         }
-        value += Enum.GetName(typeof(DayOfWeek), ds.DayOfWeek).ToUpper().Substring(0, 2);
+
+        try
+        {
+            var name = Enum.GetName(typeof(DayOfWeek), ds.DayOfWeek);
+            value += name!.ToUpper().Substring(0, 2);
```
minichma: If `name` is null, it would throw and be catched silently, which is ok, but not too readable. Handle explicitly?

minichma (dismissed):
Wow, that's a lot of changes. LGTM, just a few thoughts. Skimmed over the code only *very briefly* though.

Ical.Net/Serialization/DataTypes/AttachmentSerializer.cs
axunonb: I will have another look.

> Not sure, returning null is the best option

I believe returning `null` to avoid throwing exceptions is overused in the codebase. Methods like `TryGet(...)` would have been a better alternative to returning a NRT. That said, enabling nullable has introduced more `null` checks but reduced overall coverage. However, this investment will pay off in terms of future code quality. 
I've migrated projects to NRT several times, but this particular migration significantly exceeded my initial workload estimates. And that's without the unit tests still ahead of us.

Ical.Net/Serialization/DataTypes/AttendeeSerializer.cs
axunonb: We have different implementations using this pattern. Leave the current behavior for now.

Ical.Net/Serialization/DataTypes/AttachmentSerializer.cs
axunonb: returned value can't be null here, will update the code.
Even if the `uriValue` were null, it would throw, then the exception is eaten up and null ist returned anyway

Ical.Net/Serialization/DataTypes/AttendeeSerializer.cs
axunonb: Returned value can't be null here when args are not null.

Ical.Net/Serialization/DataTypes/AttendeeSerializer.cs
axunonb: It's because of overriding the base method, `SerializeToString(object? obj)` which is nullable

Ical.Net/Serialization/DataTypes/WeekDaySerializer.cs
axunonb: Agree, `if (name == null) return null;` is more readable

minichma (approved):

PR #775, opened 2025-04-24
EvaluationOptions: Fix off-by-one issue of `MaxUnmatchedIncrementsLimit`

minichma:
The evaluation of `EvaluationOptions.MaxUnmatchedIncrementsLimit` is off by one, which is fixed by this PR.

### DIFF
--- a/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
@@ -192,9 +192,10 @@ public class RecurrencePatternEvaluator : Evaluator
                 dateCount++;
             }
 
-            noCandidateIncrementCount++;
             if (noCandidateIncrementCount > options?.MaxUnmatchedIncrementsLimit)
                 throw new EvaluationLimitExceededException();
+
+            noCandidateIncrementCount++;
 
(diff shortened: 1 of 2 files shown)


Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -192,10 +192,11 @@ private IEnumerable<CalDateTime> EnumerateDates(CalDateTime originalDate, CalDat
                 dateCount++;
             }
 
-            noCandidateIncrementCount++;
             if (noCandidateIncrementCount > options?.MaxUnmatchedIncrementsLimit)
                 throw new EvaluationLimitExceededException();
 
+            noCandidateIncrementCount++;
```
axunonb, line 198: oh yes, sure

axunonb (approved):
Good finding, tnx

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
minichma: Its not critical at all, as the property really is intended to be used in a heuristical manner. That it was wrong only got obvious when setting the value to 0, which should work fine, as long as no iterations are skipped, but didn't. Thanks for the review!

PR #778, opened 2025-04-28
NRT clean-up after all NRT pull requests got merged

axunonb:
* `ISerializer.SerializationContext`is not nullable
* `SerializerBase.SerializationContext`: Ensure it is initialized and non-null by design
* `ServiceProvider.GetService(...)` returns a non-null instance or throws an exception if the service is unavailable.
* `EncodingProvider.DecoderDelegate` throws for invalid encoding argument
* `EncoderDelegate GetEncoderFor` throws for invalid encoding argument
* `SimpleDeserializer`: Ensure that `ParseContentLine(context, contentLineString).Vaue` is a string 
* Remove unneccessary conditional access qualifier according to nullable reference types' annotations
* Other small fixes in ical.net
* Small fixes n ical.net.tests
* NRT warninga left are with target `nestandard2.0`

Unit tests untouched except for those with `nullable enable`

### DIFF
--- a/Ical.Net/Serialization/SerializationContext.cs
+++ b/Ical.Net/Serialization/SerializationContext.cs
@@ -83,13 +83,13 @@ public class SerializationContext
         return null;
     }
 
-    public virtual object GetService(Type serviceType) => _mServiceProvider.GetService(serviceType);
+    public virtual object? GetService(Type serviceType) => _mServiceProvider.GetService(serviceType);
 
-    public virtual object GetService(string name) => _mServiceProvider.GetService(name);
+    public virtual object? GetService(string name) => _mServiceProvider.GetService(name);
 
-    public virtual T GetService<T>() => _mServiceProvider.GetService<T>();
+    public virtual T? GetService<T>() => _mServiceProvider.GetService<T>();
 
-    public virtual T GetService<T>(string name) => _mServiceProvider.GetService<T>(name);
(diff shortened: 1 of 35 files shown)


Ical.Net/Serialization/SerializationContext.cs
```diff
@@ -83,13 +83,13 @@ public virtual void Push(object? item)
         return null;
     }
 
-    public virtual object GetService(Type serviceType) => _mServiceProvider.GetService(serviceType);
+    public virtual object? GetService(Type serviceType) => _mServiceProvider.GetService(serviceType);
```
axunonb: @minichma This is in line with the current implementation, but I'd rather change the behavior to throw when a service cannot be found?

minichma: Agree. Actually I think we should get rid of this whole pseudo-DI stuff. It significantly increases complexity and decreases readability while not bringing much value. But for now I agree that throwing would be better.

axunonb: @minichma Completed - not only more simple but also brings a nice increase in coverage

minichma:
Wow, -37 branches, -17 partials, +12 hits - very nice! Will try to find time for a review later this week, ideally this evening.

minichma (approved):

axunonb:
Thanks for the quick review!

PR #785, opened 2025-05-07
Evaluation: Raise `EvaluationOutOfRangeException` if year 10k is hit during evaluation

### LINKED ISSUE #784: `CalendarEvent.GetOccurrences` without limiting the enumeration throws
**Describe the bug**
`CalendarEvent.GetOccurrences` throws if not `.TakeUntil` is applied to the enumerable:
```
System.ArgumentOutOfRangeException : The added or subtracted value results in an un-representable DateTime. (Parameter 'value')
```

**To Reproduce**
https://github.com/ical-org/ical.net/blob/d306ab599eef6b98e87a19240194e596fcad826e/Ical.Net.Tests/RecurrenceTests.cs#L127-L136
will fail, if the enumerable is converted `.ToList()` without the `.TakeUntil(new CalDateTime(2011, 1, 1))`

**Expected behavior**
Reaching the maximum supported `DateTime` should stop the enumeration.

- OS: Windows 11
- .NET version: .NET8.0
- ical.net version: current Main

minichma:
If recurrence evaluation produced an unrepresentable date/time, it used to throw an unspecific `ArgumentOutOfRangeException` or `OverflowException`. With this PR we throw a more specific `EvaluationoutOfRangeException`.

As an overflow could happen virtually anywhere throughout the evaluation code, the overflow handling is implemented in a quite coarse manner. I.e. all `ArgumentOutOfRangeException`s (raised by `DateOnly` et al) as well as `OverflowException`s (raised by NodaTime) are caught in central places. They could theoretically also be caused for different reasons, but, as patterns are validated upfront, it should be quite safe to assume they are caused due to y10k overflows.

Addresses #784

### DIFF
--- a/Ical.Net.Tests/RecurrenceTests.cs
+++ b/Ical.Net.Tests/RecurrenceTests.cs
@@ -3670,6 +3670,14 @@ END:VCALENDAR";
         Assert.That(occurrences, Has.Count.EqualTo(5));
     }
 
+    public enum RecurrenceTestExceptionStep
+    {
+        None,
+        Construction,
+        GetOccurrenceInvocation,
+        Enumeration,
+    }
+
     public class RecurrenceTestCase
     {
         public int LineNumber { get; set; }
@@ -3678,11 +3686,15 @@ END:VCALENDAR";
 
         public CalDateTime? DtStart { get; set; }
 
+        public Duration? Duration { get; internal set; }
+
         public CalDateTime? StartAt { get; set; }
 
         public IReadOnlyList<CalDateTime>? Instances { get; set; }
 
         public string? Exception { get; set; }
+
+        public RecurrenceTestExceptionStep? ExceptionStep { get; set; }
 
         public override string ToString()
             => $"Line {LineNumber}: {DtStart}, {RRule}";
@@ -3732,6 +3744,10 @@ END:VCALENDAR";
                     current.DtStart = new CalDateTime(val, "UTC");
                     break;
 
+                case "DURATION":
+                    current.Duration = Duration.Parse(val);
+                    break;
+
                 case "START-AT":
                     current.StartAt = new CalDateTime(val, "UTC");
                     break;
@@ -3742,6 +3758,11 @@ END:VCALENDAR";
 
                 case "EXCEPTION":
                     current.Exception = val;
+                    current.ExceptionStep ??= RecurrenceTestExceptionStep.Construction;
+                    break;
+
+                case "EXCEPTION-STEP":
+                    current.ExceptionStep = (RecurrenceTestExceptionStep)Enum.Parse(typeof(RecurrenceTestExceptionStep), val);
                     break;
             }
         }
@@ -3775,21 +3796,45 @@ END:VCALENDAR";
 
         // Start at midnight, UTC time
         evt.Start = testCase.DtStart!;
+        evt.Duration = testCase.Duration;
 
-        if (testCase.Exception != null)
+        Type LoadType(string name) =>
+            Type.GetType(name) ?? typeof(Calendar).Assembly.GetType(name) ?? throw new Exception();
+
+        var exceptionType = (testCase.Exception == null) ? null : LoadType(testCase.Exception);
+        IConstraint throwsConstraint = (exceptionType == null) ? Throws.InstanceOf(typeof(Exception)) : Throws.InstanceOf(exceptionType);
+
+        RecurrencePattern GetPattern() => new RecurrencePattern(testCase.RRule!);
+
+        if (testCase.ExceptionStep == RecurrenceTestExceptionStep.Construction)
         {
-            var exceptionType = Type.GetType(testCase.Exception)!;
-            Assert.Throws(exceptionType, () => new RecurrencePattern(testCase.RRule!));
+            Assert.That(() => GetPattern(), throwsConstraint);
             return;
         }
 
-        evt.RecurrenceRules.Add(new RecurrencePattern(testCase.RRule!));
+        evt.RecurrenceRules.Add(GetPattern());
 
-        var occurrences = evt.GetOccurrences(testCase.StartAt ?? new CalDateTime(DateTime.MinValue)).TakeUntil(new CalDateTime(DateTime.MaxValue))
-            .ToList();
+        IEnumerable<Occurrence> GetOccurrences() => evt.GetOccurrences(testCase.StartAt ?? null);
+
+        if (testCase.ExceptionStep == RecurrenceTestExceptionStep.GetOccurrenceInvocation)
+        {
+            Assert.That(() => GetOccurrences(), throwsConstraint);
+            return;
+        }
+
+        var occurrencesEnumerator = GetOccurrences();
+
+        List<Occurrence> EnumerateOccurrences() => occurrencesEnumerator.ToList();
+
+        if (testCase.ExceptionStep == RecurrenceTestExceptionStep.Enumeration)
+        {
+            Assert.That(() => EnumerateOccurrences(), throwsConstraint);
+            return;
+        }
+
+        var occurrences = EnumerateOccurrences();
 
         var startDates = occurrences.Select(x => x.Period.StartTime).ToList();
-
         Assert.That(startDates, Is.EqualTo(testCase.Instances));
     }
 
--- a/Ical.Net.Tests/RecurrenceTests.cs
+++ b/Ical.Net.Tests/RecurrenceTests.cs
@@ -3085,6 +3085,48 @@ public class RecurrenceTests
         });
     }
 
+    [Test]
+
+    // RRULE and RDATE both exceed y10k when converted to UTC, which is done when
+    // ordering the occurrences.
+    [TestCase("""
+            BEGIN:VCALENDAR
+            BEGIN:VEVENT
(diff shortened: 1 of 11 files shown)


minichma:
@axunonb I introduced the new `EvaluationOutOfRange` exception to be thrown in case of an overflow. Maybe a `System.OverflowException` would be just as good or even more natural. What do you think?

axunonb:
> Maybe a System.OverflowException

This is more generic for arithmetic operations, while `EvaluationOutOfRangeException` is more descriptive. I'd vote for the latter. Take your choice.

Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -17,33 +17,41 @@ protected void IncrementDate(ref CalDateTime dt, RecurrencePattern pattern, int
...
+                    break;
+                case FrequencyType.Yearly:
+                    dt = old.AddDays(-old.DayOfYear + 1).AddYears(interval);
+                    break;
+                // FIXME: use a more specific exception.
+                default:
+                    throw new Exception("FrequencyType.NONE cannot be evaluated. Please specify a FrequencyType before evaluating the recurrence.");
+            }
+        }
+        catch (ArgumentOutOfRangeException)
```
axunonb, line 51: Sticking to the ical.net definitions, isn't is about the Occurrence of the Event here?
How about sth. like
```cs
catch (ArgumentOutOfRangeException)
{
    // intentionally don't include the outer exception
    throw new EvaluationOutOfRangeException(
        $"""
         Evaluation aborted: An event occurrence starting at {dt} plus the next interval exceeds the maximum supported date/time value. 
         This may be caused by a recurrence pattern with a missing or excessive COUNT, or an UNTIL date missing or beyond the supported range.
         """
    );
}
```

Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -17,33 +17,41 @@ protected void IncrementDate(ref CalDateTime dt, RecurrencePattern pattern, int
...
+                case FrequencyType.Weekly:
+                    dt = DateUtil.AddWeeks(old, interval, pattern.FirstDayOfWeek);
+                    break;
+                case FrequencyType.Monthly:
+                    dt = old.AddDays(-old.Day + 1).AddMonths(interval);
+                    break;
+                case FrequencyType.Yearly:
+                    dt = old.AddDays(-old.DayOfYear + 1).AddYears(interval);
+                    break;
+                // FIXME: use a more specific exception.
```
axunonb, line 46: If `EvaluationException` instead of `EvaluationOutOfRangeException` would suffice, we could fix this with e.g.:
```cs
default:
    throw new EvaluationException(
        $"""
         Evaluation aborted: '{pattern.Frequency}' cannot be evaluated.
         Specify a valid FrequencyType before evaluating the recurrence.
         """
    );
```

Ical.Net/Evaluation/EventEvaluator.cs
```diff
@@ -54,47 +54,55 @@ public override IEnumerable<Period> Evaluate(CalDateTime referenceDate, CalDateT
...
+            return new Period(
+                start: period.StartTime,
+                end: period.Duration == null
+                    ? endTime
+                    : period.EffectiveEndTime);
+        }
+        catch (ArgumentOutOfRangeException)
+        {
+            // intentionally don't include the outer exception
+            throw new EvaluationOutOfRangeException("Evaluation was aborted because an event's end time is out of range. This commonly happens if an event has an unbounded RRULE or a very long duration. Consider applying the .TakeWhile() operator on the returned sequence.");
```
axunonb: See below: Maybe
```cs
throw new EvaluationException(
    """
    Evaluation aborted: Calculating the end time of the event occurrence resulted in an out-of-range value. 
    This commonly happens if the event has an unbounded RRULE or an excessively long duration.
    Consider applying the .TakeWhile() operator on the returned sequence to limit the evaluation.
    """
);
```

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -3085,6 +3085,48 @@ public void Test4()
...
+            BEGIN:VEVENT
+            DTSTART;TZID=America/Chicago:99991231T221000
+            END:VEVENT
+            END:VCALENDAR
+            """)]
+    public void Recurrence_WithOutOfBoundsUtc_ShouldFailWithCorrectException(string ical)
+    {
+        var cal = Calendar.Load(ical)!;
+        Assert.That(() => cal.GetOccurrences().ToList(), Throws.InstanceOf<EvaluationOutOfRangeException>());
+    }
```
axunonb, line 3128: 👍

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -180,13 +177,13 @@ private IEnumerable<CalDateTime> EnumerateDates(CalDateTime originalDate, CalDat
...
-                if (dateCount >= pattern.Count)
-                    break;
-
                 // UNTIL is applied outside of this method, after TZ conversion has been applied.
 
                 yield return candidate;
                 dateCount++;
+
+                if (dateCount >= pattern.Count)
+                    yield break;
```
axunonb, line 186: 👍

axunonb (commented):
See remarks regarding exception details. Hope this is not too 'academic' 
Sorry for breaking the build by merging Main, but accidentally removing `using System;`

Ical.Net/Evaluation/Evaluator.cs
minichma: > Sticking to the ical.net definitions, isn't is about the Occurrence of the Event here?

Are you referring to 'recurrence' vs 'occurrence' in the the message in the `default` case? Actually I didn't touch that one, but to my understanding recurrences are occurrences that are produced by recurring elements, i.e. by RRULEs, right? Acutally this method is only used with RRULEs, so this part of the message could be ok. But the whole method is not well suited in this class. It should either be moved to a utility class or to `RecurrencePatternEvaluator`. Happen to have related work prepared in a different, WIP PR.

Will try to improve the message in the `EvaluationOutOfRangeException`

Ical.Net/Evaluation/Evaluator.cs
minichma: Yes, absolutely. But didn't touch this one, so lets do it in a different PR please. Suggest to avoid multi-line messages though.

Ical.Net/Evaluation/EventEvaluator.cs
minichma: Nice, will improve.

Ical.Net/Evaluation/Evaluator.cs
minichma: > Evaluation aborted: An event occurrence starting at {dt} plus the next interval exceeds the maximum supported date/time value. This may be caused by a recurrence pattern with a missing or excessive COUNT, or an UNTIL date missing or beyond the supported range.

Hm, this is also not fully precise. Actually there isn't necessarily an occurrence at `dt`. I feel that the actual value also doesn't add too much value. Usually the problem here is just that people try to iterate an unbounded RRULE to its end, which isn't possible. The actual value, where the overflow happens, seems to be less relevant. Will try to find something.

minichma:
No worries. I don't think there could easily be a 'too academic' ;-)

Tried to improve the messages a little - should be good enough.

Ical.Net/Evaluation/Evaluator.cs
axunonb: Expected this answer 😉

Ical.Net/Evaluation/Evaluator.cs
axunonb: > Actually there isn't necessarily an occurrence

agree. 

The only intention regarding the messages is to make people see themselves what caused the exception and what they can do about it, instead of opening a ticket. Any wording serving this goal is fine.

Ical.Net/Evaluation/Evaluator.cs
minichma: As we're beyond our time budget all the time anyways, I don't appreciate extending the scope if not necessary. I know, my own review comments often suggest otherwise though.

axunonb (approved):
Very good, thank you!

Ical.Net/Evaluation/RecurrenceUtil.cs
```diff
@@ -102,9 +102,9 @@ public static IEnumerable<T> HandleEvaluationExceptions<T>(this IEnumerable<T> s
             // There shouldn't be other causes for this type of exceptions, as most validations of the pattern
             // itself are already done earlier, before doing the actual enumeration.
             // Intentionally don't include the outer exception as this most likely is not a technical but a usage error.
-            .Catch<T, ArgumentOutOfRangeException>(_ => throw new EvaluationOutOfRangeException("The maximum supported date/time value has been exceeded while evaluating occurrences. This is likely to happen in case of unbounded RRULEs. Consider applying .TakeWhile() on the returned sequence."))
+            .Catch<T, ArgumentOutOfRangeException>(_ => throw new EvaluationOutOfRangeException("An out-of-range value was encountered while evaluating occurrences. This commonly happens when trying to enumerate an unbounded RRULE to its end. Consider applying the .TakeWhile() operator."))
```
axunonb, line 105: Nice implementation

PR #786, opened 2025-05-07
Make NRT enabled the default setting for projects and CI builds

axunonb:
* Auto-replace `#nullable enable` per file with empty string
* Add `nullable` setting for all project files (Ical.Net: enable, Ical.Net.Tests: disable)
* Update GitHub Workflows nullability parameter: Now uses project settings
* Migrate Ical.Net.Benchmarks to NRT enabled

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -34,7 +34,7 @@ jobs:
       run: dotnet add ./Ical.Net.Tests/Ical.Net.Tests.csproj package AltCover
 
     - name: Build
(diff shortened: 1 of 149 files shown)


axunonb:
Couldn't yet find a way to get rid of the nullability warnings for `netstandard2.0` on ubuntu. They don't appear on Windows.

minichma (approved):
Congrats, with all the previous NRT PRs this was a significant piece work. Thanks for the nice commit structure of this one, made the review a breeze.

axunonb:
Thanks for the review. The NRT for the ical.net project is finally completed. Implementing NRT for the unit tests is less painful, and we can handle it one file at a time.

PR #789, opened 2025-05-10
Refactor handling of FREQ in recurrence pattern

### LINKED ISSUE #788: Handling of FREQ in recurrence pattern
### `FrequencyType.None`

The `FrequencyType.None` value should be removed because it is neither meaningful in the context of frequency enumeration nor does it have a valid string representation. Its presence could lead to confusion when users create a `RecurrencePattern`.

In turn `RecurrencePattern.Frequency` should either become nullable or - for simplicity - have a default value (e.g. `FrequencyType.Yearly`.

The only test for a valid `FrequencyType` is necessary in `RecurrencePatternSerializer` where an invalid input string should throw.

### Check for valid `RecurrencePattern.Frequency` in `Evaluator`

Currently we have a check for a valid `FrequencyType`, that originates in the existence of `FrequencyType.None`.
https://github.com/ical-org/ical.net/blob/ca0e6e452a7f34e8bac891b861291dd4fdc4eba8/Ical.Net/Evaluation/Evaluator.cs#L46-L48

However, Frequency should always be valid at this stage.
The `default` case should just trigger a `Debug.Fail` as a safeguard indicating a missing implementation.
There is no need to throw an Exception for users.

_Originally posted by @axunonb in https://github.com/ical-org/ical.net/pull/785#discussion_r2081344204_

axunonb:
Resolves #788

### DIFF
--- a/Ical.Net/DataTypes/RecurrencePattern.cs
+++ b/Ical.Net/DataTypes/RecurrencePattern.cs
@@ -52,11 +52,13 @@ public class RecurrencePattern : EncodableDataType
 
 
     /// <summary>
-    /// Specifies the interval between occurrences of the recurrence. The default value is 1.
-    /// <para/>
-    /// The value represents the number of units of time, as defined by the <see cref="Frequency"/> property, 
-    /// between each occurrence. For example, if <see cref="Frequency"/> is set to "Weekly" and Interval is 2, 
-    /// the recurrence will occur every 2 weeks.
+    /// The INTERVAL rule part contains a positive integer representing at
+    /// which intervals the recurrence rule repeats. The default value is
+    /// 1, meaning every second for a SECONDLY rule, every minute for a
+    /// MINUTELY rule, every hour for an HOURLY rule, every day for a
+    /// DAILY rule, every week for a WEEKLY rule, every month for a
+    /// MONTHLY rule, and every year for a YEARLY rule. For example,
+    /// within a DAILY rule, a value of 8 means every eight days.
     /// </summary>
     public int Interval
     {
--- a/Ical.Net/DataTypes/RecurrencePattern.cs
+++ b/Ical.Net/DataTypes/RecurrencePattern.cs
@@ -19,13 +19,25 @@ namespace Ical.Net.DataTypes;
 public class RecurrencePattern : EncodableDataType
 {
     private int? _interval;
+    private FrequencyType _frequency;
     private CalDateTime? _until;
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


Ical.Net/DataTypes/RecurrencePattern.cs
```diff
@@ -37,13 +44,19 @@ public CalDateTime? Until
...
+    /// </summary>
     public int? Count { get; set; }
 
+
     /// <summary>
-    /// Specifies how often the recurrence should repeat.
-    /// - 1 = every
-    /// - 2 = every second
-    /// - 3 = every third
+    /// Specifies the interval between occurrences of the recurrence. The default value is 1.
```
minichma: That's not fully precise, as due to the other rule parts, more or less occurrences could happen per interval. This is what the RFC says, maybe just copy?

> The INTERVAL rule part contains a positive integer representing at
      which intervals the recurrence rule repeats.  The default value is
      "1", meaning every second for a SECONDLY rule, every minute for a
      MINUTELY rule, every hour for an HOURLY rule, every day for a
      DAILY rule, every week for a WEEKLY rule, every month for a
      MONTHLY rule, and every year for a YEARLY rule.  For example,
      within a DAILY rule, a value of "8" means every eight days.

Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -43,9 +43,10 @@ protected void IncrementDate(ref CalDateTime dt, RecurrencePattern pattern, int
                 case FrequencyType.Yearly:
                     dt = old.AddDays(-old.DayOfYear + 1).AddYears(interval);
                     break;
-                // FIXME: use a more specific exception.
                 default:
-                    throw new Exception("FrequencyType.NONE cannot be evaluated. Please specify a FrequencyType before evaluating the recurrence.");
+                    // Frequency should always be valid at this stage.
+                    System.Diagnostics.Debug.Fail($"'{pattern.Frequency}' as RecurrencePattern.Frequency is not implemented.");
```
minichma, line 48: Not sure we should fail here. Users could easily set any int as frequency type, which would not be an internal error. I think throwing was not wrong.

minichma (dismissed):

Ical.Net/Evaluation/Evaluator.cs
axunonb: ... kike with `(FrequencyType) 123` - yes, you're right. Wouldn't checking for valid values in the setter of `RecurrencePattern.Frequency` be better?

minichma (approved):

PR #790, opened 2025-05-10
Remove unnessary null checks

axunonb:
Remove null checks having "Expression is always true or false according to NRT annotations" from Code Inspector.
Again, increasing coverage.


minichma (approved):
Nice to see, how NRT helps improving the code.

PR #793, opened 2025-05-12
Remove unnecessary `GroupedListEnumerator`

axunonb:
This also simplified the code IMHO.


minichma:
> This also simplified the code IMHO.

Absolutely. But do we need this at all? The only usage I can find is here:

https://github.com/ical-org/ical.net/blob/c7fd6867169c4b426d763575d75bc51da1a34ff3/Ical.Net/Collections/GroupedList.cs#L221

Couldn't we just implement that like so?

```cs
_lists.SelectMany(x => x).GetEnumerator();
```

axunonb:
> But do we need this at all

This kind of simplification really goes way to far 😂

minichma (approved):
Love it!

PR #794, opened 2025-05-12
Enable `CA1305` warnings and fix them

### LINKED ISSUE #792: Parsing fails in certain cultures
**Describe the bug**
Parsing of certain calendars fails if certain cultures are configured as current system culture.

**To Reproduce**

The following test case fails in the `ar-SA` and `he-IL` cases when trying to parse the negative BY value `-1`.

```cs
    [TestCase("en-US")] // English (United States)
    [TestCase("de-DE")] // German (Germany)

    // fails
    [TestCase("ar-SA")] // Arabic (Saudi Arabia)

    // fails
    [TestCase("he-IL")] // Hebrew (Israel)

    [TestCase("hi-IN")] // Hindi (India)
    [TestCase("zh-CN")] // Chinese (Simplified, China)
    [TestCase("ja-JP")] // Japanese (Japan)
    [TestCase("th-TH")] // Thai (Thailand)
    [TestCase("ru-RU")] // Russian (Russia)
    [TestCase("ko-KR")] // Korean (Korea)
    public void TestConvertToInt32WithNegativeNumberInDifferentCultures(string cultureStr)
    {
        var originalCulture = System.Globalization.CultureInfo.CurrentCulture;
        try
        {
            var culture = new System.Globalization.CultureInfo(cultureStr);
            System.Globalization.CultureInfo.CurrentCulture = culture;

            Assert.That(() => Calendar.Load("""
                BEGIN:VCALENDAR
                BEGIN:VEVENT
                DTSTART:20251231
                RRULE:FREQ=YEARLY;BYYEARDAY=-1
                END:VEVENT
                END:VCALENDAR
                """), Throws.Nothing);
        } finally
        {
            System.Globalization.CultureInfo.CurrentCulture = originalCulture;
        }
    }
```

- ical.net version: 5.0.0-pre.42

All parsing and formatting should be implemented in a culture-invariant manner. Consider enabling [CA1305](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1305) warnings.

axunonb:
* Add package `Microsoft.CodeAnalysis.NetAnalyzers` (for build only)
* Add `dotnet_diagnostic.CA1305.severity = warning` to .editorconfig
* Fix `CA1305` warning or disable with pragma where appropriate
* Add unit test as proposed in issue

Resolves #792

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,6 +18,8 @@ insert_final_newline = true
 csharp_new_line_before_members_in_anonymous_types = true
 
 [*.cs]
+#parsing and formatting should be implemented in a culture-invariant manner
+dotnet_diagnostic.CA1305.severity = warning
 
 #place catch statements on a new line
 csharp_new_line_before_catch = true
--- a/Ical.Net.Benchmarks/Ical.Net.Benchmarks.csproj
+++ b/Ical.Net.Benchmarks/Ical.Net.Benchmarks.csproj
@@ -7,6 +7,7 @@
     </PropertyGroup>
     <ItemGroup>
         <PackageReference Include="BenchmarkDotNet" Version="0.14.0" />
+        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="9.0.0" PrivateAssets="all" />
     </ItemGroup>
     <ItemGroup>
         <ProjectReference Include="..\Ical.Net\Ical.Net.csproj" />
--- a/Ical.Net.Tests/CalDateTimeTests.cs
+++ b/Ical.Net.Tests/CalDateTimeTests.cs
@@ -93,7 +93,7 @@ public class CalDateTimeTests
     [Test(Description = "A certain date/time value applied to different timezones should return the same UTC date/time")]
     public void SameDateTimeWithDifferentTzIdShouldReturnSameUtc()
     {
-        var someTime = DateTimeOffset.Parse("2018-05-21T11:35:00-04:00");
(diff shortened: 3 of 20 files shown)


minichma (dismissed):
Wow, this was fast. Just a note on string interpolation. In a work project we have some utility method like the following:

```cs
        public static string Invariant(IFormattable v)
            => (v == null) ? null : v.ToString(null, CultureInfo.InvariantCulture);
```

This can be used to make string interpolation culture invariant like so:

`Invariant($"{...}")`

Maybe an option instead of the pragmas.

axunonb:
>  public static string Invariant(IFormattable v)
>           => (v == null) ? null : v.ToString(null, CultureInfo.InvariantCulture);

Thanks for the hint.

We can also use the built-in `FormattableString.Invariant` e.g.
```cs
sb.Append(FormattableString.Invariant($"{sign * (ts.Hours)}H"));
```
directly. I was not aware of that.

Here the compiler automatically creates a `FormattableString` object instead of a string under the hood

```cs
FormattableString fs = $"{sign * (ts.Hours ?? 0)}H";
sb.Append(FormattableString.Invariant(fs));
```

From NET6.0 the right way seems to be (according to SonarCloud)
```string.Create(CultureInfo.InvariantCulture, $"{sign * (ts.Hours)}H");``` with less memory allocation
which unfortunately is not supported in netstandard2.x

minichma:
Very nice and good to know, haven't been aware of that either.

minichma (approved):

PR #796, opened 2025-05-15
Implement `CollectionExtensions.TakeWhileBefore` extensions

### LINKED ISSUE #787: Add `TakeBefore` extension methods for `IEnumerable<Period>` and `IEnumerable<Occurrence>` to public API
> I understand your concerns about exposing these extensions as part of the public API, but I think there are ways to address them while still making the functionality available to users in a clear and user-friendly way. Let me address your points one by one:
> 
> 1. **Complexity of the API**:  
>    While it's true that the extensions do more than one thing (null-checking and comparing order), this is a common pattern in utility methods. The null-check is a convenience feature that avoids boilerplate code for the user. From an API perspective, we could document the behavior clearly to ensure users understand what the method does. This would make the API more approachable, not less.
> 
> 2. **Name Suggestion**:  
>    I agree that the name `TakeUntil` can be misleading because it suggests inclusivity. A name like `TakeUntilExclusive` or even  shorter `TakeBefore` would better reflect the behavior.
> 
> 3. **Precondition of Ordered Sequences**:  
>    You're right that the method assumes the sequence is ordered. However, this is also true for `TakeWhile`. The same assumption applies. We can add a note in the XML documentation to make this explicit for `TakeBefore`, while we cannot for `TakeWhile`..
> 
> 4. **Comparison to `TakeWhile`**:  
>    While `TakeWhile` is straightforward, it requires users to write a lambda expression every time, which can be error-prone and repetitive. The proposed extensions encapsulate this logic, making the API more user-friendly and reducing boilerplate code.  
> 
> 5. **Documentation and Clarity**:  
>    I agree that exposing these methods requires clear documentation to explain their behavior and preconditions, as for any API. If we document that the method assumes an ordered sequence and is exclusive of `periodEnd`, I think it would be clear enough for users.
> 
> ### Proposal:
> To address your concerns while still making the functionality available:
> 1. Rename the methods to something more descriptive I'd prefer `TakeBefore` over `TakeUntilExclusive`.
> 2. Add XML documentation to explain the behavior, preconditions (e.g., ordered sequences), and examples of usage.
> 3. Consider exposing these methods as part of a utility class in the library (e.g., `Ical.Net.Utility.SequenceExtensions`) rather than directly on the core types.
> 
> The convenience the extensions provide outweighs the potential confusion, especially since similar assumptions exist in LINQ  `TakeWhile`.
> 
> Would this be a way to proceed? 

 _Originally posted by @axunonb in [#784](https://github.com/ical-org/ical.net/issues/784#issuecomment-2860280780)_

axunonb:
These are convenience methods for `Enumerable.TakeWhile{TSource}(IEnumerable{TSource}, Func{TSource, bool})`.

Resolves #787


Ical.Net/CollectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// This is a convenience method for <see cref="Enumerable.TakeWhile{TSource}(IEnumerable{TSource}, Func{TSource, bool})"/>.
+    /// But even with this method, the input sequence must be ordered according to the type's default equality comparer.
+    /// </summary>
+    /// <param name="sequence">The ordered sequence to be filtered.</param>
+    /// <param name="periodEnd">The exclusive end of the period to be used as a filter.</param>
+    /// <returns>
+    /// The elements of the sequence of occurrences to only include those that start before the specified period end.
+    /// </returns>
+    public static IEnumerable<Occurrence> TakeBefore(this IEnumerable<Occurrence> sequence, CalDateTime? periodEnd)
+        => (periodEnd == null) ? sequence : sequence.TakeWhile(p => p.Period.StartTime < periodEnd);
```
minichma: I think `periodEnd` shouldn't be nullable. In the test space I made it nullable to act as a c&p replacement for the previous periodEnd param and its used just internally (actually this was part of the reason not making it public). But in this case I feel its not quite natural allowing to pass a null value, making it a noop. I therefore suggest making it non-null. In the test projects we could keep an internal extension method that still allows passing null for simplicity.

Ical.Net/CollectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// The elements of the sequence of periods to only include those that start before the specified period end.
+    /// But even with this method, the input sequence must be ordered according to the type's default equality comparer.
+    /// </returns>
+    /// </summary>
+    /// <param name="sequence">The ordered sequence to be filtered.</param>
+    /// <param name="periodEnd">The exclusive end of the period to be used as a filter.</param>
+    /// <returns>
+    /// The elements of the sequence of periods to only include those that start before the specified period end.
+    /// </returns>
+    public static IEnumerable<Period> TakeBefore(this IEnumerable<Period> sequence, CalDateTime? periodEnd)
```
minichma: Regarding the naming: `TakeBefore` could be interpreted as filter (aka 'take all items from the sequence where the value is _before_ x). In this interpretation it wouldn't reflect the fact that 'Before' refers to the items in the sequence, which means _take those that are returned before hitting or exceeding the threshold). An alternative could be `TakeWhileBefore`, which more clearly suggests 'take all items from the sequence while the their date value is before x', but the name is a bit clumsy. Not sure.

minichma (dismissed):

Ical.Net/CollectionExtensions.cs
axunonb: Yes, really makes sense to think about the best term. Coming back to one of the initial ideas using like **`.GetOccurrences(...).WhileBefore(...)`** sounded quite natural, and it was short... Better?

Ical.Net/CollectionExtensions.cs
axunonb: Actually this was my first thought as well.: `null` would be a noop. Just like `"abc".ToString()` returning `this`.
However, users could then use `toDate` e.g. in a loop, without caring for nullability. See our own tests below.

If `toDate` was **not** nullable, we could (would) write the unit test `RecurrenceTests.EventOccurrenceTest(...)` like
```cs
var occurrences = toDate != null
    ? evt.GetOccurrences(fromDate).TakeBefore(toDate).ToList()
    : evt.GetOccurrences(fromDate).ToList();
```
without the need for an extension method just for this test.
What's better?

minichma: I'd prefer the not nullable, just like .TakeWhile or .Skip don't accept null. For our own use we can add the nullable version, like everyone can easily do. Outside of test cases I wouldn't expect it to be a common case that invocations need the nullable version. Either they want to limit or they don't I'd expect.

Ical.Net/CollectionExtensions.cs
minichma: Sounds better to me. It's not quite in line with linq naming though. Not sure it has to be in this case. On the other hand, if it is named like TakeWhileXxx it could be easier to find, because it pops up in auto completion when people are looking for linq operators. Anyone, any of the discussed names is good.

Ical.Net/CollectionExtensions.cs
axunonb: Then so be it.

Ical.Net/CollectionExtensions.cs
axunonb: "easier to find" is convincing for `.TakeWhileBefore(...)`, tnx

minichma (approved):

axunonb:
@minichma BTW quite nice how code quality has improved since October 2024: From 274 issues to 105 now.
https://sonarcloud.io/project/overview?id=ical-org_ical.net
Once you login to Sonar using your GitHub account I should be able to give you permissions to more insights.

minichma:
> @minichma BTW quite nice how code quality has improved since October 2024: From 274 issues to 105 now. https://sonarcloud.io/project/overview?id=ical-org_ical.net Once you login to Sonar using your GitHub account I should be able to give you permissions to more insights.

Very nice to see. It seems lots of the improvements coming from NRT - great work!

Issue #156, opened 2016-10-05
bug: Categories are output into multiple instances of CATEGORIES lines.


zijianhuang:
https://tools.ietf.org/html/rfc5545#section-3.8.1.2

The standard does not say "This property can be specified multiple times" in Conformance, though it does not forbid multiple instances either. MS Outlook is able to read an event with multiple CATEGORIES instances, but the export will output only 1 line.

It will be good the iCal component could conform to the standard when outputting ics though it could tolerate multiple instances of CATEGORIES.

I understand this is the legacy behavior of DDay.iCal component.

NRG-Drink:
@rianjs Can you please reopen this issue? The CATEGORIES property is printed multiple times on serialize and I cannot read it properly.
The test to solve this issue was Ignored some time ago. ([SymmetricSerializationTests.CategoriesTest](https://github.com/ical-org/ical.net/blame/95bbc53991009e64c208b827f3c89bd49680c6bc/Ical.Net.Tests/SymmetricSerializationTests.cs))

axunonb:
@NRG-Drink Confirm the behavior still exists. I think it's not too difficult to resolve. Could you help us with a PR fixing it?

Output is
```ics
CATEGORIES:Foo
CATEGORIES:Bar
CATEGORIES:Baz
```
instead of the RFC5545, s. 3.8.1.2 compliant output
```ics
CATEGORIES:Foo,Bar,Baz
```

EDIT:
After double-checking with RFC5545, the conclusion is:

### **RFC 5545, Section 3.6.1 (Event Component)**  (page 53)
> *"Properties defined by the iCalendar object MAY be specified more than once in a "VEVENT" calendar component."*  

Since **CATEGORIES** is a general property (not marked as singular in the property table), this means it **can appear multiple times** in a **VEVENT**.  

### **Key References:**  
1. **Section 3.6.1** (VEVENT) – Allows most properties to repeat unless restricted.  
2. **Section 3.8.1.2** (CATEGORIES) – Defines the property but doesn’t restrict multiplicity.  
3. **Section 3.2** (Property Occurrence) – By default, properties may repeat unless specified otherwise.  

### **Valid example**  
```ical
BEGIN:VEVENT
UID:12345
CATEGORIES:MEETING
CATEGORIES:WORK
...
END:VEVENT
```  

Here is the corresponding unit test:
[...]

PR #803, opened 2025-05-19
feat: Serialize multiple categories and resources to one line

NRG-Drink:
It's allowed to have `CATEGORIES` property multiple times in an event but it's not that pretty and some other programs will have trouble reading such events. So it's more a feature than a bug (#156).
Here is a quick fix to display multiple categories in one line.

### DIFF
--- a/Ical.Net.Tests/SymmetricSerializationTests.cs
+++ b/Ical.Net.Tests/SymmetricSerializationTests.cs
@@ -238,17 +238,19 @@ public class SymmetricSerializationTests
         yield return new TestCaseData("\\\\uncPath\\to\\resource.txt", new Uri("\\\\uncPath\\to\\resource.txt")).SetName("UNC path URL");
     }
 
-    [Test]
-    public void CategoriesTest()
+    [TestCase("Foo", "Bar", "Baz")]
+    [TestCase("Hello", "World", null)]
+    public void CategoriesTest(string cat1, string cat2, string cat3)
     {
         var vEvent = GetSimpleEvent();
-        vEvent.Categories = new List<string> { "Foo", "Bar", "Baz" };
+        vEvent.Categories = [cat1, cat2, cat3];
         var c = new Calendar();
         c.Events.Add(vEvent);
 
         var serialized = SerializeToString(c);
         var categoriesCount = Regex.Matches(serialized, "CATEGORIES").Count;
         Assert.That(categoriesCount, Is.EqualTo(1));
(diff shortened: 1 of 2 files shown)


Ical.Net.Tests/SymmetricSerializationTests.cs
```diff
@@ -238,7 +238,7 @@ public static IEnumerable UriAttachment_TestCases()
         yield return new TestCaseData("\\\\uncPath\\to\\resource.txt", new Uri("\\\\uncPath\\to\\resource.txt")).SetName("UNC path URL");
     }
 
-    [Test, Ignore("TODO: Fix CATEGORIES multiple serializations")]
+    [Test]
```
axunonb: Maybe assure also that the list contains expected Categories?
```cs
Assert.That(serialized, Does.Contain("CATEGORIES:Foo,Bar,Baz"));
```

Ical.Net/Serialization/PropertySerializer.cs
```diff
@@ -36,15 +36,35 @@ public PropertySerializer(SerializationContext ctx) : base(ctx) { }
         // the property and parameter values
         var sf = GetService<ISerializerFactory>();
 
+        // TODO: Exhaust this list with all properties which can be displayed in one line.
```
axunonb, line 39: Remove the TODO and for now add RESOURCES, as another frequent and simple use case?

axunonb:
Thanks, very good! See comments if your time allows.

Ical.Net.Tests/SymmetricSerializationTests.cs
NRG-Drink: I had this first but was not sure if the sequence matters or not.
The test is now extended as suggested and with another test-case (including null).

axunonb (approved):
That was fast, great! Thanks for your support.
Would you like to join the ical.net party?

NRG-Drink:
> That was fast, great! Thanks for your support. Would you like to join the ical.net party?

You were even faster. I wasn't done and there is an error in the test 😅
btw. is it possible that ATTENDEE is broken? I tried to enable oneline for attendees too but it only performs a ToString on it and not serializing it properly. I think I will leave it for the moment.

NRG-Drink:
> That was fast, great! Thanks for your support. Would you like to join the ical.net party?

I can try to help out on smaller tasks if it helps 😊

axunonb:
> btw. is it possible that ATTENDEE is broken? I tried to enable oneline for attendees

`Attendee` requires its own `AttendeeSerializer`, as it's more complex than e.g. CATEGORIES. That's why.

> You were even faster. 

Should have awaited your review request, right

PR #804, opened 2025-05-19
test: refine category and resource online test (+fix resource test)

NRG-Drink:
- Fix resources test (test on resource, not on categories)
- Remodel test structure


axunonb (approved):
Nice refactoring of the tests, tnx

PR #805, opened 2025-05-20
feat: Add evaluation layer to `CalDateTime`

### LINKED ISSUE #677: Clean-up redundant implicit/arithmetic operators in CalDateTime
`Subtract()` => `Add()` should be enough
`Add()` overloads for hour, minute, second et al could easily use `Add(TimeSpan)`
`implicit operator CalDateTime(DateTime left)` is not too helpful anymore
See also [this discussion](https://github.com/ical-org/ical.net/discussions/606#discussioncomment-11867080)

### LINKED ISSUE #737: Incorrect timezone calculation if short recurrence falls into an ambiguous local time
**Describe the bug**

**To Reproduce**

```ics
DURATION:PT45M
```

The expected outcome would be

```
20201024T023000/PT45M
20201025T023000/PT45M
```

```
```

axunonb:
Add evaluation layer to `CalDateTime` (WIP!)

* Introduced `CalDateTimeExtensions` for extension methods on `CalDateTime`
* Moved `CalDateTime` arithmetic and conversion methods to `CalDateTimeExtensions`
* Adjusted related classes with new references

Note: 
* `CalDateTimeExtensions.ToTimeZone` links the input `CalDateTime` with its `ZonedDateTime` representation, using a cache with weak references to `CalDateTime`. The cache is implemented with `ConditionalWeakTable<CalDateTime, ZonedDateTime>`.
* This way, once a `CalDateTime` is resolved to a `ZonedDateTime`, the `ZonedDateTime` will be used for further conversions.

Why cache `ZonedDateTime`?
* If a `CalDateTime` is instantiated from a lenient timezone conversion, this leniently determined value must not be converted again. 
* Lenient conversions resolve non-existing or ambiguous times caused by DST transitions.

Resolves #737
Resolves #677

### DIFF
--- a/Ical.Net.Benchmarks/ApplicationWorkflows.cs
+++ b/Ical.Net.Benchmarks/ApplicationWorkflows.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Ical.Net.Evaluation;
 
 namespace Ical.Net.Benchmarks;
 
--- a/Ical.Net.Benchmarks/CalDateTimePerfTests.cs
+++ b/Ical.Net.Benchmarks/CalDateTimePerfTests.cs
@@ -6,6 +6,7 @@
 using BenchmarkDotNet.Attributes;
 using Ical.Net.DataTypes;
 using System;
+using Ical.Net.Evaluation;
 
 namespace Ical.Net.Benchmarks;
 
--- a/Ical.Net.Benchmarks/OccurencePerfTests.cs
+++ b/Ical.Net.Benchmarks/OccurencePerfTests.cs
@@ -9,6 +9,7 @@ using Ical.Net.DataTypes;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Ical.Net.Evaluation;
 
 namespace Ical.Net.Benchmarks;
 
--- a/Ical.Net.Tests/CalDateTimeTests.cs
+++ b/Ical.Net.Tests/CalDateTimeTests.cs
@@ -12,6 +12,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using Ical.Net.Evaluation;
 
 namespace Ical.Net.Tests;
 
@@ -55,10 +56,10 @@ public class CalDateTimeTests
     [Test, TestCaseSource(nameof(ToTimeZoneTestCases))]
     public void ToTimeZoneTests(CalendarEvent calendarEvent, string targetTimeZone)
     {
-        var startAsUtc = calendarEvent.Start.AsUtc;
+        var startAsUtc = calendarEvent.Start!.AsUtc();
 
         var convertedStart = calendarEvent.Start.ToTimeZone(targetTimeZone);
-        var convertedAsUtc = convertedStart.AsUtc;
+        var convertedAsUtc = convertedStart.AsUtc();
 
         Assert.That(convertedAsUtc, Is.EqualTo(startAsUtc));
     }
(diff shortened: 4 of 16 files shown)


axunonb:
@minichma As long as we have `CalDateTime ToTimeZone` for the conversion this might be a simple fix for #737

minichma:
@axunonb Not sure we should track the precise instant in CalDateTime. I.e. `CalDateTime` represents values from the ICAL format, which just doesn't provide any means of resolving this ambiguity. By tracking the instant (and the exact time zone object), we lose fundamental properties like equality after a serialization/deserialization round trip. I think to address #737 we should either represent recurrences via a data type other than `Period`, i.e. one that allows unambiguously tracking the end date (e.g. by specifying a TimeSpan rather than a end). Alternatively we could be waiting for what we discussed regarding v6, where we could consider separating the model layer from the evaluation layer. In the model layer CalDateTime would represent an ICAL date/time value, but in the evaluation layer we could work with ZonedDateTime. I feel resolving #805 is not worth this change.

axunonb:
> In the model layer CalDateTime would represent an ICAL date/time value, but in the evaluation layer we could work with ZonedDateTime.

If your time allows, I'd be interested in your opinion on this kind of "evaluation layer" which comes as extension methods for `CalDateTime`. FYI: Over all unit tests, were getting about 40% cache hits for `CalDateTime` to `ZonedDateTime`, and a bit more for `GetOccurrences` only.

TBD:
* Part of the operators make use of the  `CalDateTimeExtensions` - move to the extensions?
* Remove `IFormattable` from `CalDateTime`? Or just remove the zone-aware `ZonedDateTime`-aware implementation?

> recurrences via a data type other than `Period`

Needs to be covered separately.

minichma:
@axunonb I love all the evaluation/arithmetics code being moved out of CalDateTime into extension methods, very nice! I'm not convinced by the caching code though for several reasons.

The outcome is very hard to reason upon. Consider the following test case:

```cs
    [Test]
    public void CalDateTime_AmbiguousTime_ResolvedCorrectly2()
    {
        var dt1 = new CalDateTime(2024, 10, 27, 1, 30, 0, "UTC").ToTimeZone("CET");
        var dt2 = new CalDateTime(2024, 10, 27, 2, 30, 0, "CET");

        // This is a fundamental assumption
        Assert.That(dt1 == dt2, Is.EqualTo(dt1.AsUtc() == dt2.AsUtc()));
    }
```

It's not quite obvious that this will fail. Or this one:

```cs
    [Test]
    public void CalDateTime_AmbiguousTime_ResolvedCorrectly()
    {
        TimeZoneResolvers.TimeZoneResolver = _ => NodaTime.DateTimeZoneProviders.Tzdb.GetZoneOrNull("UTC");

        var dt1 = new CalDateTime(2024, 10, 27, 1, 30, 0, "CET");
        var dt2 = new CalDateTime(2024, 10, 27, 1, 30, 0, "CET");
        var _ = dt1.AsUtc();

        TimeZoneResolvers.TimeZoneResolver = TimeZoneResolvers.Default;

        Assert.That(dt1, Is.EqualTo(dt2));
        Assert.That(dt1.AsUtc(), Is.EqualTo(dt2.AsUtc()));
    }
```

Its not obvious that the first assertion will succeed but the second will fail (clearly this is not a typical case, it rather shows the problems arising from `TimeZoneResolvers.TimeZoneResolver` being a global static). 

By attaching the converted value via a cache does still attach it so the problems as outlined in my previous comment still exist but with the external cache its even harder to understand what's happening.

Regarding the serialization round-trip, the following test will still fail:

```cs
    [Test]
    public void CalDateTime_AmbiguousTime_ResolvedCorrectly3()
    {
        var dt1 = new CalDateTime(2024, 10, 27, 1, 30, 0, "UTC").ToTimeZone("CET");

        var serializer = new DateTimeSerializer();
        var stream = new System.IO.MemoryStream();
        serializer.Serialize(dt1, stream, System.Text.Encoding.UTF8);
        stream.Position = 0;
        var restored = ((CalDateTime)serializer.Deserialize(stream, System.Text.Encoding.UTF8)).ToTimeZone("CET");

        Assert.That(dt1.AsUtc(), Is.EqualTo(restored.AsUtc()));
    }
```

The evaluation layer as I was thinking about it would be a completely independent set of classes. They might be allowed to reference the model layer but not the other way round. The model layer would just represent the ICAL model as it is but without any evaluation/arithmetic operations on it, including adding, converting, ... Only after conversion to the evaluation layer all the magic would be available. The date-time classes there would probably operate on something like ZonedDateTime, including the concrete instant they represent. Then the ambiguity would be gone.

However, in the meantime it would be quite straight-forward to return something else than Period from GetOccurrences() et al. I.e. we could easily introduce a new type that doesn't contain an end time but only a TimeSpan representing the exact duration, instead. Then there wouldn't be any ambiguity either.

axunonb:
Closed in favor of #809

PR #808, opened 2025-05-25
Determine `Occurrence.Period` by `Duration` only

### LINKED ISSUE #737: Incorrect timezone calculation if short recurrence falls into an ambiguous local time

axunonb:
* Update `EventEvaluator` to use the start time with `Duration` instead of end time to create a `Period` for `Occurrence`
* `Period`: Trying to create a `Period` by using a date-only `CalDateTime` and `Duration.HasTime` will throw
* Modify `.ics` files of unit tests to either contain `DTEND` or `DURATION` (never both) to get unambiguous test results
* Modify unit tests to use `Period.EffectiveEndTime` instead of `Period.EndTime`
* Add unit test for issue #737

Resolves #737


Ical.Net/DataTypes/Period.cs
```diff
@@ -99,6 +99,9 @@ public Period(CalDateTime start, Duration duration)
         if (duration.Sign < 0)
             throw new ArgumentException($"Duration ({duration}) must be greater than or equal to zero.", nameof(duration));
 
+        if (!start.HasTime && duration.HasTime)
+            throw new ArgumentException($"Exact Duration '{duration}' cannot be added to date-only value '{start}'");
```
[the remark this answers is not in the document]
minichma: Related to the model vs evaluation discussion: As this is (at least also) a model class, I feel that we should be careful in being too strict with input validation. As long as its syntactically correct, I think we should rather accept it, even if its not fully in line with the RFC. This is because I believe these kinds of inconsistent data do exist in the wild. And while it might not be possible to reasonably do certain kinds of evaluation on them, they can perfectly be serialized, deserialized, etc. Thinking of my own projects, which involves customer-specified ICALs, I'm not sure, all data will be fully RFC conformant. Actually I do know, that certain data isn't. So by becoming more strict with input validation can easily prevent such projects from being upgraded to v5. In the end its not always easy to properly tweak existing user-specified data.

So long story short: I think I'd rather avoid this additional check.

Ical.Net/Evaluation/EventEvaluator.cs
```diff
@@ -52,52 +52,38 @@ public override IEnumerable<Period> Evaluate(CalDateTime referenceDate, CalDateT
...
+                duration: duration.Value);
 
-            // Return the Period object with the calculated end time.
-            // Only EndTime is relevant for further processing,
-            // so we have to set it.
-            // If the period duration is not null here, it is an RDATE period
-            // and has priority over the calculated end time.
+            // Ensure that adding the duration to the start time
+            // results in a representable date/time
+            _ = newPeriod.EffectiveEndTime;
```
minichma: In which scenario would we get a non-representable date/time. And if we did, why would we throw here? Wouldn't it be one of the strengths of this approach that we can let the user choose, how to deal with this case? Maybe they don't need the EndTime at all, so why throw then?

minichma (dismissed):
Very nice!

Ical.Net/Evaluation/EventEvaluator.cs
axunonb: Can be caused by unlimited GetOccurrences, see #785 
But yes, using Duration makes this unnecessary (would throw later) and break the unit test introduced in #785

minichma: oh, I see, right. Well, but then still, why not let the exception be thrown later? I think, there's no need to throw here.

axunonb: Agree

axunonb: Hm, then we get an <System.ArgumentOutOfRangeException: The added or subtracted value results in an un-representable DateTime. (Parameter 'value') when accessing Period.EffectiveEndTime. ~~It wouldn't make sense to replace it with `EvaluationOutOfRangeException` inside the `Period`.~~ `EvaluationOutOfRangeException` is even better than`System.ArgumentOutOfRangeException`.

PR #809, opened 2025-05-29
Add evaluation classes for `CalDateTime`

### LINKED ISSUE #677: Clean-up redundant implicit/arithmetic operators in CalDateTime

### LINKED ISSUE #737: Incorrect timezone calculation if short recurrence falls into an ambiguous local time

axunonb:
* Added **internal** clss `CalDateTimeEvaluator`
* Added lightweight **internal** `CalDateTimeZoned` which is returned from `CalDateTimeEvaluator` methods
* `CalDateTime` changes with low impact on existing code:
   * Mark all comparison operators as obsolete with error
   * Mark `Utc` property as obsolete with error
   * Move all arithmetic methods and comparison methods to new `CalDateTimeExtensions`. The extension methods use the internal `CalDateTimeZoned` behind the scene.
* Refactor `Period` to be usable with `CalDateTimeZoned` internally
   * Update `EventEvaluator` to use `CalDateTimeZoned` with `Period` (no change to the public API
   * Add unit test AmbiguousLocalTime_WithShortDurationOfRecurrence which resolves #737
* Added `Microsoft.Bcl.HashCode` for simplified HashCode generation with `netstandard2.x`

Resolves #677
Resolves #737


axunonb:
~~@minichma  Currently, `CalDateTime` still contains the comparison and arithmetic methods (flagged with a comment), that should eventually be removed.
Should we keep them for v5, remove, or flag as obsolete? What do you think?
`CalDateTimeZoned` could also be returned with `GetOccurrences` etc.
(Also, the unit tests are not yet adopted.)~~

minichma:
> @minichma Currently, `CalDateTime` still contains the comparison and arithmetic methods (flagged with a comment), that should eventually be removed. Should we keep them for v5, remove, or flag as obsolete? What do you think? `CalDateTimeZoned` could also be returned with `GetOccurrences` etc. (Also, the unit tests are not yet adopted.)

If we have a new type particularly for doing arithmetics, etc., then I feel it would be confusing to have all the methods also available on `CalDateTime`. But removing them from there will be a big deal, because `CalDateTimeZoned` would have to be introduced everywhere.

Not sure I'd introduce a new type (i.e. `CalDateTimeZoned`) at all. Wouldn't it be an option just to use NodaTime's `ZonedDateTime` instead. Problems are that it can't represent a date-only or a floating time value. Floating time could be represented by introducing a dedicated timezone class for that purpose. Regarding date-only, could it be feasible that we don't represent those as date-only during evaluation at all? Not sure. Unfortunately I don't have the time right now to go into more detail. Anyhow, quite a challenge for v5, should we want to release it any time soon. I guess it would be more reasonable to postpone this whole topic to v6.

axunonb:
> If we have a new type particularly for doing arithmetics, etc., then I feel it would be confusing to have all the methods also available on CalDateTime.

~~The methods in `CalDateTime` would be marked as obsolete with a hint to `CalDateTimeZoned` for easier migration.~~
`CalDateTimeZoned` can be internal with the latest commits. `CalDateTimeExtensions` be for zoned arithmetics and comparisons, as you suggested.

PR #812, opened 2025-06-06
fix: `Period.CollidesWith`

### LINKED ISSUE #811: iCalendar semantics in `Period.CollidesWith`
There seem to be 2 cases where `Period.CollidesWith` does not implement iCalendar semantics correctly. The test cases fail.

```cs
[Test, TestCaseSource(nameof(CollidesWithTestCases))]
public void CollidesWith_VariousCases(Period period1, Period period2, bool expected)
{
    Assert.That(period1.CollidesWith(period2), Is.EqualTo(expected));
    Assert.That(period2.CollidesWith(period1), Is.EqualTo(expected));
}

private static IEnumerable CollidesWithTestCases
{
    get
    {
        // Touching periods (end of one is start of another, exclusive end)
        yield return new TestCaseData(
            new Period(new CalDateTime(2025, 1, 1, 0, 0, 0), Duration.FromHours(1)),
            new Period(new CalDateTime(2025, 1, 1, 1, 0, 0), Duration.FromHours(1)),
            false
        ).SetName("Touching: period1 ends when period2 starts (no overlap)");

        // Open-ended period (no end or duration) collides only if start is the same
        yield return new TestCaseData(
            new Period(new CalDateTime(2025, 1, 1, 0, 0, 0)),
            new Period(new CalDateTime(2025, 1, 1, 0, 0, 0)),
            true
        ).SetName("OpenEnded: open-ended periods with same start collide");
    }
}
```
The description (expected behavior) should read:
**CollidesWith(Period period)**
Determines whether this period overlaps (collides) with another period, using standard inclusive start and exclusive end semantics, and handles open-ended periods according to iCalendar semantics.
* **For periods with both start and end:** The periods collide if the start of one is before the end of the other and vice versa. This means intervals that only touch at the end/start boundary (e.g., the end of one is the start of the other) do not collide.
* **For open-ended periods** (no end or duration): Two such periods only collide if their start times are equal.
* **If only one period is open-ended:** The periods collide if the open-ended period's start is before the other's end, and the other period's start is before the open-ended period's start.

[Edit] No end or duration is treated as a zero-duration event. But can two same instants in time "overlap"?
if End and Duration are null, should EffectiveDuration be Duration.Zero instead of null?

@minichma What do you think?

Version: v5.0.0-pre.43

axunonb:
* Add checks for edge cases of periods to `Period.CollidesWith`
   * periods with `Duration.Zero` do not collide
   * periods with `EffectiveDuration == null` (only `StartTime` is set) throw as they are not considered as periods (should use `Contains(CalDateTime)` instead
   * contiguous periods are not evaluated as colliding any more
* Remove redundant null check in `Period.EffectiveEndTime`
* Enhance unit test coverage for `Period` and `FreeBusy`

Resolves #811


minichma:
@axunonb Sorry for not being able to review any earlier. Great increase of test coverage! The only detail to consider (not sure this is the case anyways) would be that `Period` instances that represent an `RDATE` of type `DATE` or `DATE-TIME` might not have a duration set. Not sure that could be a problem in the current implementation.

axunonb:
> Not sure that could be a problem in the current implementation.

In the `FreeBusyStatus` context I don't see a problem. With `RDATE` the method is not in use internally, but right, this could cause an issue.

> Great increase of test coverage

Still too low over ical.net, unfortunately, putting any code changes at risk for unintentionally breaking things.

PR #820, opened 2025-06-19
Fix: Exception for Blazor WebAssembly using FileVersionInfo

### LINKED ISSUE #819: Exception for Blazor WebAssembly using FileVersionInfo
**Describe the bug**
 System.TypeInitializationException: The type initializer for 'Ical.Net.LibraryMetadata' threw an exception.
   ---> System.PlatformNotSupportedException: System.Diagnostics.FileVersionInfo is not supported on this platform.

**To Reproduce**
Build a Blazor WebAssembly application using ical.net

**Expected behavior**
No exception thrown.

- OS: [macOS 15.5]
- .NET version: [ .NET 9.0.6]
- ical.net version: [e.g. 5.0.0]

**Last working version**
ical.net 4.3.1

b-straub:
Correct fallback solution when 'FileVersionInfo.GetVersionInfo' is not available on a specific platform. 
Synchronize versions.

Resolves #819

### DIFF
--- a/Ical.Net/Constants.cs
+++ b/Ical.Net/Constants.cs
@@ -254,12 +254,21 @@ public static class LibraryMetadata
 
     private static string GetAssemblyVersion()
     {
+        string? version;
         var assembly = typeof(LibraryMetadata).Assembly;
-        var fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
+
         // Prefer the file version, but fall back to the assembly version if it's not available.
-        return fileVersionInfo.FileVersion
-               ?? assembly.GetName().Version?.ToString() // will only change for major versions
-               ?? "1.0.0.0";
+        try
+        {
+            var fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
+            version = fileVersionInfo.FileVersion;
+        }
+        catch (Exception)
+        {
+            version = assembly.GetName().Version?.ToString();
+        }
+
+        return version ?? "1.0.0.0";
     }
 }
 
--- a/Ical.Net/Constants.cs
+++ b/Ical.Net/Constants.cs
@@ -268,7 +268,7 @@ public static class LibraryMetadata
             version = assembly.GetName().Version?.ToString();
         }
 
-        return version ?? "1.0.0.0";
+        return version ?? "2.0";
     }
 }
 
--- a/Ical.Net/Constants.cs
+++ b/Ical.Net/Constants.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Reflection;
 
 namespace Ical.Net;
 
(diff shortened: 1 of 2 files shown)


Ical.Net/Constants.cs
```diff
@@ -254,12 +254,21 @@ public static class LibraryMetadata
...
     {
+        string? version;
         var assembly = typeof(LibraryMetadata).Assembly;
-        var fileVersionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
+
         // Prefer the file version, but fall back to the assembly version if it's not available.
-        return fileVersionInfo.FileVersion
-               ?? assembly.GetName().Version?.ToString() // will only change for major versions
-               ?? "1.0.0.0";
+        try
```
axunonb: The code introduced in [this GitHub comment](https://github.com/ical-org/ical.net/issues/819#issuecomment-2985721326) doesn't work for WebAssembly? The advantage was that we wouldn't need the try/catch. In both cases, caching the value would be advantageous.

Ical.Net/Ical.Net.csproj
```diff
@@ -5,6 +5,7 @@
         <AssemblyOriginatorKeyFile>..\IcalNetStrongnameKey.snk</AssemblyOriginatorKeyFile>
         <nullable>enable</nullable>
         <LangVersion>latest</LangVersion>
+        <AssemblyVersion>2.0</AssemblyVersion>
```
axunonb: All version attributes get set during publishing with the CI workflow.
Mixing up 2.0 with the iCalendar version defined in RFC 5545 Section 3.7.4?

Ical.Net/Constants.cs
b-straub: Sorry, was too fast with my fix. Working indeed, so you can throw away the pull request and replace with the suggested code.

Ical.Net/Ical.Net.csproj
b-straub: Understood. Anyway obsolete see comment above.

Ical.Net/Constants.cs
```diff
@@ -268,7 +268,7 @@ private static string GetAssemblyVersion()
             version = assembly.GetName().Version?.ToString();
         }
 
-        return version ?? "1.0.0.0";
+        return version ?? "2.0";
```
axunonb: see [below](https://github.com/ical-org/ical.net/pull/820/files/ca3826f3203bf585e75c1c27b3df435230cd23a6..a7d964401e86dcb6091ec470bd0e7e2b76a598ca#r2156289462) regarding "2.0"

axunonb:
Appreciate your feedback and PR. Why not update and we're done?
Hoping this is the only issue re WebAssembly

b-straub:
> Appreciate your feedback and PR. Why not update and we're done? Hoping this is the only issue re WebAssembly

Done! I have a heavy solution to check. Will report when I find something else.

axunonb (approved):
LGTM, tnx

PR #822, opened 2025-06-19
Add implicit conversion of DateTime/DateOnly to CalDateTime

### LINKED ISSUE #821: Implicit Conversion Operators for CalDateTime (DateTime -> CalDateTime, DateOnly -> CalDateTime)
v5.0.0 replaced the IDateTime interface with the new CalDateTime. This works fine, and migrating from v4 is fairly straightforward. Any DateTime now has to be constructed as

```csharp
new CalDateTime(DateTime)
```

I would suggest to add an implicit operator to the CalDateTime class, so that DateTime & DateOnly can be implicitly casted to CalDateTime, as that would require less code changes (since the DateTime & DateOnly can now be automatically converted)

NRG-Drink:
- Add implicit conversion from DateTime to CalDateTime
- Add implicit conversion from DateOnly to CalDateTime

closes #821

### DIFF
--- a/Ical.Net.Tests/CalDateTimeTests.cs
+++ b/Ical.Net.Tests/CalDateTimeTests.cs
@@ -361,12 +361,47 @@ public class CalDateTimeTests
         ];
 
     [Test, TestCaseSource(nameof(CalDateTime_FromDateTime_HandlesKindCorrectlyTestCases))]
-
-
     public void CalDateTime_FromDateTime_HandlesKindCorrectly(DateTimeKind kind, IResolveConstraint constraint)
     {
         var dt = new DateTime(2024, 12, 30, 10, 44, 50, kind);
 
         Assert.That(() => new CalDateTime(dt), constraint);
+    }
+
+    private static TestCaseData[] CalDateTimeImplicitFromDateTime_TestCastes => [
+        new TestCaseData(DateTimeKind.Unspecified, Is.EqualTo(new CalDateTime(2024, 12, 30, 10, 44, 50, null))),
+        new TestCaseData(DateTimeKind.Utc, Is.EqualTo(new CalDateTime(2024, 12, 30, 10, 44, 50, "UTC"))),
+        new TestCaseData(DateTimeKind.Local, Throws.ArgumentException),
+    ];
+
+    [Test, TestCaseSource(nameof(CalDateTimeImplicitFromDateTime_TestCastes))]
+    public void CalDateTimeImplicitFromDateTime(DateTimeKind kind, IResolveConstraint constraint)
+    {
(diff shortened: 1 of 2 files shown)


minichma:
@NRG-Drink The implicit conversion operators have intentionally been removed. The reason is that there is no obvious conversion between the two. E.g. one has a TZ the other hasn't. And then there's the problem with `DateTimeKind`, which is a complete story of its own. By offering an implicit conversion, it will be much easier for users to incorrectly confuse DateTime and CalDateTime. It might also be worth comparing NodaTime's appraoch. There, no implicit conversion exists between Instant, ZonedDateTime, LocalDateTime, etc. and System.DateTime either.

So I suggest to *not* introduce an implicit conversion. I'd also discourage an explicit conversion. To simplify things, I could imagine publishing `DateUtil.AsCalDateTime()`. This would require moving the method to a different, public type, probably introducing a new one like `DateTimeExtensions`. This would also allow using it with null-coalescing (`dt?.AsCalDateTime()`). (Maybe even consider renaming the method from `AsCalDateTime` to `ToCalDateTime`).

axunonb (changes requested):
Referring to the https://github.com/ical-org/ical.net/pull/822#issuecomment-2990225115: 
There are use cases where implicit operators are great. For `CalDateTime` vs. `DateTime` they can be quite confusing.

So what @minichma proposes could be viable option to simplify conversion.
Currently we use the method from `DateUtil` 
```cs
public static CalDateTime AsCalDateTime(this DateTime t)
    => new CalDateTime(t);
```
in our unit tests, and yes, we could add it to the public API e.g. in a `static` class `DateTimeExtensions` with  
```cs
public static CalDateTime ToCalDateTime(this DateTime t)
    => new CalDateTime(t);
public static CalDateTime ToCalDateTime(this DateOnly t)
    => new CalDateTime(t);
```
and add the corresponding xmldoc form the `CalDateTime` CTORs. As opposed to implicit operator we can explain how the conversion works.

Is this case it would be good to migrate `DateUtil.AsCalDateTime` to the new extension method(s).

What do you think?

NRG-Drink:
Thanks for the comments @minichma @axunonb .
Personally, I'm not a big fan of static util classes, specially when we have a constructor that does the same thing. So it would be a discussion about how to construct an object and there the construction should win imo.
```cs
new CalDateTime(dt);
dt.AsCalDateTime();
DateUtil.AsCalDateTime(dt);
```

A more straight forward way is to use the `implicit` operator but I can understand the point that it can confuse some users with the time-zone and the exact behavior. It has to be practical for the user, not me 😅

So I suggest to close this PR (#822) and the corresponding issue (#821).
Is that okay or would you like to implement the extension-property way?

PR #827, opened 2025-06-26
Use UTF-8 Encoding Without BOM by Default in All Serializers

### LINKED ISSUE #823: iCalendar UTF8 Encoding with/without BOM
RFC5545 sect. 3.4.1 says that iCal default charset is UTF8.
Additionally, applications must generate iCalendar streams in the UTF-8 charset.

Regarding the use of a Byte Order Mark (BOM) with UTF-8 there is no mention.
However, from my experience Android, Google, Apple **fail for UTF8 with BOM**.

Question @ical-org/maintainers 
When we serialize to a stream, should we use **`UTF8Encoding(false)` by default** (encoding parameter is null), unless the user provides an explicit (other) `Encoding`?

axunonb:
Summary:
Changed the default encoding for all serializers to UTF-8 without BOM (`new UTF8Encoding(false)`). This ensures compatibility with iCalendar consumers that do not support BOM and provides consistent encoding across all serialization operations.

Affected serializers:
All serializers inheriting from `SerializerBase`, `ComponentSerializer`, `EncodableDataTypeSerializer`, or `DataTypeSerializer` now default to UTF-8 without BOM unless an explicit encoding is provided.

- **Directly from `SerializerBase`:**
  - `SimpleDeserializer`
  - `ComponentSerializer`
  - `DataMapSerializer`
  - `DataTypes.DataTypeSerializer`
  - `DataTypes.DateTimeSerializer`
  - `DataTypes.DurationSerializer`
  - `GenericListSerializer`
  - `ParameterSerializer`
  - `PropertySerializer`

- **From `ComponentSerializer`:**
  - `CalendarSerializer`
  - `EventSerializer`
  - `SerializerFactory`

- **From `EncodableDataTypeSerializer` or `DataTypeSerializer`:**
  - `DataTypes.AttachmentSerializer`
  - `DataTypes.EnumSerializer`
  - `DataTypes.GeographicLocationSerializer`
  - `DataTypes.IntegerSerializer`
  - `DataTypes.PeriodListSerializer`
  - `DataTypes.PeriodSerializer`
  - `DataTypes.RecurrencePatternSerializer`
  - `DataTypes.StringSerializer`
  - `DataTypes.UriSerializer`
  - `DataTypes.UtcOffsetSerializer`
  - `DataTypes.WeekDaySerializer`

**Rationale:**
Ensures all iCalendar output is encoded in UTF-8 without BOM by default, as required by many consumers.

Resolves #823

### DIFF
--- a/Ical.Net.Tests/SerializationTests.cs
+++ b/Ical.Net.Tests/SerializationTests.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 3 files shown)


minichma:
Not sure, I understand this correctly. As the method is only invoked from test cases, this only makes a difference when invoked from externally if omitting the encoding, right? Doesn't this also affect all the places where serialization is triggered internally? Were searching for `UTF8` throughout the code base and found 17 matches. Do they require individual attention?

axunonb:
> this only makes a difference when invoked from externally

Yes. Without an encoding parameter we use UTF8 without BOM for serialization (writing).

> Doesn't this also affect all the places where serialization is triggered internally

Yes, as far as writing is concerned. For reading, BOM-handling is transparent by .Net.

> UTF8 throughout the code base and found 17 matches

All where writing is concerned should be covered.

minichma (approved):

Issue #239, opened 2017-02-20
Serialized EXDATE sometimes breaks spec


ChlorideCull:
In addition, changing the source code to this, to produce an all-day event:

```csharp
var now = new DateTime(2017, 1, 1);

//Repeat daily for 10 days
var rrule = new RecurrencePattern(FrequencyType.Daily, 1) { Count = 10 };

var e = new Event
{
    DtStart = new CalDateTime(now),
    RecurrenceRules = new List<IRecurrencePattern> { rrule },
};

var exceptionDateList = new PeriodList();
exceptionDateList.Add(new Period(new CalDateTime(now.AddDays(1))));
e.ExceptionDates.Add(exceptionDateList);

var calendar = new Calendar();
calendar.Events.Add(e);

var serializer = new CalendarSerializer(new SerializationContext());
var serializedCalendar = serializer.SerializeToString(calendar);
```

Produces this output in `serializedCalendar`:

```
BEGIN:VCALENDAR
PRODID:-//github.com/rianjs/ical.net//NONSGML ical.net 2.2//EN
VERSION:2.0
BEGIN:VEVENT
DTSTAMP:20170220T144719Z
DTSTART;VALUE=DATE:20170101
EXDATE:20170102/P1D
RRULE:FREQ=DAILY;COUNT=10
SEQUENCE:0
UID:08670eeb-76ad-46d1-bbfa-2e88f1597d58
END:VEVENT
END:VCALENDAR
```

In this case, EXDATE breaks the spec completely, as it's a date, but VALUE=DATE hasn't been set, and it has a weird duration suffixed to it.

----

In the long term, it's reasonable to consider changing `ExceptionDates` to not be a list of `IPeriodList`s. EXDATE can only be DATE-TIME or DATE according to the standard.

In the short term, suppressing duration output when serializing to ICS, and emitting the TZID parameter, should be enough.

rianjs:
We had to work around this recently, because I didn't have time to fix it in the library itself... You can specify TzId via Parameters. IIRC:

`exDate.Parameters.Add("TZID", "America/New_York");`

You may want to check if the parameter is already present before doing a blind add, as there is no duplicate checking.

But yes, it's a bug, and should be fixed.

rianjs:
> In this case, EXDATE breaks the spec completely, as it's a date, but VALUE=DATE hasn't been set, and it has a weird duration suffixed to it.

That "weird duration" is part of the spec:
[...]

PR #830, opened 2025-07-02
fix: Evaluation of `EXDATE` when date-only while `DTSTART` is date-time

### LINKED ISSUE #829: `GetOccurrences()` ignores `ExceptionDates` for today.
**Describe the bug**
If getting occurrences  `CaldendarEvent`, the `ExceptionDates` on the same day as `startTime` of the `GetOccurrences()`, are ignored.

**To Reproduce**
Simple .NET9 console app:

```cs
using Ical.Net.DataTypes;

void ShowOccurrences(CalendarEvent calendarEvent, string dateTime)
{
    Console.WriteLine($"Showing 2 occurrences starting from {dateTime}:");
    var occurrences = calendarEvent.GetOccurrences(new CalDateTime(DateTime.Parse(dateTime))).Take(2);

    foreach (var occurrencePeriod in occurrences.Select(o => o.Period))
    {
        Console.WriteLine($"Occurrence StartTime: {occurrencePeriod.StartTime}");
        Console.WriteLine($"Occurrence EffectiveEndTime: {occurrencePeriod.EffectiveEndTime}");
    }

    Console.WriteLine();
}

var cal = new CalendarEvent
          {
              Start = new CalDateTime(DateTime.Parse("2025-01-01T10:00:00")),
              End = new CalDateTime(DateTime.Parse("2025-01-01T20:00:00")),
              RecurrenceRules = [new RecurrencePattern("FREQ=DAILY")],
          };

cal.ExceptionDates.Add(new CalDateTime(DateOnly.Parse("2025-01-02")));

ShowOccurrences(cal, "2025-01-01T12:00:00");

ShowOccurrences(cal, "2025-01-02T12:00:00");
```

Output:
```
Showing 2 occurrences starting from 2025-01-01T12:00:00:
Occurrence StartTime: 01/01/2025 10:00:00 +01:00
Occurrence EffectiveEndTime: 01/01/2025 20:00:00 +01:00
Occurrence StartTime: 01/03/2025 10:00:00 +01:00
Occurrence EffectiveEndTime: 01/03/2025 20:00:00 +01:00

Showing 2 occurrences starting from 2025-01-02T12:00:00:
Occurrence StartTime: 01/02/2025 10:00:00 +01:00
Occurrence EffectiveEndTime: 01/02/2025 20:00:00 +01:00
Occurrence StartTime: 01/03/2025 10:00:00 +01:00
Occurrence EffectiveEndTime: 01/03/2025 20:00:00 +01:00
```

Notes:
 * The event is daily from 10:00 - 20:00
 * The time of `GetOccurrences()` is 12:00, i.e. after the start time, but before the end time of the event.
 * 2025-01-02 is excluded:
    * The first iteration is as expected: 2025-01-02 is not part of the occurrences.
    * The second iteration unexpectedly includes the occurrence `StartTime: 01/02/2025 10:00:00 +01:00`.

**Expected behavior**

Occurrences starting on an `ExceptionDates` should be ignored.

- OS: Windows 11
- .NET version: .NET 9
- ical.net version: 5.0.0 GA

**Additional context**

Seems somewhat related to #825, but not that I did not specify a duration. However, as mentioned in #828 occurrence periods (unexpectedly) do get a duration.

### LINKED ISSUE #832: GetOccurrences() ignores ExceptionDates
This is a follow-up of #829 and #830.

Mixing floating time and non-floating doesn't work well, because evaluation heavily relies on proper ordering of sequences, which isn't clear in such mixed cases. Usually this is forbidden, but we seem to support all-day EXDATEs (which per definition are floating) with non-floating date-times.

The following test case therefore fails:

```cs
    [Test]
    public void GetOccurrences_WithMixedKindExDatesAndTz_ShouldProperlyConsiderAll()
    {
        var cal = Calendar.Load("""
            BEGIN:VCALENDAR
            BEGIN:VEVENT
            DTSTART:20250701T000000Z
            DURATION:PT1H
            RDATE;TZID=Etc/GMT+12:20250702T234500
            RDATE;TZID=Etc/GMT-13:20250704T001500
            RDATE:20250703T113000Z
            EXDATE;VALUE=DATE:20250703
            END:VEVENT
            END:VCALENDAR
            """)!;

        var evt = cal.Events.Single();

        var occurrences = cal.GetOccurrences()
            .Select(o => o.Period.StartTime)
            .ToList();

        var expectedDates = new[]
        {
            new CalDateTime("20250701T000000", "UTC"),
            // unclear: in UTC (which also is the TZ of DTSTART) this is
            // on 20250703, so should it be excluded? Probably not.
            new CalDateTime("20250704T001500", "Etc/GMT-13"),

            // 20250703T113000 is on 20250703, so it should be removed according
            // to our implementation that considers DATE-only EXDATEs as all-day.
            // DISCUSS: Is this really, what the RFC intends? Or should mixing of
            // all-day and date-time be forbidden?
            // new CalDateTime("20250703T113000", "UTC"),

            // same as above: in UTC (which also is the TZ of DTSTART) this is
            // on 20250703, so should it be excluded? Probably not.
            new CalDateTime("20250702T234500", "Etc/GMT+12"),
        };

        Assert.That(occurrences, Is.EqualTo(expectedDates));
    }
```

Version:

Ical.Net 5.0.0 
Ical.Net 5.0.0 + PR #830

minichma:
Fix multiple issues related to the evaluation of `EXDATE`s:

* If `EXDATE`s of type `DATE`-only were used together with `DTSTART` of type `DATE-TIME`, and `GetOccurrences()` was used with a `periodStart` on the day of an `EXDATE`, then this EXDATE could be ignored.
* If `EXDATE`s of type `DATE` were mixed with those of type `DATE-TIME`, then the order could be confused, which could cause individual `EXDATE`s to be ignored

Its not clear from RFC 5545, whether EXDATEs and DTSTARTs of mixed, mismatching value types should be supported. Probably it shouldn't, because many things become unclear in such cases, especially when DTSTART has a time zone. Anyhow, because it seems to be used in the wild and has been supported in v4, we should probably continue to support it.

fixes #829, #832

### DIFF
--- a/Ical.Net/Evaluation/RecurringEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurringEvaluator.cs
@@ -117,9 +117,9 @@ public class RecurringEvaluator : Evaluator
         // by the RFC, but it seems to be used in the wild (see https://github.com/ical-org/ical.net/issues/829).
         // So we must make sure to return all-day EXDATEs that could overlap with recurrences, even if the day starts
         // before `periodStart`. We therefore start 2 days earlier (2 for safety regarding the TZ).
-        // DISCUSS: Should we support date-only EXDATEs being mixed with date-time DTSTARTs at all? What
-        // if one has a time zone and the other doesn't? Many details are unclear.
-        var exDateExclusionsDateOnly = EvaluateExDate(periodStart?.AddDays(-2), PeriodKind.DateOnly);
+        var exDateExclusionsDateOnly = new HashSet<DateOnly>(EvaluateExDate(periodStart?.AddDays(-2), PeriodKind.DateOnly)
+            .Select(x => x.StartTime.Date));
+
         var exDateExclusionsDateTime = EvaluateExDate(periodStart, PeriodKind.DateTime);
 
         var periods =
@@ -128,19 +128,19 @@ public class RecurringEvaluator : Evaluator
             .OrderedDistinct()
             .OrderedExclude(exRuleExclusions)
             .OrderedExclude(exDateExclusionsDateTime)
-            .OrderedExclude(exDateExclusionsDateOnly, DateOnlyExDateComparer)
+
+            // We accept date-only EXDATEs to be used with date-time DTSTARTs. In such cases we exclude those occurrences
+            // that, in their respective time zone, have a date component that matches an EXDATE.
+            // See https://github.com/ical-org/ical.net/pull/830 for more information.
+            //
+            // The order of dates in the EXDATEs doesn't necessarily match the order of dates returned by RDATEs
+            // due to RDATEs could have different time zones. We therefore use a regular `.Where()` to look up
+            // the EXDATEs in the HashSet rather than using `.OrderedExclude()`, which would require correct ordering.
+            .Where(dt => !exDateExclusionsDateOnly.Contains(dt.StartTime.Date))
 
             // Convert overflow exceptions to expected ones.
             .HandleEvaluationExceptions();
 
         return periods;
     }
-
-    /// <summary>
-    /// Compares whether the given period's date or date-time overlaps with the given date-only EXDATE.
-    /// The dates are considered to overlap if the date component equals. The time zone and time component
-    /// are ignored.
-    /// </summary>
-    private static readonly Comparer<Period> DateOnlyExDateComparer = Comparer<Period>.Create(
-        (p1, p2) => p1.StartTime.Date.CompareTo(p2.StartTime.Date));
 }
--- a/Ical.Net/Evaluation/RecurringEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurringEvaluator.cs
@@ -87,9 +87,9 @@ public class RecurringEvaluator : Evaluator
     /// </summary>
     /// <param name="referenceDate"></param>
     /// <param name="periodStart">The beginning date of the range to evaluate.</param>
-    protected IEnumerable<Period> EvaluateExDate(CalDateTime referenceDate, CalDateTime? periodStart)
+    private IEnumerable<Period> EvaluateExDate(CalDateTime referenceDate, CalDateTime? periodStart, params PeriodKind[] periodKinds)
     {
-        var exDates = Recurrable.ExceptionDates.GetAllPeriodsByKind(PeriodKind.DateOnly, PeriodKind.DateTime)
+        var exDates = Recurrable.ExceptionDates.GetAllPeriodsByKind(periodKinds)
             .AsEnumerable();
 
         if (periodStart != null)
@@ -119,14 +119,16 @@ public class RecurringEvaluator : Evaluator
         // before `periodStart`. We therefore start 2 days earlier (2 for safety regarding the TZ).
         // DISCUSS: Should we support date-only EXDATEs being mixed with date-time DTSTARTs at all? What
         // if one has a time zone and the other doesn't? Many details are unclear.
-        var exDateExclusions = EvaluateExDate(referenceDate, periodStart?.AddDays(-2));
+        var exDateExclusionsDateOnly = EvaluateExDate(referenceDate, periodStart?.AddDays(-2), PeriodKind.DateOnly);
+        var exDateExclusionsDateTime = EvaluateExDate(referenceDate, periodStart, PeriodKind.DateTime);
 
         var periods =
             rruleOccurrences
             .OrderedMerge(rdateOccurrences)
             .OrderedDistinct()
             .OrderedExclude(exRuleExclusions)
-            .OrderedExclude(exDateExclusions, Comparer<Period>.Create(CompareExDateOverlap))
+            .OrderedExclude(exDateExclusionsDateTime)
+            .OrderedExclude(exDateExclusionsDateOnly, DateOnlyExDateComparer)
 
             // Convert overflow exceptions to expected ones.
             .HandleEvaluationExceptions();
@@ -135,18 +137,10 @@ public class RecurringEvaluator : Evaluator
     }
 
     /// <summary>
-    /// Compares whether the given period's date overlaps with the given EXDATE. The dates are
-    /// considered to overlap if they start at the same time, or the EXDATE is an all-day date
-    /// and the period's start date is the same as the EXDATE's date.
-    /// <para/>
-    /// Note: <see cref="Period.EffectiveDuration"/> for <paramref name="exDate"/> is always <see langword="null"/>.
+    /// Compares whether the given period's date or date-time overlaps with the given date-only EXDATE.
+    /// The dates are considered to overlap if the date component equals. The time zone and time component
+    /// are ignored.
     /// </summary>
-    private static int CompareExDateOverlap(Period period, Period exDate)
-    {
-        var cmp = period.CompareTo(exDate);
-        if ((cmp != 0) && !exDate.StartTime.HasTime && (period.StartTime.Value.Date == exDate.StartTime.Value))
-            cmp = 0;
-
-        return cmp;
-    }
+    private static readonly Comparer<Period> DateOnlyExDateComparer = Comparer<Period>.Create(
+        (p1, p2) => p1.StartTime.Date.CompareTo(p2.StartTime.Date));
 }
--- a/Ical.Net/Evaluation/RecurringEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurringEvaluator.cs
@@ -46,7 +46,7 @@ public class RecurringEvaluator : Evaluator
     }
 
     /// <summary> Evaluates the RDate component. </summary>
-    protected IEnumerable<Period> EvaluateRDate(CalDateTime referenceDate, CalDateTime? periodStart)
+    protected IEnumerable<Period> EvaluateRDate(CalDateTime? periodStart)
     {
         var recurrences = Recurrable.RecurrenceDates
                 .GetAllPeriodsByKind(PeriodKind.Period, PeriodKind.DateOnly, PeriodKind.DateTime)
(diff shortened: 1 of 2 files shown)


minichma:
@ical-org/maintainers It might be good to fix this issue, but in general, mixing DATE-only, DATE-TIME, floating and non-floating time, doesn't seem to be a good idea to me. Would be good to know, how much this is actually used in the wild.

E.g. see this test case:

https://github.com/ical-org/ical.net/pull/830/files#diff-e28a8bc50c87925affccd06958f171b527c5abe69dcb4b7b2cd90705e1ddf07fR4108-R4115

The problem arises from the fact, that the order relation is unclear in such cases. The problem gets worse, if timezones would be involved.

What do you think?

NRG-Drink:
I read the RFC definition of `EXDATE` and stumbled across this:
> The "EXDATE" property can be used to exclude the value specified in "DTSTART".

I interpret this as: The `EXDATE` must match exactly with the `DTSTART` of the occurrence to exclude.
I don't agree to exclude all events on the given `EXDATE` date-value, because there can be multiple occurrences on the same day and than all would be ignored even if I just want to ignore one of them. This is the case for recurrences smaller than daily or on any pattern when we move an event.
An edge-case would be, when I move an event's start-date to another event's start-date. In this case I would have to exclude both or non. But I think this is intended and is not part of our scope.

If we match `DTSTART` exactly with `EXDATE` we can allow date-only and timezones without much trouble (at least in my head for the moment of writing 😅)

On the other hand, I can see the case to exclude all events by only providing the date since most calendars have daily as the smalles recurrence pattern. But even there we have to think about the moved events.

axunonb:
Very good discussion!

Having another look at the iCalendar specification section 3.8.5.1 we can see
```
exdate = "EXDATE" exdtparam ":" exdtval *("," exdtval) CRLF
exdtparam = *(
;
; The following are OPTIONAL,
; but MUST NOT occur more than once.
;
(";" "VALUE" "=" ("DATE-TIME" / "DATE")) /
;
(";" tzidparam) /
;
; The following is OPTIONAL,
; and MAY occur more than once.
;
(";" other-param)
;
)
exdtval = date-time / date
;Value MUST match value type
```

**`Value MUST match value type`:**
`EXDATE` property interacts with `DTSTART`, including the requirement that their value types must match (i.e., both must be either `DATE` or `DATE-TIME`). So this supports @NRG-Drink 's arguments.

I also agree with @minichma "but in general, mixing DATE-only, DATE-TIME, floating and non-floating time, doesn't seem to be a good idea".

There are some supporting references from "the wild":

1. In the past we already had a related discussion in #239.

2. A [Stack Overflow discussion](https://stackoverflow.com/questions/48481980/exdate-recurrence-setting-being-ignored) with the very same topic.
It perfectly demonstrates Google Calendar's preference for matching the DTSTART type for EXDATE, even if VALUE=DATE is explicitly used.

3. There is a [Google Calendar Community Thread](https://support.google.com/calendar/thread/286605127/exdate-doesn-t-work-as-expected-when-ical-file-is-imported?hl=en). Multiple users report that when importing .ics files into Google Calendar, EXDATE entries (especially those intended to exclude specific dates for recurring events that have a time component) are not being correctly honored. The excluded events still appear in their calendar. The common suggested solution from contributors is to ensure that the `EXDATE` entry includes the time and timezone, matching the DTSTART of the recurring event.

4. Outlook 365: Importing an `.ics` file with `EXDATE` with type `DATE-TIME` or `DATE` while `DTSTART` is `DATE-TIME`: The date is excluded in both cases.
This behavior is specified [MS-STANOICAL Section 2.2.85](https://learn.microsoft.com/en-us/openspecs/exchange_standards/ms-stanoical/dcd6ef46-96c6-4d8a-88a9-3dfcc8b2f874): "Outlook can import EXDATE as either a DATE-TIME or DATE, provided the value falls on the same day as the original start date of an instance of the recurrence, in the recurrence's time zone."
Unfortunately it's not documented how MS implements this behavior in code.

What **could** be our strategy - to be discussed:

#### 1. **Strict RFC Compliance by Default**
- Ensure that serialization and validation logic **enforces matching value types** between `DTSTART` and `EXDATE`, per RFC 5545.
- This keeps iCal.NET predictable and standards-compliant for systems like Google Calendar.

#### 2. **Optional Leniency Mode**
- Introduce a configurable flag to `EvaluationOptions` (e.g., `AllowLooseExdateMatching`) that enables Outlook-style behavior.
- When enabled, allow `EXDATE` values of type `DATE` to match `DATE-TIME` instances by comparing only the date portion.

#### 3. **Parsing Logic**
- If we had a `Logger` we could detect mismatched types and log a warning. Unfortunately we don't.
- Normalize `EXDATE` values to `DATE-TIME` if `DTSTART` is `DATE-TIME`, using midnight (`T000000`) or the original `DTSTART` time. This might be better than to ignore `EXDATE` because of a type mismatch.

#### 4. **Developer Guidance**
- Document the behaviors above in the wiki and XML comments.
- Offer examples showing both strict and lenient modes.

BTW: #829 shouldn't be considered a bug, should it?

minichma:
> exdtval = date-time / date
  ;Value MUST match value type

> Value MUST match value type:
EXDATE property interacts with DTSTART, including the requirement that their value types must match (i.e., both must be either DATE or DATE-TIME).

I think that the spec only says that the value type specified *with the property*  must match. It doesn't seem to refer to DTSTART.

axunonb:
> spec only says that the value type specified with the property must match

Didn't interpret this way, but yes: good point. But still: Microsoft acknowledges deviations from RFC 5545 where Outlook accepts `EXDATE` values of type `DATE` even when `DTSTART` is `DATE-TIME`, as long as the date portion matches a recurrence instance.

minichma:
> BTW: https://github.com/ical-org/ical.net/issues/829 shouldn't be considered a bug, should it?

I think it should. The core of the problem there is that `periodStart` is not handled properly. Even if we'd specify that mixing of DATE-only and DATE-TIME wouldn't be allowed, there would still be a inconsistency, because without `periodStart` the EXDATE would be considered.

Fuzzillogic:
FWIW, as mentioned in https://github.com/ical-org/ical.net/issues/829#issuecomment-3023957975, our use case is that some recurring events (with type DATE-TIME as DTSTART / DTEND) should not occur on holidays (i.e. EXDATE of type DATE). So, we feed the relevant list of holidays as EXDATES for those recurring events. To determine whether a specific occurrence is in such EXDATE date, I'd expect it to simply use the DATE part of the calculated start DATE-TIME of the occurrence. That might handle the time zone question.

Thus, being able to use EXDATE of type DATE while the recurring event is DATE-TIME certainly simplifies things for us.

minichma:
Sorry, I commented only on parts of your comment so far. Thanks for the list of how 'the big ones' deal with this case. So I guess its clear then, that we'll have to support the case too.

> Optional Leniency Mode

I think we should implement one way of how we do it. If we add more modes, testing will become a nightmare. I also don't see a major benefit in having different modes here. Moreover, the RFC is not really clear in that respect.

Regarding how Outlook deals with the case:

> Outlook can import EXDATE as either a DATE-TIME or DATE, provided the value falls on the same day as the original start date of an instance of the recurrence, in the recurrence's time zone.

That the date is compared 'in the recurrence's time zone' prevents us from using `OrderedExclude`, because the order of dates _in the individual recurrence's time zone_ doesn't necessarily have to be the same as in UTC (compare #832). Anyhow, that shouldn't be a problem. We can just use Linq's regular `.Except()`, because the list of dates will usually be short.

minichma:
Implemented it in a way now that each recurrence's date value in the recurrence's individual time zone is matched against the list of date-only EXDATEs. If the date portion matches, the recurrence is excluded.

Tried to find out what outlook.com and Google Calendar are doing. This is what I found:

* outlook.com: Doesn't seem to support importing RDATEs altogether. So the question of how to deal with different TZ within an event doesn't really apply. The date-only EXDATE is considered though and obviously matched against the recurrence's date part (in the TZ of DTSTART).
* Google Calendar: Supports importing RDATEs. If an RDATE or EXDATE are DATE-only, while DTSTART is DATE-TIME, then it seems to assume the DATE-only values are at 0:00 in the TZ of DTSTART. 

So those implementations seem to be conflicting in this respect and the now proposed is yet different.

@ical-org/maintainers What do you think?

Ical.Net/Evaluation/RecurringEvaluator.cs
```diff
@@ -85,11 +85,11 @@ protected IEnumerable<Period> EvaluateExRule(CalDateTime referenceDate, CalDateT
     /// <summary>
     /// Evaluates the ExDate component.
     /// </summary>
-    /// <param name="referenceDate"></param>
     /// <param name="periodStart">The beginning date of the range to evaluate.</param>
-    protected IEnumerable<Period> EvaluateExDate(CalDateTime referenceDate, CalDateTime? periodStart)
+    /// <param name="periodKinds">The period kinds to be returned. Used as a filter.</param>
+    private IEnumerable<Period> EvaluateExDate(CalDateTime? periodStart, params PeriodKind[] periodKinds)
```
axunonb, line 90: Changing the signature and access modifier would mean a breaking change?

axunonb:
> What do you think?

* No problem with Outlook
* Yes, that's what Google Calendar has implemented. Normalizing to midnight is (IMHO a worse but more simple) alternative to using `DTSTART` time part. We can't cover all.

I'd say the updated implementation in this PR is fine and will get another entry to FAQ. 
@NRG-Drink Your opinion?

Ical.Net/Evaluation/RecurringEvaluator.cs
minichma: Puh, right, theoretically yes. But for this to be an issue, somebody would have to inherit Recurring Evaluator and then call this method. I really wouldn't expect this to be the case. Would wait for potential complaints.

axunonb (approved):

PR #833, opened 2025-07-04
fix: `CalDateTime` CTOR using ISO 8601 UTC string resolves to UTC

### LINKED ISSUE #831: CalDateTime constructor doesn't properly parse `Z` time
**Describe the bug**

Calling the `CalDateTime` constructor with a `Z` date-time like `new CalDateTime("20250703T060000Z")` should result in an instance having `IsUtc` set but it doesn't.

Consider the following test case, which fails:

```cs
    [Test]
    public void TestParseTimeWithZ()
    {
        var dt = new CalDateTime("20250703T060000Z");
        Assert.Multiple(() =>
        {
            Assert.That(dt.Value, Is.EqualTo(new DateTime(2025, 7, 3, 6, 0, 0, DateTimeKind.Utc)));
            Assert.That(dt.IsUtc, Is.True);
        });
    }
```

The fact that the trailing `Z` refers to UTC is lost during deserialization.

Version:

Ical.Net 5.0.0 initial release

axunonb:
Throws for controversion of ISO 8601 UTC string, while timezone ID is not UTC

chore: `ExcludeFromCodeCoverage` for private constructor
chore: Remove unused method `TruncateTimeToSeconds`

Resolves #831

### DIFF
--- a/Ical.Net/DataTypes/CalDateTime.cs
+++ b/Ical.Net/DataTypes/CalDateTime.cs
@@ -7,6 +7,7 @@ using Ical.Net.Serialization.DataTypes;
 using Ical.Net.Utility;
 using NodaTime;
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 
@@ -55,6 +56,7 @@ public sealed class CalDateTime : IComparable<CalDateTime>, IFormattable
     /// <summary>
     /// This constructor is required for the SerializerFactory to work.
     /// </summary>
+    [ExcludeFromCodeCoverage]
     private CalDateTime()
     {
         // required for the SerializerFactory to work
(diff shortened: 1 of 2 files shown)


Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -201,9 +203,15 @@ public CalDateTime(string value, string? tzId = null)
     {
         var serializer = new DateTimeSerializer();
         CopyFrom(serializer.Deserialize(new StringReader(value)) as CalDateTime
-                 ?? throw new InvalidOperationException($"$Failure for deserializing value '{value}'"));
+                 ?? throw new InvalidOperationException($"$Failure when deserializing value '{value}'"));
+
         // The string may contain a date only, meaning that the tzId should be ignored.
-        _tzId = HasTime ? tzId : null;
+        _tzId ??= HasTime ? tzId : null;
```
NRG-Drink: `CopyFrom` is used only once and it will set `_tzId` regardless of the time.
I suggest to delete `CopyFrom` and use `Initialize`.


```suggestion
        var dt = serializer.Deserialize(new StringReader(value)) as CalDateTime
                 ?? throw new InvalidOperationException($"$Failure when deserializing value '{value}'");

        Initialize(dt._dateOnly, dt._timeOnly, dt._tzId);
```

axunonb: Good catch, `CopyFrom`  is redundant.
Initializing should read
```cs
Initialize(dt._dateOnly, dt._timeOnly, dt.IsUtc ? UtcTzId : tzId);
```
to reflect that `UTC` may be set from the deserialized string (appended 'Z') *_or_* from the `tzId` argument. This correction is actually the aim of the PR.

axunonb:
@NRG-Drink Thanks for the helpful review

PR #835, opened 2025-07-05
Nrg/feat/caldatetime value kind

### LINKED ISSUE #834: CalDateTime.Value should return proper DateTimeKind on UTC time zone
`CalDateTime.Value` will always return the `DateTimeKind.Unspecified` even if the time zone is UTC.
I suggest that `CalDateTime.Value` should return `DateTimeKind.Utc` when the time zone is UTC.

This is also the case in `CalDateTime.AsUtc` but not tested.

NRG-Drink:
closes #834


axunonb:
See my comment to #834. Please let's usually create an issue, discuss it and create a PR after that, in order make the best out of the time we invest into the project.

PR #836, opened 2025-07-09
Fix `GetOccurrences(periodStart)` to return still active occurrences

### LINKED ISSUE #825: GetOccurrences() with periodStart skips first occurrences if VEVENT has duration
**Describe the bug**

The following test case fails but shouldn't.

```cs
    [Test]
    public void GetOccurrences_WithPeriodStart_ShouldIncludeEventsEndingAfterPeriodStart()
    {
        var ical = """
                   BEGIN:VCALENDAR
                   BEGIN:VEVENT
                   DTSTART:20250626T000000
                   DURATION:P2D
                   RRULE:FREQ=DAILY
                   END:VEVENT
                   END:VCALENDAR
                   """;
        var cal = Calendar.Load(ical)!;
        var actual = cal.GetOccurrences(new CalDateTime("20250630T235959")).First();

        // Expect the first occurence returned to be 20250629T000000, because it ends at
        // 20250701T000000, and therefore is the first to end after periodStart.
        Assert.That(actual.Period.StartTime, Is.EqualTo(new CalDateTime("20250629T000000")));
    }
```

fails with

```
    Assert.That(actual.Period.StartTime, Is.EqualTo(new CalDateTime("20250629T000000")))
    Expected: <06/29/2025 00:00:00 +02:00>
    But was:  <06/30/2025 00:00:00 +02:00>
```

Stack Trace: 

According to this code

https://github.com/ical-org/ical.net/blob/7556e7b6b77dcfd5ab6b5c5ba795b3f72ac2fa64/Ical.Net/Evaluation/RecurrenceUtil.cs#L42

an occurrence is returned not only if it *starts* after `periodStart`, but also if it *ends* after it. However, as `RecurrencePatternEvaluator` already does some pre-filtering for `periodStart` but doesn't consider the event's duration, the first entries are missing.

[Edit] Fixed duration in test case from 1D to 2D

- ical.net version: 5.0.0

**Additional context**
Has been identified when working on #824

**Discuss**
By `GetOccurrences` filtering not only based on an occurrence's start but also its end increases complexity as can be seen by this issue. Maybe reconsider this behavior in the future (v6).

minichma:
See #825: According to existing code, `GetOccurrences(periodStart)` seems to be intended to return not only those recurrences that start at or after `periodStart`, but also those still active at that date-time. There's has been some discussion in #825 about whether to follow the original intention or rather just return those that _start_ at or after `periodStart`.

This PR introduces a new evaluation option that allows to control the intended behavior. By default only those are returned that start at or after `periodStart`. By setting the new option `EvaluationOptions.IncludeStillActiveOccurrences`, also those are returned that started in the past but are still active at the given point in time.

Fixes #825

### DIFF
--- a/Ical.Net/CalendarComponents/RecurringComponent.cs
+++ b/Ical.Net/CalendarComponents/RecurringComponent.cs
@@ -22,7 +22,7 @@ namespace Ical.Net.CalendarComponents;
 /// RRULEs, RDATE, EXRULEs, and EXDATEs, as well as the DTSTART
 /// for the recurring item (all recurring items must have a DTSTART).
 /// </remarks>
-public class RecurringComponent : UniqueComponent, IRecurringComponent
+public abstract class RecurringComponent : UniqueComponent, IRecurringComponent
 {
     public static IEnumerable<IRecurringComponent> SortByDate(IEnumerable<IRecurringComponent> list) => SortByDate<IRecurringComponent>(list);
 
@@ -151,17 +151,15 @@ public class RecurringComponent : UniqueComponent, IRecurringComponent
     /// </summary>
     public virtual ICalendarObjectList<Alarm> Alarms => new CalendarObjectListProxy<Alarm>(Children);
 
-    private RecurringEvaluator? _evaluator;
+    public abstract IEvaluator? Evaluator { get; }
 
-    public virtual IEvaluator? Evaluator => _evaluator;
-
-    public RecurringComponent()
+    protected RecurringComponent()
     {
         Initialize();
         EnsureProperties();
     }
 
-    public RecurringComponent(string name) : base(name)
+    protected RecurringComponent(string name) : base(name)
     {
         Initialize();
         EnsureProperties();
@@ -169,7 +167,6 @@ public class RecurringComponent : UniqueComponent, IRecurringComponent
 
     private void Initialize()
     {
-        _evaluator = new RecurringEvaluator(this);
         ExceptionDates = new ExceptionDates(ExceptionDatesPeriodLists);
         RecurrenceDates = new RecurrenceDates(RecurrenceDatesPeriodLists);
     }
--- a/Ical.Net/Utility/DateUtil.cs
+++ b/Ical.Net/Utility/DateUtil.cs
@@ -132,4 +132,12 @@ internal static class DateUtil
     /// </remarks>
     internal static DataTypes.Duration ToDuration(this TimeSpan timeSpan)
         => DataTypes.Duration.FromTimeSpan(timeSpan);
+
+    internal static CalDateTime AddLeniently(this CalDateTime dt, DataTypes.Duration d)
+    {
+        if (d.HasTime && !dt.HasTime)
+            dt = new CalDateTime(dt.Date, TimeOnly.MinValue);
+
+        return dt.Add(d);
+    }
 }
--- a/Ical.Net.Tests/RecurrenceTests.cs
+++ b/Ical.Net.Tests/RecurrenceTests.cs
@@ -4170,4 +4170,42 @@ END:VCALENDAR";
 
         Assert.That(occurrences, Is.EqualTo(expectedDates));
     }
+
+    [Test]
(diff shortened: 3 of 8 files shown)


minichma:
@ical-org/maintainers Not sure it is the best idea to make the behavior configurable, as it introduces additional complexities, requires additional testing, etc. Please let me know what you think.

If someone would find the time, the PR would also deserve additional tests related to occurrences that should be returned and cross a DST change _before_ `periodStart`. Not sure this case is handled correctly, probably not (anyhow, its better than today).

minichma:
@ical-org/maintainers formally the changes are breaking, at least due to classes becoming abstract. However, effectively those are not intended for being constructed/inherited by user code, so I'd consider this acceptable.

axunonb:
Thanks! Few first comments:

1. Looks like a good solution.
2. Should `EvaluationOptions.IncludeStillActiveOccurrences == true` be the default behavior (also mentioned in #825)? Then we wouldn't even need the new option, because the caller could easily identify ongoing occurrences.
3. Naming: `EvaluationOptions.IncludeOngoingOccurrences` instead of `EvaluationOptions.IncludeStillActiveOccurrences`?
4. XmlDoc for `IGetOccurrences.GetOccurrences` should reflect the behavior regarding ongoing occurrences.
5. Not part of the PR: Update the [Wiki](https://github.com/ical-org/ical.net/wiki/Working-with-recurring-elements)

[Edit]
> not intended for being constructed/inherited by user code,

Agree, should we then consequently disallow construction inheritance by user code?

minichma:
> Then we wouldn't even need the new option, because the caller could easily identify ongoing occurrences.

Agree, so I removed the last commit. Introduced the configuration because I don't like the behavior of filtering by the end time too much because the other way is simpler and easier to understand. Anyhow, the configuration adds even more complexity.

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -22,7 +22,7 @@ namespace Ical.Net.CalendarComponents;
 /// RRULEs, RDATE, EXRULEs, and EXDATEs, as well as the DTSTART
 /// for the recurring item (all recurring items must have a DTSTART).
 /// </remarks>
-public class RecurringComponent : UniqueComponent, IRecurringComponent
+public abstract class RecurringComponent : UniqueComponent, IRecurringComponent
```
axunonb, line 25: Should the class rather become `internal`?

Ical.Net/Utility/DateUtil.cs
```diff
@@ -132,4 +132,12 @@ internal static DataTypes.Duration ToDurationExact(this TimeSpan timeSpan)
     /// </remarks>
     internal static DataTypes.Duration ToDuration(this TimeSpan timeSpan)
         => DataTypes.Duration.FromTimeSpan(timeSpan);
+
+    internal static CalDateTime AddLeniently(this CalDateTime dt, DataTypes.Duration d)
```
axunonb, line 136: Couldn't find a better name either ;)

axunonb (approved):
LGTM - thanks!

Ical.Net/Utility/DateUtil.cs
minichma: 😄

Ical.Net/CalendarComponents/RecurringComponent.cs
minichma: Would probably be good, but as it would be a breaking change and not strictly necessary, I left as is. Maybe in a separate PR?

axunonb: > Would probably be good, but as it would be a breaking change and not strictly necessary, I left as is. Maybe in a separate PR?

Actually making `RecurringComponent` `internal` is not that straightforward, so we should leave it as is:
```txt
Inconsistent accessibility: base class 'RecurringComponent' is less accessible than class 'CalendarEvent'
Inconsistent accessibility: base class 'RecurringComponent' is less accessible than class 'Journal'
Inconsistent accessibility: base class 'RecurringComponent' is less accessible than class 'Todo'
```

Issue #455, opened 2019-08-13
`RECURRENCE-ID;RANGE=THISANDFUTURE` implementation is missing


amstech:
there are two possible paramters that can be provided with Recurrence-Id: value and range.  Does anyone know how to provide Range to recurrence-id.  I am looking for ability to provide update to a range of dates within the recurrence.

I want the output to be like this: RECURRENCE-ID;RANGE=THISANDFUTURE:20190101T120000Z

https://tools.ietf.org/html/rfc5546#section-4.4.5

minichma:
@axunonb `RANGE=THISANDFUTURE` is still not supported. Should this be reopened?

axunonb:
@minichma Sub-issues are still in beta and not available in this repo. So let's proceed with recurrence for `RECURRENCE-ID;RANGE=THISANDFUTURE` here.

> [!NOTE]
#870 is merged and allows for serializing and deserializing `RECURRENCE-ID;RANGE=THISANDFUTURE` on events.
Implementation of the feature for recurrence is still ahead.

axunonb:
_**Copied from [PR 870](https://github.com/ical-org/ical.net/pull/870#issuecomment-3357879530)**_

@minichma:
Hm, my naive assumption was as follows, but seems to be wrong:
1. Generate recurrences of original and new VEVENT independently
2. If
    * the range is `THISANDFUTURE` do something like 'oldSequence.TakeWhile(item => item != recurrenceId).OrderedMerge(newSequence)`
    * else do something like 'oldSequence.Where(item => item != recurrenceId).OrderedMerge(newSequence)`

My assumption was that the new component may or may not have an RRULE or RDATE. If it hasn't, it produces just a single occurrence. But looking at [the RFC](https://www.rfc-editor.org/rfc/rfc5545.html#section-3.8.4.4)

> When the given recurrence instance is
      rescheduled, all subsequent instances are also rescheduled by the
      same time difference.  For instance, if the given recurrence
      instance is rescheduled to start 2 hours later, then all
      subsequent instances are also rescheduled 2 hours later.

this sounds more like the new VEVENT (the one with the RECURRENCE-ID) doesn't have its own RRULE/RDATE/EXDATE properties but inherits those of the original one, only the offset is considered, right? If so, several questions arise:

* What if the new RRULE *does* have an RRULE? Does it replace the original one?
* What if the new component's DTSTART doesn't match the original RRULE? Probably undefined.
[...]

PR #864, opened 2025-09-17
Fix: Event with `RECURRENCE-ID` overrides occurrence from orginal event

### LINKED ISSUE #863: Event with RECURRENCE-ID removes occurrence completely
Copied from @ramonsmits's closed PR #861

With v5.1.0 the console output is:
```
Events:
	00p0ja7 11/3/2025 12:00:00 AM 11/24/2025 12:00:00 AM QWERTY
	00p0ja7 11/3/2025 12:00:00 AM 11/24/2025 12:00:00 AM QWERTY
Occurrences:
	10/5/2026 12:00:00 AM
	10/5/2026 12:00:00 AM
	10/5/2026 12:00:00 AM
```

With v4 is does return an occurrence for 2025.

When I remove the `.OrderedDistinct()` call in `GetOccurrences<T>` I see too many results but I do see the occurrence there. 

Maybe a issue in the `EqualityComparer`?

```cs
[Test, Explicit]
public void MissingOccurrences()
{
    var cal = Calendar.Load("""
        BEGIN:VCALENDAR
        PRODID:-//Google Inc//Google Calendar 70.9054//EN
        VERSION:2.0
        CALSCALE:GREGORIAN
        METHOD:PUBLISH
        X-WR-CALNAME:Non-Critical Support Roster
        X-WR-TIMEZONE:UTC
        BEGIN:VEVENT
        DTSTART;VALUE=DATE:20251103
        DTEND;VALUE=DATE:20251124
        RRULE:FREQ=WEEKLY;WKST=MO;INTERVAL=48;BYDAY=MO
        DTSTAMP:20250912T100327Z
        UID:00p0ja7t2446ja22sl02ah2uun@google.com
        CREATED:20250214T230308Z
        LAST-MODIFIED:20250409T173619Z
        SEQUENCE:1
        STATUS:CONFIRMED
        SUMMARY:QWERTY
        TRANSP:TRANSPARENT
        END:VEVENT
        BEGIN:VEVENT
        DTSTART;VALUE=DATE:20251103
        DTEND;VALUE=DATE:20251124
        DTSTAMP:20250912T100327Z
        UID:00p0ja7t2446ja22sl02ah2uun@google.com
        RECURRENCE-ID;VALUE=DATE:20251103
        CREATED:20250214T230308Z
        LAST-MODIFIED:20250428T170953Z
        SEQUENCE:1
        STATUS:CONFIRMED
        SUMMARY:QWERTY
        TRANSP:TRANSPARENT
        END:VEVENT
        END:VCALENDAR
        """)!;

    Console.WriteLine("Events:");
    foreach (var e in cal.Events.OrderBy(x=>x.Start))
    {
        Console.WriteLine($"\t{e.Uid.Substring(0,7)} {e.Start.Value} {e.End.Value} {e.Summary}");
    }

    var dt = new CalDateTime(DateOnly.FromDateTime(new DateTime(2026,1,1)));
    var occurrences = cal
        .GetOccurrences<CalendarEvent>()
        .TakeWhile(p => p.Period.StartTime <= dt);

    Console.WriteLine("Occurrences:");
    foreach (var o in occurrences)
    {
        var e = (CalendarEvent)o.Source;
        Console.WriteLine($"\t{e.Uid.Substring(0,7)} {o.Period.StartTime.Value} {o.Period.EffectiveEndTime.Value} {e.Summary}");
    }

    Assert.That(occurrences, Is.Not.Empty); // fails
}
```

### LINKED ISSUE #865: Chained `RECURRENCE-ID` not working
If a calendar contains multiple recurrence objects having the same `UID` that replace each other via `RECURRENCE-ID`, then Ical.Net doesn't properly handle the case.

I.e. the following code will return any recurrence from an object having a `RECURRENCE-ID` even if it has been replaced by another one.

https://github.com/ical-org/ical.net/blob/bd4225eb4670d0d3f212daf57a0bc81db13d7279/Ical.Net/Calendar.cs#L217-L219

Ical.Net Version 5.1.0

axunonb:
Resolves #863
Resolves #865


ramonsmits:
@axunonb I applied this fix to my earlier branch and minimized the ics data but that still seems to not generate all occurrences:

https://github.com/ical-org/ical.net/compare/main...ramonsmits:ical.net:occurence-issue

I also pushed the same test but for v4 that does generate the expected occurrences:

https://github.com/ical-org/ical.net/compare/main...ramonsmits:occurrence-issue-v4

axunonb:
Yes, I didn't use the same DTSTART in Master and Override event. That seems to be the reason, and I got excited too soon ;-)
The Override event is not part of occurrences, while it is expected to be. That's why the Master gets filtered out and so the expected date is missing. We're getting closer though.

axunonb:
@ramonsmits Should work now also when the Override Event has the same DTSTART

Ical.Net/Calendar.cs
```diff
@@ -187,15 +187,22 @@ public virtual IEnumerable<Occurrence> GetOccurrences(CalDateTime? startTime = n
...
 
         var occurrences = RecurringItems
             .OfType<T>()
-            .Select(recurrable => recurrable.GetOccurrences(startTime, options))
+            .Select(recurrable => recurrable.GetOccurrences(startTime, options)
+                .Where(r =>
+                    (r.Source.RecurrenceId != null) ||
+                    r.Source is not IUniqueComponent uniqueComp ||
+                    !recurrenceIdsAndUids.ContainsKey((uniqueComp.Uid, r.Period.StartTime.Value)))
+            )
```
ramonsmits: Maybe I suggest you add a code comment here to explain the order/behavior?

Please fill the `.......`

```suggestion
            // Filter UID/RECURRENCE-ID combinations that replace other occurrences before using `.OrderedDistinct()` to ensure that ......
            .Where(r =>
                (r.Source.RecurrenceId != null) ||
                r.Source is not IUniqueComponent uniqueComp ||
                !recurrenceIdsAndUids.ContainsKey((uniqueComp.Uid, r.Period.StartTime.Value))
              )
            )
```

Ical.Net/Calendar.cs
```diff
@@ -210,13 +217,7 @@ public virtual IEnumerable<Occurrence> GetOccurrences<T>(CalDateTime? startTime
             .OrderedDistinct()
```
ramonsmits, line 211: What is the expected order? Looking at the implementation of `OrderedDistinct` it it keeps the order of the enumerator and the first of its "sequence" is kept. However, in the code above I do not see any sorting happening. Is there some order implied from the `RecurringItems` items?

ramonsmits:
btw, tested last against my own ical and it seems to now correctly generate the occurrences 💪

Ical.Net/Calendar.cs
axunonb: It ordered by `Period.StartTime`

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -4247,4 +4323,288 @@ public void GetOccurrences_WithPeriodStart_ShouldConsiderTzCorrectly(string dtSt
...
+            // endTime = 20161211T000000
+            Assert.That(occurrences.Select(x => x.Period.StartTime), Is.EqualTo(expectedStartDates));
+
+            // endTime = 20161128T002000
+            Assert.That(occurrences2.Select(x => x.Period.StartTime), Is.EqualTo(expectedStartDates.Take(4)));
+        });
+    }
+
+    [Test]
+    public void GetOccurrencesWithRecurrenceId_DateOnly_ShouldEnumerate()
```
minichma, line 4645: The new tests actually don't reproduce the original issue. Maybe add the test from the ticket?

https://github.com/ical-org/ical.net/issues/863#issue-3423167159

Ical.Net/Calendar.cs
```diff
@@ -187,15 +187,25 @@ public virtual IEnumerable<Occurrence> GetOccurrences(CalDateTime? startTime = n
...
     {
         // These are the UID/RECURRENCE-ID combinations that replace other occurrences.
-        var recurrenceIdsAndUids = this.Children.OfType<IRecurrable>()
+        var recurrenceIdsAndUids = Children.OfType<IRecurrable>()
             .Where(r => r.RecurrenceId != null)
-            .Select(r => new { (r as IUniqueComponent)?.Uid, Dt = r.RecurrenceId!.Value })
-            .Where(r => r.Uid != null)
+            // Create a value tuple instead of an anonymous type, because
+            // anonymous types don't work well as dictionary keys due to equality semantics.
+            .Select(r => ((r as IUniqueComponent)?.Uid, r.RecurrenceId!.Value))
```
minichma: I fully agree that it is better to use tuples here rather than anonymous types but I don't think this is the root cause of the problem (as suspected in the issue's comments).

Ical.Net/Calendar.cs
```diff
@@ -187,15 +187,25 @@ public virtual IEnumerable<Occurrence> GetOccurrences(CalDateTime? startTime = n
...
             .ToDictionary(x => x);
 
         var occurrences = RecurringItems
             .OfType<T>()
-            .Select(recurrable => recurrable.GetOccurrences(startTime, options))
+            .Select(recurrable => recurrable.GetOccurrences(startTime, options)
+                // Exclude occurrences that are overridden by other components with the same UID and RECURRENCE-ID.
+                // This must happen before .OrderedDistinct() because that method would remove duplicates
+                // based on the occurrence time, and we need to remove them based on UID + RECURRENCE-ID.
+                .Where(r =>
```
minichma: Probably just a matter of taste: For simplicity I'd rather place the filter *after* the `ToList()`. The idea is that _before_ the `ToList()` we only deal with the individual sequences of occurrences, but not with the occurrences themselves, which is done only _after_ the `ToList()`. If the filter is placed here, the effect is the same, but the code is less linear to read. I suggest to place it after the `OrderedMergeMany()` and before the `OrderedDistinct()`.

Ical.Net/Calendar.cs
minichma: Actually the question is legit. The sequence is ordered by StartTime, true, but `OrderedDistinct()` uses `Ocurrences.Equals()` to compare for equality, which compares *more* than just the `StartTime`, i.e. also the `Source`. But the sequence is *not* ordered by `Source`, so this could become a problem.

Actually the real core of the problem we are dealing with here is a shortcoming (or misuse) of `Occurrences.Equals`, which compares the `Source` property via `UniqueComponent.Equals` under the hood, which in turn only looks at the objects' `Uid`. In case of `VEVENT`s with `RECURRENCE-ID` there are multiple objects having the same `UID`. In the case of this issue they also have the same `DTSTART`, so the resulting `Occurrence` instances are considered equal although they aren't. `OrderedDistinct()` will therefore, more or less randomly, return only one of both (except if there are even more entries with the same StartDate and the other issue kicks in that the sequence is not properly ordered due to the `Source` property). As a consequence the filter for the `RECURRENCE-ID` that has formerly been placed *after* the `OrderedDistinct()` couldn't see it.

While this change is a significant improvement, it is not a final fix for that problem. Actually it should be clearly defined how `Occurrence` instances are compared for equality and I'd say this should not happen based on just the UID. Actually I'd say that two Occurrences should be considered equal only if the `Source`s are equal by reference.

To keep the issues apart I created #865, #866.

minichma (dismissed):
Wow, this is a subtle one. Please see which of my comments make sense. I think the issue is just a symptom for problems that go deeper. Created follow-up tickets #865, #866 so we can keep things apart.

Ical.Net/Calendar.cs
minichma: Btw, due to #455 it might be a good idea to keep this filter here, or at least before the OrderedMerge.

Ical.Net.Tests/RecurrenceTests.cs
axunonb: Will flag the tests with Category `RECURRENCE-ID" and add a comment referencing to the test covering the #863

Ical.Net/Calendar.cs
axunonb: Fully agree

Ical.Net/Calendar.cs
axunonb: Will leave the position unchanged

minichma (approved):
Very nice!

axunonb:
@minichma Thanks for your support and review

minichma:
@axunonb urw

PR #870, opened 2025-09-29
feat: `RecurringComponent` supports `RECURRENCE-ID` with optional `RANGE` parameter

axunonb:
`VTimeZoneInfo` and `RecurringComponent`:
* Marked `CalDateTime? RecurrenceId` as obsolete in favor of `RecurrenceId? RecurrenceInstance`, but both properties can still be used in parallel, when the `RANGE` parameter is not required
* Added new data type `RecurrenceIdentifier`
* Added new  `RecurrenceIdentifierSerializer`

Supports #455 - implementation for recurrence not included

### DIFF
--- a/Ical.Net/CalendarComponents/RecurringComponent.cs
+++ b/Ical.Net/CalendarComponents/RecurringComponent.cs
@@ -10,7 +10,6 @@ using System.Runtime.Serialization;
 using Ical.Net.DataTypes;
 using Ical.Net.Evaluation;
 using Ical.Net.Proxies;
-using Ical.Net.Utility;
 
 namespace Ical.Net.CalendarComponents;
 
@@ -113,9 +112,30 @@ public abstract class RecurringComponent : UniqueComponent, IRecurringComponent
         set => Properties.Set("RRULE", value);
     }
 
+    /// <summary>
+    /// Gets or sets the recurrence identifier for a specific instance of a recurring event.
+    /// </summary>
+    /// <remarks>Use <see cref="RecurrenceInstance"/> instead, which
+    /// supports the RANGE parameter for recurring events.</remarks>
+    [Obsolete("Use RecurrenceInstance instead, which supports the RANGE parameter.")]
     public virtual CalDateTime? RecurrenceId
     {
-        get => Properties.Get<CalDateTime>("RECURRENCE-ID");
+        get => RecurrenceInstance?.Range == RecurrenceRange.ThisInstance ? RecurrenceInstance.StartTime : null;
+        set => RecurrenceInstance = value is null ? null : new RecurrenceId(value, RecurrenceRange.ThisInstance);
+    }
+
+    /// <summary>
+    /// Gets or sets the recurrence identifier for a specific instance of a recurring event.
+    /// <para/>
+    /// The <see cref="RecurrenceId.Range"/> sets the scope of the recurrence instance:
+    /// With <see cref="RecurrenceRange.ThisInstance"/>, the instance is limited to the specific
+    /// occurrence identified by the <see cref="RecurrenceId.StartTime"/>.<br/>
+    /// With <see cref="RecurrenceRange.ThisAndFuture"/>, the instance applies to the specified
+    /// <see cref="RecurrenceId.StartTime"/> and all future occurrences.
+    /// </summary>
+    public virtual RecurrenceId? RecurrenceInstance
+    {
+        get => Properties.Get<RecurrenceId>("RECURRENCE-ID");
         set => Properties.Set("RECURRENCE-ID", value);
     }
 
--- a/Ical.Net/CalendarComponents/RecurringComponent.cs
+++ b/Ical.Net/CalendarComponents/RecurringComponent.cs
@@ -115,27 +115,27 @@ public abstract class RecurringComponent : UniqueComponent, IRecurringComponent
     /// <summary>
     /// Gets or sets the recurrence identifier for a specific instance of a recurring event.
     /// </summary>
-    /// <remarks>Use <see cref="RecurrenceInstance"/> instead, which
+    /// <remarks>Use <see cref="RecurrenceIdentifier"/> instead, which
     /// supports the RANGE parameter for recurring events.</remarks>
-    [Obsolete("Use RecurrenceInstance instead, which supports the RANGE parameter.")]
+    [Obsolete("Use RecurrenceIdentifier instead, which supports the RANGE parameter.")]
     public virtual CalDateTime? RecurrenceId
     {
-        get => RecurrenceInstance?.Range == RecurrenceRange.ThisInstance ? RecurrenceInstance.StartTime : null;
-        set => RecurrenceInstance = value is null ? null : new RecurrenceId(value, RecurrenceRange.ThisInstance);
+        get => RecurrenceIdentifier?.Range == RecurrenceRange.ThisInstance ? RecurrenceIdentifier.StartTime : null;
+        set => RecurrenceIdentifier = value is null ? null : new RecurrenceIdentifier(value, RecurrenceRange.ThisInstance);
     }
 
     /// <summary>
     /// Gets or sets the recurrence identifier for a specific instance of a recurring event.
     /// <para/>
-    /// The <see cref="RecurrenceId.Range"/> sets the scope of the recurrence instance:
+    /// The <see cref="RecurrenceIdentifier.Range"/> sets the scope of the recurrence instance:
     /// With <see cref="RecurrenceRange.ThisInstance"/>, the instance is limited to the specific
-    /// occurrence identified by the <see cref="RecurrenceId.StartTime"/>.<br/>
+    /// occurrence identified by the <see cref="RecurrenceIdentifier.StartTime"/>.<br/>
     /// With <see cref="RecurrenceRange.ThisAndFuture"/>, the instance applies to the specified
-    /// <see cref="RecurrenceId.StartTime"/> and all future occurrences.
+    /// <see cref="RecurrenceIdentifier.StartTime"/> and all future occurrences.
     /// </summary>
-    public virtual RecurrenceId? RecurrenceInstance
+    public virtual RecurrenceIdentifier? RecurrenceIdentifier
     {
-        get => Properties.Get<RecurrenceId>("RECURRENCE-ID");
+        get => Properties.Get<RecurrenceIdentifier>("RECURRENCE-ID");
         set => Properties.Set("RECURRENCE-ID", value);
     }
 
--- a/Ical.Net/DataTypes/RecurrenceId.cs
+++ b/Ical.Net/DataTypes/RecurrenceId.cs
@@ -1,0 +1,63 @@
+﻿//
+// Copyright ical.net project maintainers and contributors.
+// Licensed under the MIT license.
+//
+
+namespace Ical.Net.DataTypes;
+
+/// <summary>
+/// Represents the identifier for a specific instance of a recurring event.
+/// </summary>
+/// <example>
+/// RECURRENCE-ID:20250401T133000Z: The instance for this date only
+/// RECURRENCE-ID;RANGE=THISANDFUTURE:20250401T133000Z: This specifies the instance for this date and all future instances.
+/// </example>
+/// <remarks>
+/// This class is used to uniquely identify a particular occurrence within a recurring series. The
+/// identifier consists of a date and a recurrence range, which together specify the instance.
+/// </remarks>
+public class RecurrenceId : ICalendarParameterCollectionContainer
+{
+    /// <summary>
+    /// Initializes a new instance with the specified start time and recurrence range.
+    /// </summary>
+    /// <param name="start">The start time of the recurrence instance.</param>
+    /// <param name="range">The recurrence range that defines the scope of the instance.
+    /// If <paramref name="range"/> is <see langword="null"/>, the default value
+    /// <see cref="RecurrenceRange.ThisInstance"/> is used.</param>
+    public RecurrenceId(CalDateTime start, RecurrenceRange? range = null)
+    {
+        StartTime = start;
+        Range = range ?? RecurrenceRange.ThisInstance;
+        Parameters = new ParameterList();
+    }
+
+    /// <summary>
+    /// Gets or sets the start date and time of the specific instance within the recurring series
+    /// that this identifier refers to and that should get overridden.
+    /// </summary>
+    public CalDateTime StartTime { get; set; }
+
+    /// <summary>
+    /// Gets or sets the recurrence range that determines the scope of the recurrence pattern.
+    /// </summary>
+    public RecurrenceRange Range { get; set; }
+
+    public IParameterCollection Parameters { get; set; }
+}
+
+/// <summary>
+/// The range of recurrence instances that a <see cref="RecurrenceId"/> applies to.
+/// </summary>
+public enum RecurrenceRange
+{
+    /// <summary>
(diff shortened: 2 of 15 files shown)


Ical.Net/Serialization/PropertySerializer.cs
```diff
@@ -89,7 +89,7 @@ private void SerializeValue(StringBuilder result, ICalendarProperty prop, object
 
         // Get the list of parameters we'll be serializing
         var parameterList =
-            (IList<CalendarParameter>?)(value as ICalendarDataType)?.Parameters
+            (IList<CalendarParameter>?)(value as ICalendarParameterCollectionContainer)?.Parameters
```
axunonb, line 92: `ICalendarDataType` implements (the very light-weight) `ICalendarParameterCollectionContainer`

Ical.Net/DataTypes/RecurrenceId.cs
```diff
@@ -0,0 +1,63 @@
...
+/// </summary>
+/// <example>
+/// RECURRENCE-ID:20250401T133000Z: The instance for this date only
+/// RECURRENCE-ID;RANGE=THISANDFUTURE:20250401T133000Z: This specifies the instance for this date and all future instances.
+/// </example>
+/// <remarks>
+/// This class is used to uniquely identify a particular occurrence within a recurring series. The
+/// identifier consists of a date and a recurrence range, which together specify the instance.
+/// </remarks>
+public class RecurrenceId : ICalendarParameterCollectionContainer
```
axunonb: Should `IParameterProvider` on `RecurrenceIdSerializer` be preferred to get parameters?

axunonb:
@minichma How about going the first step in terms of `RECURRENCE-ID;RANGE=THISANDFUTURE` this way?

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -113,9 +112,30 @@ public virtual IList<RecurrencePattern> RecurrenceRules
...
+    /// <summary>
+    /// Gets or sets the recurrence identifier for a specific instance of a recurring event.
+    /// <para/>
+    /// The <see cref="RecurrenceId.Range"/> sets the scope of the recurrence instance:
+    /// With <see cref="RecurrenceRange.ThisInstance"/>, the instance is limited to the specific
+    /// occurrence identified by the <see cref="RecurrenceId.StartTime"/>.<br/>
+    /// With <see cref="RecurrenceRange.ThisAndFuture"/>, the instance applies to the specified
+    /// <see cref="RecurrenceId.StartTime"/> and all future occurrences.
+    /// </summary>
+    public virtual RecurrenceId? RecurrenceInstance
```
minichma: I feel the name `RecurrenceInstance` is misleading. Not sure, what a good name would be (other than `RecurrenceId`). Maybe `RecurrenceIdentifier`? Somewhat long, but that's close to how it is named in the rfc.

Ical.Net/DataTypes/RecurrenceId.cs
minichma: Should probably have Equals/GetHashCode

Ical.Net/Serialization/DataTypes/RecurrenceIdSerializer.cs
```diff
@@ -0,0 +1,95 @@
...
+        if (obj is not RecurrenceId recurrenceId)
+        {
+            return null;
+        }
+        
+        var factory = GetService<ISerializerFactory>();
+        var dtSerializer = factory.Build(typeof(CalDateTime), SerializationContext) as DateTimeSerializer;
+
+        recurrenceId.Parameters.AddRange(dtSerializer!.GetParameters(recurrenceId.StartTime));
+        if (recurrenceId.Range == RecurrenceRange.ThisAndFuture)
```
minichma: fail on unknown ones?

Ical.Net/Serialization/SerializerFactory.cs
```diff
@@ -35,61 +32,59 @@ public SerializerFactory()
...
             return null;
         }
-        ISerializer s;
-
         if (typeof(Calendar).IsAssignableFrom(objectType))
         {
-            s = new CalendarSerializer(ctx);
+            return new CalendarSerializer(ctx);
         }
-        else if (typeof(ICalendarComponent).IsAssignableFrom(objectType))
```
minichma, line 44: This would be a perfect fit for pattern matching nowadays.

Ical.Net/VTimeZoneInfo.cs
```diff
@@ -134,9 +135,30 @@ public virtual IList<RecurrencePattern> RecurrenceRules
...
+    /// <summary>
+    /// Gets or sets the recurrence identifier for a specific instance of a recurring event.
+    /// <para/>
+    /// The <see cref="RecurrenceId.Range"/> sets the scope of the recurrence instance:
+    /// With <see cref="RecurrenceRange.ThisInstance"/>, the instance is limited to the specific
+    /// occurrence identified by the <see cref="RecurrenceId.StartTime"/>.<br/>
+    /// With <see cref="RecurrenceRange.ThisAndFuture"/>, the instance applies to the specified
+    /// <see cref="RecurrenceId.StartTime"/> and all future occurrences.
+    /// </summary>
+    public virtual RecurrenceId? RecurrenceInstance
```
minichma: Consider naming?

minichma (commented):
LGTM! Based on that change the actual filtering should be quite straight forward. Let me know if you'd like me to assist there.

axunonb:
Refactoring from review

Renamed the `RecurrenceId` class to `RecurrenceIdentifier`
* Removed `ICalendarParameterCollectionContainer` because custom parameters should not be allowed here
* Added `IComparable<RecurrenceIdentifier>`

Renamed the `RecurrenceIdSerializer` class to `RecurrenceIdentifierSerializer`
* Added `IParameterProvider`

Added class `ParameterProviderHelper` that contains methods
* `GetCalDateTimeParameters`
* `GetRecurrenceIdentifierParameters`

Miscellaneous
* Update `DateTimeSerializer.GetParameters` to call `ParameterProviderHelper.GetCalDateTimeParameters`
* Update `IParameterProvider.GetParameters` to accept a nullable object
* Update `SerializerFactory.Build` to use pattern matching

Note: Did not investigate possible code duplication in `RecurringComponent` / `VTimeZoneInfo` as they existed before, too.

> Based on that change the actual filtering should be quite straight forward. Let me know if you'd like me to assist there.

Oh yes, that would be more than welcome.

Ical.Net/DataTypes/RecurrenceId.cs
axunonb: Now implements `IComparable<RecurrenceIdentifier>`

Ical.Net/Serialization/DataTypes/RecurrenceIdSerializer.cs
axunonb: log warning

minichma:
> Based on that change the actual filtering should be quite straight forward.

Having a closer look at the RFC, my assumption might be mistaken. 😆 Anyhow, will give it a try.

minichma:
Hm, my naive assumption was as follows, but seems to be wrong:
1. Generate recurrences of original and new VEVENT independently
2. If
    * the range is `THISANDFUTURE` do something like 'oldSequence.TakeWhile(item => item != recurrenceId).OrderedMerge(newSequence)`
    * else do something like 'oldSequence.Where(item => item != recurrenceId).OrderedMerge(newSequence)`

My assumption was that the new component may or may not have an RRULE or RDATE. If it hasn't, it produces just a single occurrence. But looking at [the RFC](https://www.rfc-editor.org/rfc/rfc5545.html#section-3.8.4.4)

> When the given recurrence instance is
      rescheduled, all subsequent instances are also rescheduled by the
      same time difference.  For instance, if the given recurrence
      instance is rescheduled to start 2 hours later, then all
      subsequent instances are also rescheduled 2 hours later.

this sounds more like the new VEVENT (the one with the RECURRENCE-ID) doesn't have its own RRULE/RDATE/EXDATE properties but inherits those of the original one, only the offset is considered, right? If so, several questions arise:

* What if the new RRULE *does* have an RRULE? Does it replace the original one?
* What if the new component's DTSTART doesn't match the original RRULE? Probably undefined.
* What about RDATEs of the original component? Are they transposed too? (while doing this for RRULEs should be doable, for RDATEs its somewhat more involved. What if the TZ changes, etc.)
* What about EXDATEs/EDRULEs?
* What if the original component has RDATEs but no RRULEs and the new one has RRULEs (or the other way round)?
* What if the new DTSTART is in a different TZ with different DST changes. How to transpose RDATEs from the original sequence?
* ...

Had a look what Google Calendar and Outlook would export in case of modified time series.
* Google Calendar: Didn't get it to export `RECURRENCE-ID` neither for single days being shifted nor for 'this and after'
* Outlook: Used 'RECURRENCE-ID' when modifying single dates. When modifying 'this and after', the original sequence had a 'UNTIL' added to the RRULE and an additional, independent VEVENT (with a different UID but the same title) was created


Any comments are welcome.
This is what ChatGPT says:

Short, precise take based on RFC 5545 and vendor docs:

### Mental model (what to implement)

1. **Master set:** Build the recurrence set **S** from the *master* VEVENT (same UID, **no** RECURRENCE-ID): expand `DTSTART` + all `RRULE` + `RDATE`, then remove `EXDATE`. Order by effective local time (respect `VTIMEZONE`). ([[icalendar.org](https://icalendar.org/iCalendar-RFC-5545/3-8-5-3-recurrence-rule.html?utm_source=chatgpt.com)][1])
2. **Overrides:** For each VEVENT with `RECURRENCE-ID` (same UID):

   * **No `RANGE` param:** override **only** the single instance whose original start equals `RECURRENCE-ID`. Replace properties given on the override; others inherit from the master. Remove the original instance from **S**. ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])
   * **`RANGE=THISANDFUTURE`:** compute Δ = (`override.DTSTART` − `RECURRENCE-ID`, in the `RECURRENCE-ID`’s timezone/value type), then **shift every subsequent instance** in **S** by exactly Δ. (Also shift `DTEND`/`DURATION` by Δ unless the override sets them explicitly.) ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])
3. **Conflict resolution:** If multiple overrides target the same instance, pick the one with highest `SEQUENCE` (or latest `LAST-MODIFIED` as tie-break). ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])

### Answers to your questions

* **Does the override (the component with `RECURRENCE-ID`) carry its own `RRULE/RDATE/EXDATE`?**
  **No, not in the RFC’s model.** The recurrence set comes from the master; overrides change properties of one instance (or the tail with `RANGE=THISANDFUTURE`). Treat recurrence-defining props on an override as **non-standard**; safest is to ignore them. (Outlook exports exception instances with `RECURRENCE-ID` but not their own `RRULE`.) ([[icalendar.org](https://icalendar.org/iCalendar-RFC-5545/3-8-5-3-recurrence-rule.html?utm_source=chatgpt.com)][1])
* **If an override *does* include an `RRULE`, does it replace the original?**
  **Undefined/non-standard.** Some clients may treat this as a “split series,” but RFC 5545 doesn’t define that behavior. Prefer: ignore and log. ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])
* **What if the override’s `DTSTART` doesn’t match any generated instance from the master?**
  Then it’s invalid per spec (the `RECURRENCE-ID` must equal the *original* instance `DTSTART`). Robust parsers often ignore it. ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])
* **Do `RDATE` instances get transposed under `THISANDFUTURE`?**
  **Yes.** “Subsequent instances” means subsequent members of the recurrence set—i.e., including future `RDATE` occurrences—shift by the same Δ. Apply `EXDATE` filtering **after** shifting. (Mind time zones/DST using the `RECURRENCE-ID`’s value type/TZ.) ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])
* **EXDATEs/EXRULEs?**
  `EXDATE` continues to exclude instances (post-shift). `EXRULE` is **deprecated** in RFC 5545; don’t generate it, and if present, treat cautiously. ([[iana.org](https://www.iana.org/assignments/icalendar?utm_source=chatgpt.com)][3])
* **What if the master has only `RDATE`s (no `RRULE`), and you do `THISANDFUTURE`?**
  Still apply Δ to all later members of **S** (which, in this case, are just the later `RDATE` instances). ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])

### Implementation sketch (deterministic)

1. Expand **S** from master (bounded window).
2. Index overrides by (`RECURRENCE-ID`, `RANGE`).
3. Apply per-instance overrides (no RANGE).
4. Sort remaining overrides with `RANGE=THISANDFUTURE` by `RECURRENCE-ID` ascending, and for each, compute Δ and shift every `t >= RECURRENCE-ID`.
5. Re-apply `EXDATE` after all shifts; then property-merge (inherit from master; override wins where set).
6. Handle ties with `SEQUENCE`/`LAST-MODIFIED`.

References: RFC 5545 §3.8.4.4 (`RECURRENCE-ID` incl. `RANGE=THISANDFUTURE`), §3.8.5.* (recurrence set construction), §A.3 (deprecated `EXRULE`); Microsoft iCalendar notes confirming exception semantics. ([[datatracker.ietf.org](https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com)][2])

If you want, I can turn this into a small testable expansion/override function (e.g., Python with `dateutil.rrule`) with a few edge-case fixtures.

[1]: https://icalendar.org/iCalendar-RFC-5545/3-8-5-3-recurrence-rule.html?utm_source=chatgpt.com "3.8.5.3. Recurrence Rule"
[2]: https://datatracker.ietf.org/doc/html/rfc5545?utm_source=chatgpt.com "RFC 5545 - Internet Calendaring and Scheduling Core ..."
[3]: https://www.iana.org/assignments/icalendar?utm_source=chatgpt.com "iCalendar Element Registries"

minichma:
As the changes in this PR are an improvement also without having the full evaluation implemented, it might be a good idea to complete this one and create a new one for the rest. This would avoid delaying this one and would remove time pressure. What do you think?

minichma (approved):

axunonb:
First of all thanks for the reviews, which are always inspiring!

> it might be a good idea to complete this one and create a new one for the rest.

Yes, that's what I had in mind as well when mentioning "implementation for recurrence not included".

> Having a closer look at the RFC, my assumption might be mistaken.

Kind of 😉, I'm afraid. Google doesn't support it, and I also understood that we should follow the Microsoft way:
"the original sequence had a 'UNTIL' added to the RRULE and an additional, independent VEVENT (with a different UID but the same title) was created". The UID for the new internal VEVENT is random, so we don't have to care for references to this UID.
(I've been asking Gemini).

Never created a sub issue before, but this topic could a good trial for creating one to https://github.com/ical-org/ical.net/issues/455. I'll create it, and there we could discuss further details.

axunonb:
@minichma  Would #864 and #870 deserve a new release? 5.1.1 or 5.2.0?

Issue #791, opened 2025-05-11
Test cases related to `BYWEEKNO` + `BYMONTHDAY`, etc. could be wrong


minichma:
**Describe the bug**

The following test cases should be reconsidered:

https://github.com/ical-org/ical.net/blob/19cf410f188c1e7ab8b6fbd8b00bfb39387ec24e/Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt#L31-L89

They are related to `BYWEEKNO` in conjunction with other `BY` rules like `BYYEARDAY`, `BYMONTHDAY`, `BYMONTH`. The cases seem to be quite poorly documented in the RFC, but the way they are specified in the test cases and implemented in Ical.Net is likely to be wrong. Actually the test cases assume that BY rules after `BYWEEKNO` would be limiting in those cases, which would probably be the most natural thing, but actually the RFC 5545 'expand' matrix clearly states that all BY rules (except for BYSETPOS, BYDAY) should be expanding in case of `YEARLY` rules. It's somewhat unclear though, how expanding would work in those cases.

The way they are implemented are also responsible for some somewhat weird code

https://github.com/ical-org/ical.net/blob/19cf410f188c1e7ab8b6fbd8b00bfb39387ec24e/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L226-L236

Anyhow, should be reconsidered.

**Environment (please complete the following information):**
- ical.net version: 5.0.0-pre.42

The original test cases have been introduced in #731

axunonb:
```ics
RRULE:FREQ=YEARLY;BYMONTH=12;BYMONTHDAY=1,30;BYWEEKNO=1;UNTIL=20320101 
DTSTART:20241230 
```
For this test case RFC seems to be quite clear:

RFC says: "Week number one of the calendar year is the **first week** that contains at least four (4) days **in that calendar year**". So  week number 1 of 2024 started on January 1, 2024 - not in December. This complies with the "standard interpretation" based on ISO 8601, RFC5545 does not explicitly tell, what "in that calendar year" should mean, though.
The ```RRULE``` generates an empty set of dates for the recurrence instances, because of ```BYWEEKNO=1```.

RFC says: "The recurrence set generated with a "DTSTART" property value that doesn’t match the pattern of the rule is **undefined**."

So I'd say the the expected occurrences 20241230,20251230,20301230,20311230 are not correct - should be empty (or better throw for "undefined"?). Maybe compare again to libical?

[...]

PR #884, opened 2025-11-30
Fixes for `RRULE:FREQ=YEARLY` combined with `BYMONTH`

### LINKED ISSUE #885: `RRULE:FREQ=YEARLY` combined with `BYMONTH` may create wrong occurrences
> `YEARLY` + `BYMONTHDAY` does seem to be wrong though. The test cases are only correct because `BYMONTH` is included.
> 
> See [this test](https://github.com/ical-org/ical.net/blob/4e51391167e4f3889b062a8f9e2f2495b6bb00ac/Ical.Net.Tests/RecurrenceTests.cs#L2007) and the rule for it:
> 
> ```ical
> DTSTART;TZID=US-Eastern:20050409T070000
> RRULE:FREQ=YEARLY;INTERVAL=2;BYDAY=MO,TU;BYMONTHDAY=8,9,10,11,12,13,14
> SUMMARY:Every other year, on Monday and Tuesday on the 2nd week of each month.
> ```
> 
> The summary here and the expected values are wrong. I think `BYMONTHDAY` should expand only for month 4 (from `DTSTART`) because no other month is specified. The test only passes because it only asks for month 1, which just happens to be where `YEARLY` increments to. If the test were changed to start evaluation at `DTSTART`, the results would be:
> ```
> 2005-04-11T07:00:00
> 2005-04-12T07:00:00
> 2007-01-08T07:00:00
> 2007-01-09T07:00:00
> 2009-01-12T07:00:00
> 2009-01-13T07:00:00
> ```
> 
> I would expect the results to be:
> ```
> 2005-04-11T07:00:00
> 2005-04-12T07:00:00
> 2007-04-09T07:00:00
> 2007-04-10T07:00:00
> 2009-04-13T07:00:00
> 2009-04-14T07:00:00
> ```
> 
> Note that rrule.js handles this differently by expanding for every month. There is an [rrule.js issue](https://github.com/jkbrzt/rrule/issues/275) that talks about that being wrong. 

 _Originally posted by @maknapp in [#791](https://github.com/ical-org/ical.net/issues/791#issuecomment-3553325859)_

axunonb:
### `RecurrencePatternEvaluator`:
1. Extended `GetIntervalLowerLimit` calculation:
  * YEARLY + BYMONTH: compute a conservative earliest possible candidate inside the interval
  * YEARLY + BYWEEKNO: shift interval start to the week’s first day
  * default return intervalRefTime
2. Intoduced an `anchorMonth` as an explicit month anchor used when FREQ=YEARLY and BYMONTH is not specified so that BYMONTHDAY expansion/limiting is performed relative to the original DTSTART month rather than the current intervalRefTime month. `anchorMonth` only affects cases where BYMONTH is absent.

### `Evaluator`:
Added a distinction for case `FrequencyType.Yearly`: Is BYWEEKNO present or not.

### `RecurrenceTests`:
Extended `YearlyInterval1()` to test more occurrences and taking latest code changes into account.

See comments in https://github.com/ical-org/ical.net/issues/791#issuecomment-3553325859 for reference
Fixes #885

### DIFF
--- a/Ical.Net/Evaluation/Evaluator.cs
+++ b/Ical.Net/Evaluation/Evaluator.cs
@@ -41,7 +41,18 @@ public abstract class Evaluator : IEvaluator
                     dt = old.AddDays(-old.Day + 1).AddMonths(interval);
                     break;
                 case FrequencyType.Yearly:
-                    dt = old.AddDays(-old.DayOfYear + 1).AddYears(interval);
+                    if (pattern.ByWeekNo.Count != 0)
+                    {
+                        // When a rule uses BYWEEKNO, recurrence enumeration
+                        // is based on week numbers relative to the year.
+                        // So we preserve January 1st when adding years
+                        dt = old.AddDays(-old.DayOfYear + 1).AddYears(interval);
+                    }
+                    else
+                    {
+                        // preserve month/day when adding years
+                        dt = old.AddYears(interval);
+                    }
                     break;
                 default:
                     // Frequency should always be valid at this stage.
--- a/Ical.Net/Evaluation/Evaluator.cs
+++ b/Ical.Net/Evaluation/Evaluator.cs
@@ -41,18 +41,12 @@ public abstract class Evaluator : IEvaluator
                     dt = old.AddDays(-old.Day + 1).AddMonths(interval);
                     break;
                 case FrequencyType.Yearly:
-                    if (pattern.ByWeekNo.Count != 0)
-                    {
-                        // When a rule uses BYWEEKNO, recurrence enumeration
-                        // is based on week numbers relative to the year.
-                        // So we preserve January 1st when adding years
-                        dt = old.AddDays(-old.DayOfYear + 1).AddYears(interval);
-                    }
-                    else
-                    {
-                        // preserve month/day when adding years
-                        dt = old.AddYears(interval);
-                    }
+                    // When a rule uses BYWEEKNO, recurrence enumeration
+                    // is based on week numbers relative to the year.
+                    // So we preserve January 1st when adding years; otherwise, preserve month/day.
+                    dt = (pattern.ByWeekNo.Count != 0)
+                        ? old.AddDays(-old.DayOfYear + 1).AddYears(interval)
+                        : old.AddYears(interval);
                     break;
                 default:
                     // Frequency should always be valid at this stage.
--- a/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
+++ b/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
@@ -540,38 +540,40 @@ public class RecurrencePatternEvaluator : Evaluator
 
     private static IEnumerable<CalDateTime> GetMonthDayVariantsLimited(IEnumerable<CalDateTime> dates, RecurrencePattern pattern, int anchorMonth)
     {
+        // Helper that checks whether the given date matches any BYMONTHDAY entry
+        // taking negative values into account (relative to the month's length).
+        static bool MatchesAnyMonthDay(CalDateTime date, IEnumerable<int> monthDays)
+        {
+            var daysInMonth = Calendar.GetDaysInMonth(date.Year, date.Month);
+            foreach (var monthDay in monthDays)
+            {
+                var byMonthDay = monthDay > 0 ? monthDay : (daysInMonth + monthDay + 1);
+                if (date.Day == byMonthDay)
+                    return true;
+            }
+            return false;
+        }
+
         foreach (var date in dates)
         {
-            // If BYMONTH is specified, limit according to it; otherwise limit to the anchor month.
             if (pattern.ByMonth.Count > 0)
             {
-                var daysInMonth = Calendar.GetDaysInMonth(date.Year, date.Month);
-                foreach (var monthDay in pattern.ByMonthDay)
-                {
-                    var byMonthDay = (monthDay > 0) ? monthDay : (daysInMonth + monthDay + 1);
-                    if (date.Day == byMonthDay && pattern.ByMonth.Contains(date.Month))
-                    {
-                        yield return date;
-                        break;
-                    }
-                }
+                // If BYMONTH is specified, the date must be in one of those months
+                // and match a BYMONTHDAY value.
+                if (!pattern.ByMonth.Contains(date.Month))
+                    continue;
+
+                if (MatchesAnyMonthDay(date, pattern.ByMonthDay))
+                    yield return date;
             }
             else
             {
-                // Only consider dates that are in the anchor month and match the BYMONTHDAY.
+                // When BYMONTH is not specified, only consider dates in the anchor month.
                 if (date.Month != anchorMonth)
                     continue;
 
-                var daysInMonth = Calendar.GetDaysInMonth(date.Year, date.Month);
-                foreach (var monthDay in pattern.ByMonthDay)
-                {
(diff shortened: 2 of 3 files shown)


maknapp (approved):

axunonb (commented):
@maknapp Thanks for the review

Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -41,7 +41,12 @@ protected void IncrementDate(ref CalDateTime dt, RecurrencePattern pattern, int
                     dt = old.AddDays(-old.Day + 1).AddMonths(interval);
                     break;
                 case FrequencyType.Yearly:
-                    dt = old.AddDays(-old.DayOfYear + 1).AddYears(interval);
+                    // When a rule uses BYWEEKNO, recurrence enumeration
+                    // is based on week numbers relative to the year.
+                    // So we preserve January 1st when adding years; otherwise, preserve month/day.
```
minichma, line 46: Maybe more precisely

`So we preserve the weekday when using BYWEEKNO and preserve month/day otherwise`?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -210,16 +220,77 @@ private IEnumerable<CalDateTime> EnumerateDates(CalDateTime originalDate, CalDat
...
+                // Determine the earliest month we could possibly generate for this interval.
+                var month = pattern.ByMonth.Min();
+
+                // Determine an appropriate day in the month.
+                var daysInMonth = Calendar.GetDaysInMonth(year, month);
+                int day;
+                if (pattern.ByMonthDay.Count > 0)
+                {
+                    var md = pattern.ByMonthDay.Min();
+                    day = md > 0 ? Math.Min(md, daysInMonth) : Math.Max(1, daysInMonth + md + 1);
```
minichma, line 267: Probably not correct if positive/negative values are mixed, e.g. `BYMONTHDAY=2,-2`.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -210,16 +220,77 @@ private IEnumerable<CalDateTime> EnumerateDates(CalDateTime originalDate, CalDat
...
-    private static CalDateTime GetIntervalLowerLimit(CalDateTime intervalRefTime, RecurrencePattern pattern)
+    private static CalDateTime GetIntervalLowerLimit(CalDateTime intervalRefTime, RecurrencePattern pattern, CalDateTime originalDate)
     {
-        var intervalLowerLimit = intervalRefTime;
-        if ((pattern.Frequency == FrequencyType.Yearly) && (pattern.ByWeekNo.Count != 0))
-            intervalLowerLimit = GetFirstDayOfWeekDate(intervalRefTime, pattern.FirstDayOfWeek);
-        return intervalLowerLimit;
+        switch (pattern)
+        {
+            case { Frequency: FrequencyType.Yearly, ByMonth.Count: > 0, ByWeekNo.Count: 0 }:
```
minichma, line 238: I feel this whole method is not perfectly intuitive, maybe we should improve the naming or the description. I.e. if I remember correctly, it is there for a minor performance improvement, i.e. to avoid iterating an extra year at the end of the sequence where this can be avoided, maybe also to avoid running into a `EvaluationLimitExceededException` in certain cases. However, for some reason this is only implemented for the case of BYWEEKNO. I assume this has historic reasons and might be related to the fact that in case of BYWEEKNO we might need to step back to the previous interval (at least when implemented the way it is today). For all other cases we don't do this extra calculation, which brings up the question whether we should do it for this new special case. I'd rather say no, i.e. I'd rather recommend implementing it for *all* cases (which I'd rather try to avoid) or only for the minimum set of cases necessary. But I could also easily miss something.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -162,10 +162,20 @@ private IEnumerable<CalDateTime> EnumerateDates(CalDateTime originalDate, CalDat
...
+            if (searchEndDate < lowerLimit)
                 break;
 
-            var candidates = GetCandidates(intervalRefTime, pattern, expandBehavior);
+            // Use the original DTSTART month as the anchor for YEARLY rules when BYMONTH is not specified.
+            // This ensures BYMONTHDAY expansion/limiting is performed relative to the DTSTART month
+            // (the expected behavior for yearly rules without an explicit BYMONTH), instead of the
+            // current interval reference month.
+            var anchorMonth = pattern is { Frequency: FrequencyType.Yearly, ByMonth.Count: 0 }
+                ? originalDate.Month
```
minichma, line 175: When changing this to return `intervalRefTime.Month` in both cases, all tests still run through. Would be great to have a test case where this distinction is essential.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -465,22 +554,45 @@ private static IEnumerable<CalDateTime> GetMonthDayVariants(IEnumerable<CalDateT
...
+                if (!pattern.ByMonth.Contains(date.Month))
+                    continue;
+
+                if (MatchesAnyMonthDay(date, pattern.ByMonthDay))
+                    yield return date;
+            }
+            else
+            {
+                // When BYMONTH is not specified, only consider dates in the anchor month.
+                if (date.Month != anchorMonth)
```
minichma, line 591: Would be good to add some more context on why this was added. Not sure it is fully correct. Consider the following test case, which fails now (could be added to `RecurrenceTestCases.txt` if considered correct).

```ics
RRULE:FREQ=YEARLY;BYYEARDAY=-1,1;BYMONTHDAY=-1,1;COUNT=3
DTSTART:20250101
INSTANCES:20250101,20251231,20260101
```

minichma (commented):
I'm obviously a little late as you just merged already. Anyhow, here are my findings.

axunonb:
It's never too late to improve, thanks a lot

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: Yes, I consider the test correct. It worked before this commit. Sadly we didn't have the test before - so thanks a lot.
I wonder whether using an anchor month has even more side effects than this one.
Anyway, removing `if (date.Month != anchorMonth) continue;` makes sense, and all tests including the new one above pass.

Ical.Net/Evaluation/Evaluator.cs
axunonb: Yes, this is more precise.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: The anchorMonth was one of the ideas introduced in a very first draft. It was only partly helpful. When investigating further, anchorMonth became fully redundant - which I oversaw. Excellent finding, Please keep up these reviews.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: > For all other cases we don't do this extra calculation, which brings up the question whether we should do it for this new special case. I'd rather say no, i.e. I'd rather recommend implementing it for all cases (which I'd rather try to avoid)

Yes, agree, that's not clean. Covering special cases here is a risk. Not sure, which further special cases we can expect besides those just implemented. For the moment, they do their job.

PR #887, opened 2025-12-01
Implement `BYDAY` with offset and limiting behavior

### LINKED ISSUE #782: `BYDAY` with offset and limiting behavior not implemented
In the [RFC 5545 expand matrix](https://www.rfc-editor.org/rfc/rfc5545.html#section-3.3.10) Note 1 and Note 2 it is specified that `BYDAY` should apply limiting behavior if `BYMONTHDAY` or `BYYEARDAY` is present. This is not properly implemented in Ical.Net. The following test cases therefore fail (double-check the tests for correctness!):

```ics
# BYDAY with offset with limiting behaviour WITH MONTHLY;BYMONTHDAY (expand matrix Note 1)
RRULE:FREQ=MONTHLY;BYMONTHDAY=1,8;BYDAY=1MO,2TU;COUNT=3
DTSTART:20260601
INSTANCES:20260601,20260908,20261208

# BYDAY with offset and limiting behavior with YEARLY/BYMONTHDAY (expand matrix Note 2)  
#
# RFC 5545 is somewhat ambiguous about this case.
#
# * The offset appears to correspond to the position within the year because BYMONTH is NOT present. However,
#   the RFC mentions `BYMONTH` in a contradictory way. It seems reasonable to assume the second mention is
#   the incorrect one.
#   > The numeric value in a
#     > BYDAY rule part with the FREQ rule part set to YEARLY corresponds
#     > to an offset within the month when the BYMONTH rule part is
#     > present, and corresponds to an offset within the year when the
#     > BYWEEKNO or BYMONTH rule parts are present.
#
# * The month seems to be fixed to June because BYMONTH is not present, and BYDAY is only limiting.
#   To apply the limit, the actual dates must be determined first, which requires the month, so it must be inferred from DTSTART.
#   However, there is some room for alternative interpretations. I.e., one could argue that the day is fully specified
#   via `BYDAY`, making it unnecessary to derive the month from DTSTART. Since BYDAY is limiting, we assume the month is taken from DTSTART.
RRULE:FREQ=YEARLY;BYMONTHDAY=1,8;BYDAY=22MO,23TU,25MO,36TU;COUNT=3
DTSTART:20260601
INSTANCES:20260601,20270608,20320608

# BYDAY with offset with limiting behaviour with YEARLY/BYMONTH/BYMONTHDAY (expand matrix Note 2)
# The offset corresponds to the position within the month, because BYMONTH is present.
# > The numeric value in a
# > BYDAY rule part with the FREQ rule part set to YEARLY corresponds
# > to an offset within the month when the BYMONTH rule part is present, ...
RRULE:FREQ=YEARLY;BYMONTH=6,7,8,9,10,11,12;BYMONTHDAY=1,8;BYDAY=1MO,2TU;COUNT=3
DTSTART:20260601
INSTANCES:20260601,20260908,20261208

# BYDAY with limiting behaviour with YEARLY/BYYEARDAY (expand matrix Note 2)
RRULE:FREQ=YEARLY;BYYEARDAY=152,173,251,272,321,342;BYDAY=22MO,26MO,36TU,37TU,49TU;COUNT=3
DTSTART:20260601
INSTANCES:20260601,20260908,20261208
```

This seems to be a known limitation according to this comment:

https://github.com/ical-org/ical.net/blob/c7faa31892b665783220c8927e5d97032332a1c3/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L565

Version: 5.0.0-pre.42

### LINKED ISSUE #888: Use `RecurrenceTestCases.txt` as input for recurrence tests
these tests would be a great fit for `RecurrenceTestCases.txt`, which is somewhat more declarative and less dependent on our specific technology. Tests included there could more easily run in other contexts too, e.g. when comparing to other technologies. In libical. Particularly libical uses the same format, which allowed us to reuse their tests quite easily. Could be a chance to give back at some point.

_Originally posted by @minichma in https://github.com/ical-org/ical.net/pull/887#discussion_r2580030141_

axunonb:
The following `RRULE` cases were not implemented (see the diff for newly added unit tests for details):

1. YEARLY + BYMONTH + numeric BYDAY offsets
   - Pattern: `FREQ=YEARLY;BYMONTH=6,9;BYDAY=2MO`
   - Semantics: treat numeric BYDAY as “nth weekday inside each BYMONTH” (produce the 2nd Monday of each specified month).

2. YEARLY + numeric BYDAY without BYMONTH
   - Pattern: `FREQ=YEARLY;BYDAY=20MO`
   - Semantics: interpret numeric BYDAY as “nth weekday of the YEAR” (the 20th Monday of the year).

3. YEARLY + BYMONTH + negative numeric BYDAY
   - Pattern: `FREQ=YEARLY;BYMONTH=6,9;BYDAY=-1SU`
   - Semantics: negative offsets select from the end of the month (e.g. `-1SU` → last Sunday in each BYMONTH).

4. Implemented the per-month offset expansion and preserved BYWEEKNO / BYMONTH compatibility and negative-offset

These behaviors are closely related and required special handling in `RecurrencePatternEvaluator.GetAbsWeekDays` and `GetAbsWeekDaysYearlyPerMonthOffsets`.

[Edit]
Relevant unit tests in `RecurrenceTestCases.txt` contain a disclaimer regarding RFC5545 interpretation in ical.net

Fixes #782
Closes #888


Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -4807,4 +4806,132 @@ public void GetOccurrences_WithMultipleOverridesForSameRecurrenceId_ShouldUseLat
...
+                    RRULE:FREQ=YEARLY;BYMONTHDAY=1,8;BYDAY=22MO,23TU,25MO,36TU;COUNT=3
+                    END:VEVENT
+                    END:VCALENDAR
+                    """;
+
+        var cal = Calendar.Load(ics)!;
+        var expected = new[]
+        {
+            new Period(new CalDateTime(2026, 6, 1), Duration.FromDays(1)),
+            new Period(new CalDateTime(2027, 6, 8), Duration.FromDays(1)),
```
minichma: Shouldn't 2026-09-08 be returned too?

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -4807,4 +4806,132 @@ public void GetOccurrences_WithMultipleOverridesForSameRecurrenceId_ShouldUseLat
...
+    public void ByMonthDay_With_ByDay_SimpleCase()
+    {
+        const string tzId = "Europe/Berlin";
+        var ics = """
+                   BEGIN:VCALENDAR
+                   VERSION:2.0
+                   BEGIN:VEVENT
+                   DTSTART;TZID=Europe/Berlin:20250913T090000
+                   DURATION:PT1H
+                   RRULE:FREQ=MONTHLY;BYMONTHDAY=13;BYDAY=MO
```
minichma: these tests would be a great fit for `RecurrenceTestCases.txt`, which is somewhat more declarative and less dependent on our specific technology. Tests included there could more easily run in other contexts too, e.g. when comparing to other technologies. In libical. Particularly libical uses the same format, which allowed us to reuse their tests quite easily. Could be a chance to give back at some point.

Ical.Net.Tests/RecurrenceTests.cs
axunonb: https://github.com/ical-org/ical.net/issues/782#issuecomment-2853505427:
>  **The month seems to be fixed to June because BYMONTH is not present, and BYDAY is only limiting.**
>  To apply the limit, the actual dates must be determined first, which requires the month, so it must be inferred from 
>  However, there is some room for alternative interpretations. I.e., one could argue that the day is fully specified
>  via `BYDAY`, making it unnecessary to derive the month from DTSTART. Since BYDAY is limiting, we assume the month is taken from DTSTART.

Following the the first sentence: No. The current implementation has BYDAY **only limiting**, and it can't "escape".

minichma: Oh, its *that* topic, didn't notice at first! The main problem seems to be that the RFC is quite ambiguous. I'll be happy to review the code but given the ambiguity, some aspects might at the end be a matter of taste. Anyhow, I think the tests should be marked as 'assumption based' very clearly, so future readers understand that the tests represent our interpretation of the RFC but other interpretations might be valid too.

Unfortunately it might take a while until I find the time for a thorough review.

Ical.Net.Tests/RecurrenceTests.cs
axunonb: > `RecurrenceTestCases.txt`, which is somewhat more declarative

Indeed, will use a separate PR for that

Ical.Net.Tests/RecurrenceYearlyWithOffsetsTests.cs
```diff
@@ -0,0 +1,379 @@
...
+            // 2nd Monday of each BYMONTH
+            RecurrenceRules = [new RecurrencePattern("FREQ=YEARLY;BYMONTH=6,9;BYDAY=2MO;COUNT=4")]
+        };
+
+        var cal = new Calendar();
+        cal.Events.Add(evt);
+
+        CalDateTime[] expected =
+        [
+            CalCalc.GetNthWeekdayOfMonth(2026, 6, DayOfWeek.Monday, 2),
```
maknapp: Avoid calculating expected values? It is much easier to debug tests with the exact dates listed.

maknapp:
FYI, I made several tests related to this for a new recurrence evaluator. I ran those tests against this fix and they all pass.

maknapp (dismissed):

axunonb:
@maknapp Thanks for the review and the helpful hint

>  I made several tests related to this for a new recurrence evaluator. I ran those tests against this fix and they all pass.

So it looks like we're on the right way. As @minichma  opened the issue, I'd like have his feedback as well, if he finds the time. He also contributes to [libical](https://github.com/libical/libical).

axunonb:
@minichma Unit test for this PR are now part of `RecurrenceTestCases.txt`

Ical.Net.Tests/RecurrenceTests.cs
axunonb: @minichma  Any chance for a review?

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
```diff
@@ -181,3 +181,94 @@ INSTANCES:20250101,20251231,20260101
...
+# 
+# These 12 tests verify the behavior with the interpretation that
+# * when only BYMONTH is present, the numeric BYDAY offset applies within the month,
+# * when BYWEEKNO is present, the numeric BYDAY offset applies within the week.
+#
+# Disclaimer: Other iCalendar libraries may interpret this differently.
+
+# Yearly BYMONTH numeric BYDAY (2MO) - 2nd Monday in each BYMONTH
+RRULE:FREQ=YEARLY;BYMONTH=6,9;BYDAY=2MO;COUNT=4
+DTSTART:20260601T090000
```
minichma: DTSTART is not in line with the RRULE, which makes the case undefined according to the RFC (similar in other test cases). Unless this test should explicitly cover the undefined case, it would be good to have DTSTART aligned with RRULE, otherwise readers would have to think about whether the deviation is intentional.

The time part could be omitted, which would make the test somewhat more concise.

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
```diff
@@ -181,3 +181,94 @@ INSTANCES:20250101,20251231,20260101
...
+DURATION:PT1H
```
minichma: DURATION could be removed as it doesn't add value to the test (other tests too).

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
```diff
@@ -181,3 +181,94 @@ INSTANCES:20250101,20251231,20260101
...
+
+############################## START ERRATA 1913 TESTS ##############################
+# RFC errata 1913 (https://www.rfc-editor.org/errata_search.php?rfc=1913&amp;eid=1913):
+# 'The numeric value in a BYDAY rule part with the FREQ rule part set to YEARLY corresponds  
+# to an offset within the month when the BYMONTH rule part is present, and corresponds to an 
+# offset within the year when the BYWEEKNO or BYMONTH rule parts are NOT present.'
+# 
+# These 12 tests verify the behavior with the interpretation that
+# * when only BYMONTH is present, the numeric BYDAY offset applies within the month,
+# * when BYWEEKNO is present, the numeric BYDAY offset applies within the week.
```
minichma, line 193: Applying the numeric offset within the week is a reasonable interpretation following the logic of the case that `BYMONTH` is present. As the RFC is quite clear that BYWEEKNO + numeric BYDAY offset is not allowed, it would also be an option to throw in this case, but being lenient often is the better approach.

Another thought is that, as 1 would be the only reasonable integer (even -1 wouldn't be allowed, as every week always has 7 days, so there is no need for -1 as it isn't needed and therefore not allowed for other components like hours, which don't vary in different intervals) there isn't much point in using it, so any use of it would most likely be a usage error anyways. So throwing might reduce the risk of errors in this case, but in this case this might be a matter of taste.

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
```diff
@@ -181,3 +181,94 @@ INSTANCES:20250101,20251231,20260101
...
+# Yearly BYMONTH=3 BYDAY=6MO - invalid: 6th Monday in March doesn't exist
+RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=6MO
+DTSTART:20240101T090000
+DURATION:PT1H
+START-AT:20240101
+EXCEPTION:Ical.Net.Evaluation.EvaluationOutOfRangeException
+EXCEPTION-STEP:Enumeration
+
+# Yearly BYWEEKNO=10 BYDAY=1MO - valid: Monday of ISO week 10 each year
+RRULE:FREQ=YEARLY;BYWEEKNO=10;BYDAY=1MO;COUNT=3
```
minichma, line 227: As outlined above, maybe rather throw in this case?

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
```diff
@@ -181,3 +181,94 @@ INSTANCES:20250101,20251231,20260101
...
+
+# Yearly BYWEEKNO=10 BYDAY=2MO - invalid: 2nd Monday in a week doesn't exist -> out of range
+RRULE:FREQ=YEARLY;BYWEEKNO=10;BYDAY=2MO
+DTSTART:20240101T090000
+DURATION:PT1H
+EXCEPTION:Ical.Net.Evaluation.EvaluationOutOfRangeException
+EXCEPTION-STEP:Enumeration
+
+# Yearly BYDAY=15MO - 15th Monday of the year
+RRULE:FREQ=YEARLY;BYDAY=15MO;COUNT=3
```
minichma: Is this more or less a duplicated of line 204?

`RRULE:FREQ=YEARLY;BYDAY=20MO;COUNT=3`

minichma (commented):
I started looking at the test cases and from my point of view they look good. Commented some minor thoughts, just nits. Didn't look at the actual code so far. Will try to find time at the weekend.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -310,27 +310,29 @@ private static CalDateTime GetIntervalLowerLimit(CalDateTime intervalRefTime, Re
...
-        /// This makes a difference in case of BYWEEKNO, which might span months and years. After it was applied (BYWEEKNO would
-        /// always expand), the subsequent parts mustn't expand.
-        /// </remarks>
-        public bool DatesFullyExpanded { get; set; }
+        /// <example>
+        /// BYWEEKNO can expand the candidate set across month and year boundaries. When BYWEEKNO
+        /// performs expansion (i.e. the expand behavior is enabled for BYWEEKNO), the evaluator
+        /// marks the candidate set as expanded and subsequent BY‑parts must not expand again.
+        /// </example>
+        public bool IsCandidateSetFullyExpanded { get; set; }
```
minichma, line 321: :+1:

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -636,15 +638,25 @@ private static IEnumerable<CalDateTime> GetDayVariants(IEnumerable<CalDateTime>
...
+            // weekday dates according to the frequency and check containment.
+            dates.Where(date => pattern.ByDay.Any(weekDay =>
+            {
+                if (weekDay.Offset is null)
+                    return weekDay.DayOfWeek.Equals(date.DayOfWeek);
+
+                // When limiting with an offset (e.g. "22MO" or "1MO"), compute the
+                // absolute dates for that WeekDay in the appropriate scope and
+                // check if the candidate matches one of them.
+                return GetAbsWeekDays(date, weekDay, pattern).Any(d => d.Equals(date));
```
minichma, line 652: We have the same kind of code in the expand variant, both can be pretty expensive. Also the implementations of `GetAbsWeekDays*` could be significantly improved. Candidates for the future. :-)

minichma (dismissed):
Very nice! Sorry for the delay!

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
axunonb: Will add "RFC says that BYWEEKNO + numeric BYDAY is invalid. We're taking the lenient approach here."

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
axunonb: Right, will remove DURATION altogether.

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
axunonb: Removed

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: Found some easy to implement improvements resulting in 10-20% higher speed and less GC.

axunonb:
@minichma, @maknapp Thanks for your reviews, and happy holidays.

PR #889, opened 2025-12-02
Fix regression introduced in v5.1.3

### LINKED ISSUE #888: Use `RecurrenceTestCases.txt` as input for recurrence tests

### LINKED ISSUE #890: Incorrect Occurrence Calculation with Compound BYMONTHDAY Rules Containing Positive and Negative Offsets
Two test cases involving YEARLY recurrence rules with BYMONTHDAY containing both positive and negative offsets are failing. The library appears to incorrectly calculate occurrences in these specific scenarios, either omitting valid instances or potentially including invalid ones.

```ics
DTSTART: 2025-01-01
RRULE: FREQ=YEARLY;BYYEARDAY=-1,1;BYMONTHDAY=-1,1;COUNT=3
```
Expected instances: 2025-01-01, 2025-12-31, 2026-01-01
Filtering Logic Issue

```ics
DTSTART: 2025-06-01
RRULE: FREQ=YEARLY;BYMONTH=6;BYMONTHDAY=2,-2;UNTIL=20250627
```
Expected instances: 2025-06-02
Negative Offset Calculation Issue

Environment
[iCal.NET](https://ical.net/) Version: 5.1.3
.NET Version: NET8.0

axunonb:
Implemented review comments for PR #884 by @minichma
These `RRULE`s had no test cases and worked in v5.1.2, but failed in v5.1.3

```ics
DTSTART:20250101
RRULE:FREQ=YEARLY;BYYEARDAY=-1,1;BYMONTHDAY=-1,1;COUNT=3
INSTANCE:20250101,20251231,20260101

DTSTART:20250601
RRULE:FREQ=YEARLY;BYMONTH=6;BYMONTHDAY=2,-2;UNTIL=20250627
INSTANCE:20250602
```

Closes #888
Fixes #890


axunonb:
`GetAbsWeekDaysDaily(date, weekDay)` never gets executed with our unit tests:
https://github.com/ical-org/ical.net/blob/454247d10f4d28c3b1ebcbad8a22d95528746684/Ical.Net/Evaluation/RecurrencePatternEvaluator.cs#L669

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -4807,4 +4807,50 @@ public void GetOccurrences_WithMultipleOverridesForSameRecurrenceId_ShouldUseLat
...
+    }
+
+    [Test, Category("Recurrence")]
+    public void Yearly_ByMonth_MixedPositiveAndNegative_ByMonthDay_LowerLimitHeuristic()
+    {
+        const string ics = """
+                           BEGIN:VCALENDAR
+                           VERSION:2.0
+                           BEGIN:VEVENT
+                           DTSTART:20250601
```
minichma: Nice test! Maybe set DTSTART to `20250602`, so it is in line with the RRULE?

Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -43,7 +43,7 @@ protected void IncrementDate(ref CalDateTime dt, RecurrencePattern pattern, int
                 case FrequencyType.Yearly:
                     // When a rule uses BYWEEKNO, recurrence enumeration
                     // is based on week numbers relative to the year.
-                    // So we preserve January 1st when adding years; otherwise, preserve month/day.
+                    // So we preserve the weekday when using BYWEEKNO and preserve month/day otherwise.
```
minichma: oh, I'm sorry, I guess I was wrong with my comment in #513. We indeed preserve Jan 1st, not the weekday. So I guess the previous version was better.

Ical.Net/Evaluation/Evaluator.cs
```diff
@@ -43,7 +43,7 @@ protected void IncrementDate(ref CalDateTime dt, RecurrencePattern pattern, int
...
                     dt = (pattern.ByWeekNo.Count != 0)
                         ? old.AddDays(-old.DayOfYear + 1).AddYears(interval)
                         : old.AddYears(interval);
```
minichma: I don't fully understand this change (from #513) though, particularly the non-weekno case. It seems to be somewhat dangerous to just add an interval to the year, especially on Feb 29, as it might not exist in the next interval. Actually, when adding 1y to 2024-02-29, we end up in 2025-02-28 (just tried). However, when reaching 2028, we still stay on the 28th, even though the year has a 29th. The incorrect monthday doesn't seem to cause an issue, as it is properly handled in RecurrencePatternEvaluator, but as it doesn't seem to be considered anyways, I'd rather recommend setting it to a fixed value (Jan 1) rather than maintaining a 'wrong' month day. But then, do we need this distinction at all? We actually don't have a test that convers this distinction, i.e. when setting it always to Jan 1st, all tests are still green.

Moreover, `RecurrencePatternEvaluator` relies on the assumption that after incrementing, the new refDate is usually at the first day of an interval. For that reason, the following test case fails with the new behavior:

```ics
RRULE:FREQ=YEARLY;BYYEARDAY=1,10;UNTIL=20260105
DTSTART:20250110
INSTANCES:20250110,20260101
```

see also my comments in GetIntervalLowerLimit()

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -263,8 +255,12 @@ private static CalDateTime GetIntervalLowerLimit(CalDateTime intervalRefTime, Re
                 int day;
                 if (pattern.ByMonthDay.Count > 0)
                 {
-                    var md = pattern.ByMonthDay.Min();
-                    day = md > 0 ? Math.Min(md, daysInMonth) : Math.Max(1, daysInMonth + md + 1);
+                    // Map BYMONTHDAY entries (positive and negative) to absolute days
+                    // in the target month, then pick the smallest. This handles cases
+                    // where BYMONTHDAY mixes positive and negative values (e.g. 2,-2).
+                    var mappedDays = pattern.ByMonthDay
```
minichma, line 277: Getting back to https://github.com/ical-org/ical.net/pull/884/files/84eb777d104f900184e58cb7cbc11c6126e17816#r2575960238

I was digging into the code (and git history) to remember, why I introduced this method. The code historically relies on the assumption that after `Evaluator.IncrementDate()` the new refDate is usually set to the first day of the given interval. For cases where this assumption doesn't hold, this method is intended to return the interval's actual start date (lower limit). So far this was only required for yearly recurrences when using BYWEEKNO, because for yearly recurrences the date was always set to Jan 1 but Week 1 could start in the previous year. In all other cases the code worked without that adjustment, but that's not quite obvious too. At least the time of the day is never set to 0:00, which is compensated by the fact, that we add 1d in `GetSearchEndDate()` but this is not documented at all and by no means obvious.

For now I'd recommend to set the date to Jan 1st in `Evaluator.IncrementDate()` in case of yearly recurrences. In this method I recommend to only deal with the special case of yearly/byweekno or alternatively return Jan 1st for the yearly case if byweekno is not used (but not only in BYMONTH is set).

At a later time we might investigate how to improve GetIntervalLowerLimit and GetSearchEndDate. The 1d adjustment in GetSearchEndDate is only intended to compensate for potential DST changes, not for refDate not being set to the beginning of the interval. I feel we should improve this so it is always set to the beginning of the interval, including the time part, and ideally also for the yearly/byweekno case. If we manage to do so, we can get rid of GetIntervalLowerLimit altogether.

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -4807,4 +4807,50 @@ public void GetOccurrences_WithMultipleOverridesForSameRecurrenceId_ShouldUseLat
...
     }
+
+    [Test, Category("Recurrence")]
+    public void FreqYearly_With_ByYearDay_ByMonthDay()
+    {
+        const string ics = """
+                           BEGIN:VCALENDAR
+                           VERSION:2.0
+                           BEGIN:VEVENT
+                           RRULE:FREQ=YEARLY;BYYEARDAY=-1,1;BYMONTHDAY=-1,1;COUNT=3
```
minichma: These tests would be great candidates for being included in RecurrenceTestCases.txt, which would safe a lot of boiler-plate.

Ical.Net/Evaluation/Evaluator.cs
axunonb: > recommend setting it to a fixed value (Jan 1)

Fixed with next commit

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: > only deal with the special case of yearly

Fixed with next commit

Ical.Net.Tests/RecurrenceTests.cs
axunonb: Fixed with next commit

axunonb:
@minichma Thanks for the review and taking the time for a deep dive. 
See commit comments for details about how your remarks are now implemented..
Hope this version is more bullet proof.

minichma (approved):

axunonb:
@minichma Thanks!

minichma:
Urw 🙂

PR #901, opened 2025-12-31
New recurrence evaluator

maknapp:
This evaluator follows the same basic logic as the current one, but with a few key differences:

- Moves the expand/limit grid into the flow control of the evaluator. No need for `IsCandidateSetFullyExpanded`.
- BY rule normalization is moved to a separate class and cached when possible. Caching may not be very useful in real world usage though.
- BYDAY values without offsets are generated. BYDAY with offsets still need to calculate entire range then sort, which is likely a small number of values anyway.
- Enumeration is only restarted when BYSETPOS is set.
- Reduced usage of Linq - reduces allocation.
- BYSETPOS with only positive values stops evaluating the set early. Added benchmark to show this.
- RecurrencePatternEvaluator2 is used inside RecurrencePatternEvaluator just to easily comment out and compare between the two evaluators. It is intended to replace.
- Some new tests fail with current evaluator.
- Starts evaluation at `periodStart` by skipping directly to it instead of looping. There is a new test in RecurrenceTestCases.txt commented out because it takes forever on the current evaluator.
- `BugByWeekNoNotWorking()` test was changed to YEARLY because the WEEKLY+BYWEEKNO is invalid.
- The seed value is incremented nearest to reference date always, instead of the start of the week/month/year that `IncrementDate()` does. This makes the seed value require minimal changes and no extra BY rules.
- `ProcessRecurrencePattern()` is NOT applied. No extra BY rules are added to evaluation.

## Before (be1d758c):
| Method                                                     | Mean        | Error    | StdDev   | Gen0     | Gen1    | Allocated  |
|----------------------------------------------------------- |------------:|---------:|---------:|---------:|--------:|-----------:|
| GetOccurrences                                             | 1,352.93 us | 9.169 us | 8.577 us | 224.6094 | 42.9688 |  1837.3 KB |
| MultipleEventsWithUntilOccurrencesSearchingByWholeCalendar |    15.63 us | 0.175 us | 0.163 us |   3.9673 |  0.1526 |   32.63 KB |
| MultipleEventsWithUntilOccurrences                         |    74.32 us | 0.589 us | 0.551 us |  13.5498 |  0.3662 |  111.19 KB |
| MultipleEventsWithUntilOccurrencesEventsAsParallel         |    45.78 us | 0.641 us | 0.535 us |  16.1133 |  1.2207 |  130.64 KB |
| EventWithBySetPosRecurrence                                | 1,149.22 us | 9.324 us | 8.722 us | 175.7813 | 11.7188 | 1448.73 KB |
| EventWithPositiveBySetPosRecurrence                        | 1,118.22 us | 4.922 us | 4.604 us | 146.4844 |  7.8125 | 1209.14 KB |
| MultipleEventsWithCountOccurrencesSearchingByWholeCalendar |    32.70 us | 0.134 us | 0.119 us |   7.9956 |  0.7324 |    65.8 KB |
| MultipleEventsWithCountOccurrences                         |   517.66 us | 3.159 us | 2.638 us |  97.6563 | 15.6250 |  804.97 KB |
| MultipleEventsWithCountOccurrencesEventsAsParallel         |   257.24 us | 4.402 us | 3.902 us | 102.5391 | 21.4844 |  831.75 KB |

## After:
| Method                                                     | Mean      | Error     | StdDev    | Gen0     | Gen1    | Allocated |
|----------------------------------------------------------- |----------:|----------:|----------:|---------:|--------:|----------:|
| GetOccurrences                                             | 927.16 us | 17.564 us | 18.793 us | 113.2813 | 22.4609 | 928.75 KB |
| MultipleEventsWithUntilOccurrencesSearchingByWholeCalendar |  11.03 us |  0.039 us |  0.034 us |   2.8687 |  0.0763 |  23.53 KB |
| MultipleEventsWithUntilOccurrences                         |  45.64 us |  0.332 us |  0.294 us |   7.4463 |  0.1221 |  61.03 KB |
| MultipleEventsWithUntilOccurrencesEventsAsParallel         |  31.93 us |  0.462 us |  0.410 us |   9.9487 |  0.8545 |  80.48 KB |
| EventWithBySetPosRecurrence                                | 327.28 us |  1.035 us |  0.968 us |  65.9180 |  3.4180 | 540.45 KB |
| EventWithPositiveBySetPosRecurrence                        | 108.77 us |  2.161 us |  2.312 us |  13.5498 |  0.4883 | 111.47 KB |
| MultipleEventsWithCountOccurrencesSearchingByWholeCalendar |  27.50 us |  0.201 us |  0.178 us |   7.0190 |  0.6714 |  57.36 KB |
| MultipleEventsWithCountOccurrences                         | 343.27 us |  1.078 us |  0.955 us |  52.7344 |  8.3008 | 433.59 KB |
| MultipleEventsWithCountOccurrencesEventsAsParallel         | 186.75 us |  3.360 us |  2.979 us |  56.6406 | 11.4746 | 460.37 KB |


axunonb:
Thanks for this PR, @maknapp! This looks like a very promising step towards version 6.0 in terms of the recurrence evaluation.
As the PR is still marked as a draft, I have just a few high-level comments for now.

Currently `RecurrencePatternEvaluator2` and `RecurrencePatternEvaluator` are part of the PR, but I understand `RecurrencePatternEvaluator1` is considered as the replacement. For the review it would be helpful to use the final implementation.

Variable Shadowing: I noticed several places where local variables or parameters hide class-level fields with the same name. This is error-prone. Can we rename the local vars or prefix class members with _ to ensure we don't accidentally use the wrong one (like `editorconfig` suggests)?

`System.Data.EvaluateException`: Should this read as `Ical.Net.Evaluation.EvaluationException`?

Redundant Logic: There are some if...else if blocks that seem to re-check conditions or could be simplified. flattening these with guard clauses or using switch expressions would make the evaluation logic much easier to follow.

To make the review easier and ensure the long-term health of the repo, would you mind resolving the Blocker/High issues identified in the Sonar report? Regarding Coverage the 4.7% decrease in overall coverage may be caused by `RecurrencePatternEvaluator2` and `RecurrencePatternEvaluator` being part of the code base.

Thanks for the massive effort on the new evaluator.

Ical.Net/Evaluation/ByRuleValues.cs
```diff
@@ -0,0 +1,352 @@
...
+    private DayOfWeekComparer? dayOfWeekComparer;
+
+    private readonly int[] normalMonths;
+
+    private readonly int[] normalHours;
+    private readonly int[] normalMinutes;
+    private readonly int[] normalSeconds;
+
+    private readonly int[] byWeekNo;
+    private readonly NormalValues<int, int> weeks;
```
minichma: add comment: what is the cache key? In this case it is the number of weeks per year, right?

Ical.Net/Evaluation/ByRuleValues.cs
```diff
@@ -0,0 +1,352 @@
...
+
+    private readonly int[] byYearDay;
+    private readonly NormalValues<int, int> yearDays;
```
minichma: In this case the year is used as cache key. Followed the pattern of the `weeks` property, we could also use the number of days in the year as cache key. This way the number of variants would be limited (i.e. 365 or 366) and we could cache all of them. We could simply maintain something like `Dictionary<int, NormalValues<int>> weeksByYearDays` where the number of days per year would be the key. The `NormalValues` type itself wouldn't need to hold the cache key (which would be nice as caching is quite a different concern than holding the values). This way the number of cache misses could be improved even further. Same applies to other `NormalValues` properties.

maknapp:
> Currently `RecurrencePatternEvaluator2` and `RecurrencePatternEvaluator` are part of the PR...

Replaced.

> Variable Shadowing: I noticed several places where local variables or parameters hide class-level fields with the same name. This is error-prone. Can we rename the local vars or prefix class members with _ to ensure we don't accidentally use the wrong one (like `editorconfig` suggests)?

Done.

> `System.Data.EvaluateException`: Should this read as `Ical.Net.Evaluation.EvaluationException`?

Yes :D

> Redundant Logic: There are some if...else if blocks that seem to re-check conditions or could be simplified. flattening these with guard clauses or using switch expressions would make the evaluation logic much easier to follow.

Redundant within the same method? I just split some of the more complex methods, which may have been the areas you were talking about.

> resolving the Blocker/High issues identified in the Sonar report?

I fixed some of them. The "break out of recursion" issues do not make sense - they generators intended to run forever. The remaining `ExpandYearDay()` complexity issue would not benefit from changing in my opinion. The nested ifs are separated for clarity and I do not see a better way to do it.

For the minor Linq issues, some of those cannot use Linq without affecting allocations. The simple ones can use Linq, but it is easier to debug without it.

minichma (commented):
Wow, that's great work! Couldn't go through much of the code yet, so here are just some very first thoughts.

Some time ago I have also been working on some of the aspects you are addressing in this PR (e.g. caching). Unfortunately I didn't manage to complete it but just FYI, this is the branch, should you be interested. Just keep in mind that its work in progress:

 https://github.com/minichma/ical.net/tree/work/minichma/feature/modernize_rrule2

axunonb:
> The "break out of recursion" issues do not make sense

For the code logic this is the case. The analyzer just  insists on every code path should `yield` - similar to all code paths must return a value. A `yield break` makes the analyzer happy, and maybe it's also faster to understand. It has no other effect.

axunonb:
> For the minor Linq issues, some of those cannot use Linq without affecting allocations.

Yes, absolutely, and therefore we should avoid Linq here for now. (Microsoft says this will be improved with net10.0.)
The `foreach` loops have another effect: The exit point from the methods do not get hit with code coverage, because it only gets executed when the iterator finishes or is disposed explicitly. Long story short: We can live with that.

axunonb:
#### Increase code coverage for  `RecurrencePatternEvaluator`
`WeekdayValue` with zero `Offset`: Although sth. like `BYDAY=0SU` is invalid in RFC5545, the lenient behavior to ignore is maybe better to understand and also better than to throw.
These cases should also get code coverage, same as the additional `EvaluationException`s thrown. You may want to the include the attached tests cases in the PR. `RecurrencePatternEvaluator` code coverage should be as high as possible.
**[Additional_Unit_Tests.txt](https://github.com/user-attachments/files/24424031/Additional_Unit_Tests.txt)**

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+    /// <returns></returns>
+    private static IEnumerable<LocalDate> ExpandDayOfWeekWithoutOffset(
+        LocalDate start,
+        LocalDate end,
+        IsoDayOfWeek[] weekDays)
     {
-        // Helper that checks whether the given candidate matches any BYMONTHDAY entry
-        // taking negative values into account (relative to the month's length).
-        static bool MatchesAnyMonthDay(ZonedDateTime candidate, IEnumerable<int> monthDays)
+        if (weekDays.Length == 0)
```
axunonb: Can this ever be `true` at this stage? From my understanding the checks happen long before. So maybe remove it here?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+    private readonly Instant? _until;
+    private readonly FrequencyType _frequency;
+    private readonly IsoDayOfWeek _firstDayOfWeek;
+
+    private readonly IWeekYearRule _weekYearRule;
+    private readonly ZonedDateTime _zonedReferenceDate;
+    private readonly int _referenceWeekNo;
+
+    private ZonedDateTime _seed;
+    private int _unmatchedIncrementCount = 0;
```
axunonb: Redundant initialization with default value?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+                return start.Date.PlusYears(nominalByInterval)
+                    .AtNearestDayOfMonth(_zonedReferenceDate.Day)
+                    .At(start.TimeOfDay)
+                    .InZoneLeniently(start.Zone);
         }
 
-        // optimize the start time for selecting candidates
-        // (only applicable where a COUNT is not specified)
-        if (pattern.Count is null && periodStartDt is not null)
+        throw new EvaluationException("Invalid frequency type");
```
axunonb: Add the invalid `_frequency` to the message?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
         {
-            var incremented = seedCopy;
-            while (incremented.ToInstant() < periodStartDt.Value.ToInstant())
-            {
-                seedCopy = incremented;
-                IncrementDate(ref incremented, pattern, pattern.Interval);
-            }
+            return StartByRules();
+        }
+        else if (_rule.HasNegativeSetPos)
```
axunonb: Remove (all) the redundant `else` then the `if` returns a value?
IMHO this would be easier to read.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
-        if (expand == null || pattern.ByMonth.Count == 0)
-            return dates;
-
-        if (expand.Value)
+        if (_rule.HasByDayOffsets)
         {
-            // Expand behavior
-            return dates.SelectMany(d => pattern.ByMonth
-                .Select(month => d.LocalDateTime.PlusMonths(month - d.Month).InZoneLeniently(d.Zone)));
+            throw new EvaluationException($"BYDAY offsets are not supported in {_frequency}");
```
axunonb: Refer to invalidity in RFC5545 instead of "support".

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
-            dates.Where(date => pattern.ByDay.Any(weekDay =>
+    private IEnumerable<ZonedDateTime> ByMonthDay()
+    {
+        if (!_rule.ByMonthDay)
+        {
+            return ByYearDay();
+        }
+        else if (_frequency == FrequencyType.Weekly)
+        {
+            throw new EvaluationException("BYMONTHDAY is not supported in WEEKLY");
```
axunonb: Refer to invalidity in RFC5545 instead of "support"?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+        }
+        else if (_frequency < FrequencyType.Daily)
         {
-            var matchesWeekNo = byWeekNoNormalized == null || byWeekNoNormalized.Contains(weekYearRule.GetWeekOfWeekYear(date.Date));
-            var matchesMonth = pattern.ByMonth.Count == 0 || pattern.ByMonth.Contains(date.Month);
+            return LimitYearDay();
+        }
+        else
+        {
+            throw new EvaluationException($"BYYEARDAY is not supported with {_frequency}");
```
axunonb: Refer to invalidity in RFC5545 instead of "support"?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+            {
+                var result = startOfYear.PlusDays(day - 1);
 
-        // Find first occurrence of target weekday
-        var targetDayOfWeek = weekDay.DayOfWeek.ToIsoDayOfWeek();
-        if (date.DayOfWeek != targetDayOfWeek)
-        {
-            date = date.Next(targetDayOfWeek);
-        }
+                // Ignore values outside of the calendar year
```
axunonb: "of" is redundant?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+        {
+            return ByMonth();
+        }
+        else if (_frequency == FrequencyType.Yearly)
+        {
+            return ExpandWeekNo();
+        }
+        else
+        {
+            throw new EvaluationException($"BYWEEKNO is not supported with {_frequency}");
```
axunonb: Refer to invalidity in RFC5545 instead of "support"?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
 
     /// <summary>
-    /// Evaluate the occurrences of this recurrence pattern.
+    /// Gets the next seed value based on the recurrence frequency
+    /// and interval.
     /// </summary>
-    /// <param name="referenceDate">The reference date, i.e. DTSTART.</param>
-    /// <param name="periodStart">Start (incl.) of the period occurrences are generated for.</param>
-    /// <param name="options"></param>
+    /// <param name="seed"></param>
```
axunonb: No "seed" argument

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+                .At(_seed.TimeOfDay)
+                .InZoneLeniently(_seed.Zone),
+
+            FrequencyType.Yearly => _seed.Date
+                .PlusYears(_interval)
+                .AtNearestDayOfMonth(_zonedReferenceDate.Day)
+                .At(_seed.TimeOfDay)
+                .InZoneLeniently(_seed.Zone),
+
+            _ => throw new EvaluationException("Invalid frequency")
```
axunonb: Add the invalid `_frequency` value to the message?

Ical.Net.Tests/RecurrenceTests.cs
```diff
@@ -4569,4 +4562,267 @@ public void GetOccurrences_BackwardsDaylightSaving_SpanFirstHour(string start, s
 
         Assert.That(result, Is.EqualTo(expected));
     }
+
+    [Test]
```
axunonb, line 4570: Optional: Add the tests to `RecurrenceTestCases.txt`
(Avoids a lot of boilerplate code)

axunonb (changes requested):
This is great work and excellent code, thanks a lot! Good to have you working on ical.net.
See comments in files and conversation.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -5,976 +5,1184 @@
...
+    /// <returns></returns>
+    private IEnumerable<ZonedDateTime> ByDay()
     {
-        var expandContext = new ExpandContext { IsCandidateSetFullyExpanded = false };
+        if (!_rule.ByDay)
+        {
+            if (_frequency is FrequencyType.Monthly or FrequencyType.Yearly
+                && !_rule.ByMonthDay
+                && !_rule.ByYearDay
+                && !_rule.ByWeekNo)
```
axunonb: Merge into pattern?
```cs
if (_frequency is FrequencyType.Monthly or FrequencyType.Yearly
    && _rule is { ByMonthDay: false, ByYearDay: false, ByWeekNo: false })
```

Ical.Net.Tests/RecurrenceTests.cs
maknapp: The tests in `RecurrenceTestCases.txt` are all UTC events evaluated to "US-Eastern" time zone. These tests require the `CalendarEvent` to be in a specific time zone, so they cannot be added to the txt file.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
maknapp: This should never be true right now, but it is a helpful check. Future code changes can mistakenly make it reachable, and rest of the method does not support an empty array.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
maknapp: I was confused by this and had to look it up. This appears to be a US vs British English difference. :D

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
maknapp: I usually avoid that type of pattern matching in if statements as it does not look as clear to me.

maknapp:
>A `yield break` makes the analyzer happy

There is no place to `yield break` though.

For the various `EvaluationException`s, would it make sense to validate rrules before evaluation? Evaluating a `Calendar` would throw if just one event was invalid. If it were validated on creation, it could be ignored before being added to the calendar? Can this be done without breaking serialize/deserialize backwards compatibility?

Ical.Net/Evaluation/ByRuleValues.cs
maknapp: I changed it to start caching all values with `Dictionary`. Benchmarks are generally faster with similar allocation. Is this okay to do with BYSETPOS though? How many different set sizes can a single rrule produce?

axunonb:
> validate rrules before evaluation

A topic for another isssue/pr?

axunonb:
> There is no place to `yield break` though.

Marked as accepted in SonarCloud

Ical.Net/DataTypes/RecurrencePattern.cs
```diff
@@ -175,5 +175,18 @@ public override void CopyFrom(ICopyable obj)
...
+            || ByMonthDay.Count > 0
+            || ByWeekNo.Count > 0
+            || ByYearDay.Count > 0
+            || ByHour.Count > 0
+            || ByMinute.Count > 0
+            || BySecond.Count > 0
+            || BySetPosition.Count > 0;
+    }
+
     private static bool CollectionEquals<T>(IEnumerable<T> c1, IEnumerable<T> c2) => c1.SequenceEqual(c2);
```
axunonb: Not used any more - remove?

Ical.Net.Tests/Calendars/Recurrence/RecurrenceTestCases.txt
```diff
@@ -256,3 +261,159 @@ DTSTART:20260601
 INSTANCES:20260601,20260908,20261208
 
 ############################## END ERRATA 1913 TESTS ##############################
+
```
axunonb, line 264: 👍

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: Then `System.Diagnostics.Debug.Assert(weekDays.Length != 0, "Weekdays must not be empty.");` ?

Ical.Net/Evaluation/ByRuleValues.cs
minichma: > Is this okay to do with BYSETPOS though?

Good question. I guess it should be ok for BYSETPOS as well. The date part of an occurrence can have no more than 366 values per interval (days in a yearly recurrence), so 366 is a hard upper bound for the number of different set sizes (the time part doesn't affect this number, because it won't vary across intervals). The real max number of different set sizes will certainly be significantly lower, not sure what it is, but even 366 won't cause any issues. So I guess we're safe in this respect.

axunonb:
Now that `RecurrencePatternEvaluator` does not inherit from `Evaluator`, question is whether we still need `Evaluator` and `IEvaluator` at all.
The abstract `Evaluator` and `IncrementDate(ref ZonedDateTime, RecurrencePattern, int)` are the only places that reference `IEvaluator`.
`RecurringEvaluator` currently inheriting from `Evaluator` could implement `IEvaluator` directly.
This change could go into this PR or a new one. What do you think?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
maknapp: Are all of "not supported" -> "invalid" okay now?

Ical.Net/Evaluation/TodoEvaluator.cs
```diff
@@ -147,6 +147,35 @@ protected override ZonedDateTime GetEnd(ZonedDateTime start)
         return start;
     }
 
+    protected override ZonedDateTime GetStart(ZonedDateTime end)
```
axunonb: Not sure whether calculation of `DTSTART` is a good idea because it is against RFC5545 definitions. The calculation of a `DTSTART` (Start date/time) using a `DTEND` (or `DUE`) and a `DURATION` is disallowed in RFC 5545 because the specification establishes `DTSTART` as the primary anchor for all temporal and recurrence logic.

Mutual exclusivity:
The specification enforces strict mutual exclusivity between the end of an event and its duration. Within a `VEVENT`, the properties `DTEND` and `DURATION` MUST NOT occur in the same component. Similarly, within a `VTODO`, the properties `DUE` and `DURATION` MUST NOT occur together.

 Explicit Dependency for `DURATION`:
For `VTODO `components, the RFC explicitly states that if a `DURATION` is provided, then `DTSTART` MUST also appear in the same component.

Recurrence Calculation Anchor:
`DTSTART` is REQUIRED in all types of recurring calendar components.

Ical.Net/Evaluation/TodoEvaluator.cs
```diff
@@ -147,6 +147,35 @@ protected override ZonedDateTime GetEnd(ZonedDateTime start)
...
+    {
+        var dtStart = Todo.DtStart;
+
+        if (Todo.Duration is { } duration)
+        {
+            if (dtStart is null)
+            {
+                throw new EvaluationException("DtStart must be set when a TODO has a duration.");
```
axunonb: DtStart must be ~~set~~ **null**

Ical.Net/Evaluation/TodoEvaluator.cs
maknapp: Sorry, I did not intend for this last commit ([c494756](https://github.com/ical-org/ical.net/pull/901/commits/c494756e0f4132f2bb89c7fb05ec775e7429fb00)) to be here. I will be moving this to another PR.

No need to discuss in this PR, but `GetStart(...)` here is only for getting a "corrected" `periodStart` value that includes current occurrences. `periodStart` is an evaluation option for efficiency, not part of the RFC, so adjusting like this should be okay.

maknapp:
I added more comments to `ByRuleValues`, simplified the caching part, and removed the above mentioned commit.

> Now that `RecurrencePatternEvaluator` does not inherit from `Evaluator`, question is whether we still need `Evaluator` and `IEvaluator` at all. The abstract `Evaluator` and `IncrementDate(ref ZonedDateTime, RecurrencePattern, int)` are the only places that reference `IEvaluator`. `RecurringEvaluator` currently inheriting from `Evaluator` could implement `IEvaluator` directly. This change could go into this PR or a new one. What do you think?

Probably a separate PR. `TodoEvaluator` has the last usage of `IncrementDate` and I'm not sure exactly what is happening there. All tests still pass with that removed, so there are no tests specifically for that.

axunonb (approved):
Great job, thanks!

PR #905, opened 2026-01-15
breaking: Update target frameworks

### LINKED ISSUE #902: Update target frameworks
Define `net10.0`, `net8.0` and `netstandard2.0` as target frameworks for version 6.
~~Maybe also `net10.0-android` and `net10.0-ios`?~~
@minichma @maknapp Opinions?

axunonb:
Supported target frameworks

`Ical.Net`: net10.0  net8.0  netstandard2.0
`Ical.Net.Tests`: net10.0  net8.0  net48
`Ical.Net.Benchmarks`: net10.0

(Drop support for net6.0, netstandard2.1)

Resolves #902

### DIFF
--- a/Ical.Net/Utility/TextUtil.cs
+++ b/Ical.Net/Utility/TextUtil.cs
@@ -57,11 +57,8 @@ internal static class TextUtil
             byteIndex += chBytes;
         }
 
-        // Append the remaining characters to the result
-        if (charIndex > 0)
-        {
-            result.Append(lineArray, 0, charIndex);
-        }
+        // Append the remaining characters (if any, i.e. charIndex > 0) to the result
+        result.Append(lineArray, 0, charIndex);
 
         result.Append(SerializationConstants.LineBreak);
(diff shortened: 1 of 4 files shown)


Ical.Net/Utility/TextUtil.cs
```diff
@@ -57,11 +57,8 @@ public static void FoldLines(this StringBuilder result, string inputLine)
             byteIndex += chBytes;
         }
 
-        // Append the remaining characters to the result
-        if (charIndex > 0)
-        {
-            result.Append(lineArray, 0, charIndex);
-        }
+        // Append the remaining characters (if any, i.e. charIndex > 0) to the result
+        result.Append(lineArray, 0, charIndex);
```
axunonb, line 61: Test for `charIndex > 0` is redundant, because `result.Append(lineArray, 0, 0);` is a no-op anyway.

maknapp (approved):

PR #907, opened 2026-01-16
Enable nullable reference types in `Ical.Net.Tests.csproj` 

### LINKED ISSUE #904: chore: Update unit tests with project-wide `nullable enable`

axunonb:
No changes in code logic, only affecting unit tests.

* Add `nullable enable` to `Ical.Net.Tests`
* Fix nullability warnings by applying nullable and null-forgiving operators in unit tests
* Replace `Assert.Multiple` with `using (Assert.EnterMultipleScope())`

Resolves #904

### DIFF
--- a/Ical.Net.Tests/DataTypeTest.cs
+++ b/Ical.Net.Tests/DataTypeTest.cs
@@ -14,15 +14,13 @@ public class DataTypeTest
 {
     [Test, Category("DataType")]
     public void OrganizerConstructorMustAcceptNull()
-    {
-        Assert.DoesNotThrow(() => { var o = new Organizer(null); });
-    }
+        => Assert.DoesNotThrow(() => { var o = new Organizer(null!); });
 
     [Test, Category("DataType")]
     public void AttachmentConstructorMustAcceptNull()
     {
-        Assert.DoesNotThrow(() => { var o = new Attachment((byte[]) null); });
-        Assert.DoesNotThrow(() => { var o = new Attachment((string) null); });
+        Assert.DoesNotThrow(() => { var o = new Attachment((byte[]?) null); });
+        Assert.DoesNotThrow(() => { var o = new Attachment((string) null!); });
     }
 
     public static IEnumerable<TestCaseData> TestWeekDayEqualsTestCases => [
--- a/Ical.Net.Tests/DataTypeTest.cs
+++ b/Ical.Net.Tests/DataTypeTest.cs
@@ -35,22 +35,22 @@ public class DataTypeTest
     [Test, TestCaseSource(nameof(TestWeekDayEqualsTestCases)), Category("DataType")]
     public void TestWeekDayEquals(WeekDay w1, WeekDay w2)
     {
-        Assert.Multiple(() =>
+        using (Assert.EnterMultipleScope())
         {
             Assert.That(w1.Equals(w1), Is.True);
             Assert.That(w1.Equals(w2), Is.False);
-        });
+        }
     }
 
     [Test, TestCaseSource(nameof(TestWeekDayEqualsTestCases)), Category("DataType")]
     public void TestWeekDayCompareTo(WeekDay w1, WeekDay w2)
     {
-        Assert.Multiple(() =>
+        using (Assert.EnterMultipleScope())
         {
             Assert.That(w1.CompareTo(w1), Is.EqualTo(0));
 
             if (w2 != null)
                 Assert.That(w1.CompareTo(w2), Is.Not.EqualTo(0));
-        });
+        }
     }
 }
--- a/Ical.Net.Tests/DataTypeTest.cs
+++ b/Ical.Net.Tests/DataTypeTest.cs
@@ -14,13 +14,13 @@ public class DataTypeTest
 {
     [Test, Category("DataType")]
     public void OrganizerConstructorMustAcceptNull()
-        => Assert.DoesNotThrow(() => { var o = new Organizer(null!); });
+        => Assert.DoesNotThrow(() => { _ = new Organizer(null); });
 
     [Test, Category("DataType")]
     public void AttachmentConstructorMustAcceptNull()
     {
-        Assert.DoesNotThrow(() => { var o = new Attachment((byte[]?) null); });
-        Assert.DoesNotThrow(() => { var o = new Attachment((string) null!); });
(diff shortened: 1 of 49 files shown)


Ical.Net.Tests/Logging/Abstractions/InMemorySink.cs
```diff
@@ -14,13 +13,14 @@ namespace Ical.Net.Tests.Logging.Abstractions;
 
 internal class InMemorySink : ILogEventSink
 {
-    public static InMemorySink Instance;
+    public static InMemorySink Instance = null!;
 
-    private ConcurrentQueue<Serilog.Events.LogEvent> _events { get; } = new();
+    private ConcurrentQueue<Serilog.Events.LogEvent> Events { get; } = new();
```
maknapp: This should be a readonly field not a property, and probably named back to `_events`.
```c#
private readonly ConcurrentQueue<Serilog.Events.LogEvent> _events = new();
```

Ical.Net.Tests/WikiSamples/RecurrenceWikiTests.cs
```diff
@@ -129,12 +127,12 @@ public void DailyIntervalCount()
...
         var calendarSerializer = new CalendarSerializer();
-        var generatedIcs = calendarSerializer.SerializeToString(calendar);
+        var generatedIcs = calendarSerializer.SerializeToString(calendar)!;
 
         // Calculate all occurrences
         var tz = TimeZoneResolvers.Default("Europe/Zurich");
-        IEnumerable<Occurrence> occurrences = calendar.GetOccurrences(tz);
-        Assert.That(occurrences.Count(), Is.EqualTo(2));
+        IEnumerable<Occurrence> occurrences = calendar.GetOccurrences(tz).ToList();
+        Assert.That(occurrences.Count, Is.EqualTo(2));
```
maknapp: This is still using `IEnumerable.Count()` instead of `List<T>.Count`. Change to `var occurrences = ...`, and `Has.Count.EqualTo`.

Ical.Net.Tests/DataTypeTest.cs
```diff
@@ -14,15 +14,13 @@ public class DataTypeTest
 {
     [Test, Category("DataType")]
     public void OrganizerConstructorMustAcceptNull()
-    {
-        Assert.DoesNotThrow(() => { var o = new Organizer(null); });
-    }
+        => Assert.DoesNotThrow(() => { var o = new Organizer(null!); });
```
maknapp: Why does `Organizer` not accept nullable `string?` if it allows for null values? Same for `Attachment(string)` below.

Ical.Net.Tests/SerializationTests.cs
```diff
@@ -46,7 +46,7 @@ internal static void CompareComponents(ICalendarComponent cb1, ICalendarComponen
                 }
                 else if (p1.Value is IEnumerable)
                 {
-                    CompareEnumerables((IEnumerable)p1.Value, (IEnumerable)p2.Value, p1.Name);
+                    CompareEnumerables((IEnumerable)p1.Value, (IEnumerable) p2.Value!, p1.Name);
```
maknapp: In this PR or another, `CompareEnumerables()` could be replaced with `Assert.That(..., Is.EquivalentTo(...))`.

Ical.Net.Tests/SymmetricSerializationTests.cs
```diff
@@ -208,23 +205,22 @@ public void CategoriesTest(string[] categories)
 
     private static IEnumerable CategoryTest_TestCases()
     {
-        yield return new string[] { "Foo", "Bar", "Baz" };
-        yield return new string[] { "Hello", "world" };
-        yield return new string[] { "Hello", "world", null };
+        yield return new[] { "Foo", "Bar", "Baz" };
+        yield return new[] { "Hello", "world" };
```
maknapp, line 209: No need to test `null` anymore?

maknapp (changes requested):

Ical.Net.Tests/WikiSamples/RecurrenceWikiTests.cs
axunonb: oops, tnx - of course

Ical.Net.Tests/Logging/Abstractions/InMemorySink.cs
axunonb: 👍

Ical.Net.Tests/DataTypeTest.cs
axunonb: `Organizer` and `Attachment` constructors did not accept `null`, although there was a check for a `null` argumenta. Existing code might pass `null` though.
I changed the constructors' signature for now.

Ical.Net.Tests/SymmetricSerializationTests.cs
axunonb: No, indeed not, because `CalendarEvent.Categories` is an `IList<string>`

Ical.Net.Tests/SerializationTests.cs
axunonb: Indeed, this was the only reference, so no problem.

axunonb:
Thanks for review!

Ical.Net.Tests/SerializationTests.cs
```diff
@@ -44,9 +44,9 @@ internal static void CompareComponents(ICalendarComponent cb1, ICalendarComponen
                     if (((IComparable)p1.Value).CompareTo(p2.Value) != 0)
                         continue;
                 }
-                else if (p1.Value is IEnumerable)
+                else if (p1.Value is IEnumerable p1Enumerable)
                 {
-                    CompareEnumerables((IEnumerable)p1.Value, (IEnumerable) p2.Value!, p1.Name);
+                    Assert.That(p1Enumerable, Is.EquivalentTo((IEnumerable) p2.Value!));
```
maknapp: Add `Assert.That(..., message: p1.Name + " do not match");` to match the error message from before.

maknapp (changes requested):

Ical.Net.Tests/SerializationTests.cs
axunonb: Done

maknapp (approved):

Issue #908, opened 2026-01-18
Replace Regex-based parsing with high-performance `Span`-based parsers


axunonb:
#### **Summary**

The current deserialization relies on a complex Regular Expressions. While functional, this approach is suboptimal for performance and GC. We should replace the Regex logic with manual `ReadOnlySpan<char>` parsers to eliminate heap allocations and mitigate risks of catastrophic backtracking.

#### **The Problem**

1. **Allocations:** Every call to `RegEx.Match(value)` creates multiple `Group` and `Capture` objects on the heap.
2. **Backtracking Risk:** The nested optional groups can lead to high CPU usage on malformed inputs.
3. **Cognitive Complexity:** The Regex is difficult to debug or extend without specialized tools.

#### **The Proposed Solution**

Implement a non-allocating parser using `ReadOnlySpan<char>`. The new implementation:

* Uses **zero heap allocations** during the parsing phase.
* Maintains parity with RFC 5545 logic.
* Reduces cognitive complexity by breaking the state machine into clear, named helper methods.

#### **Impact / Benchmarks**

Based on initial testing of the proposed refactor:

* **Execution Time:** ~10-15x faster than `Regex.Match`.
* **Memory Usage:** Reduced from ~400+ bytes per call to **0 bytes** (excluding the input string itself).
* **Safety:** Elimination of the `RegexTimeout` requirement as the new parser is strictly linear.

#### **Affected Components**

* All serializers making use of RegEx

maknapp:
Source generated regex generates Span-based match functions, so that should be considered - especially for more complicated regex patterns. Also related PR #568 could be updated or closed after this.

Are you planning on any changes for netstandard2.0 that does not support Span or source generated regex?

axunonb:
Regex-based parsing in this context is difficult to debug, extend, and prone to performance issues like catastrophic backtracking—especially given the nested groups in the current `SimpleDeserializer` logic. While `GeneratedRegex` is a great modern alternative, it doesn't solve the allocation issues (as it still produces `Group` and `Capture` objects) and, as you noted, it isn't supported in `netstandard2.0`.

<details>
<summary>The complexity of the current Regex approach</summary>

```cs
private static string BuildContentLineRegex()
{
    // name          = iana-token / x-name
    // iana-token    = 1*(ALPHA / DIGIT / "-")
    // x-name        = "X-" [vendorid "-"] 1*(ALPHA / DIGIT / "-")
[...]

PR #912, opened 2026-01-22
`SimpleDeserializer` with improved performance

axunonb:
## Performance Optimization: Refactor SimpleDeserializer with Span-based Parsing

This PR refactors `SimpleDeserializer` to use allocation-optimized parsing, replacing the regex-based implementation with a span-based approach using `ReadOnlySpan<char>` and pooled buffers (`ArrayPool<char>`, `ZCharArray`).

**Key Changes:**
- Custom state machine for line reading/unfolding (replaces `TextReader.ReadLine()` + `StringBuilder`)
- Manual span-based parsing (replaces compiled `Regex`)
- Explicit handling of all line ending types (`\r\n`, `\n`, `\r`), but compatible with `TextReader.ReadLine()`.

**Performance Impact:**
- **25% faster** deserialization
- **28% fewer allocations**

*(Benchmark below: `version/6.0` baseline vs. this PR. **Component/property creation remains string-based in both versions.)**

**Breaking Changes:** None - public API unchanged.
Contributes to #908

### Current Bottleneck

In `SetPropertyValue`, we currently have this allocation hotspot:

```csharp
using var valueReader = new StringReader(value.ToString());  // ❌ Allocates string
// Missing overload for ReadOnlySpan<char>, so we use StringReader
var propertyValue = serializer?.Deserialize(valueReader);
```

Even though we have the property value as `ReadOnlySpan<char>`, we must:
1. Allocate a new `string` via `.ToString()`
2. Create a `StringReader` wrapper around it
3. Pass it to the serializer

This negates much of the benefit from the span-based line parsing.

### ~~Proposed~~ Solution

Add `ReadOnlySpan<char>` overloads to serializer classes:

```csharp
// IStringSerializer interface
public interface IStringSerializer : ISerializer
{
    object? Deserialize(TextReader tr);
    object? Deserialize(ReadOnlySpan<char> value);  // ✅ New overload
}

// SerializerBase implementation
public abstract class SerializerBase : IStringSerializer
{
    public virtual object? Deserialize(TextReader tr) { /* existing */ }
    public virtual object? Deserialize(ReadOnlySpan<char> value)  // ✅ New overload
    {
        // Default implementation for backward compatibility
        using var reader = new StringReader(value.ToString());
        return Deserialize(reader);
    }
}
```

Later update specific serializers to use span-based parsing.

### Expected Impact

- **Eliminates intermediate string allocations** for property values
- **Further 10-20% allocation reduction** (estimated)
- **Maintains backward compatibility** (old `TextReader` overload still works)

### Scope

This would require changes to:
- `IStringSerializer` interface
- `SerializerBase` base class
- ~~specific serializers in `Ical.Net\Serialization\DataTypes\`~~ Serializers will be updated in separate PR(s)

~~❓Expected changes from next step are small and could extend~~ -Extended this PR with another commit

**Breaking Changes:** Changed interface `IStringSerializer `.

---
```cs
[Benchmark]
public void DeserializeCalendar()
{
    const string iCalString =
        """
        BEGIN:VCALENDAR
        PRODID:-//github.com/ical-org/ical.net//NONSGML ical.net//EN
        VERSION:2.0
        BEGIN:VEVENT
        DTSTAMP:20250608T164638Z
        DTSTART:20250301T000000
        RRULE:FREQ=DAILY;COUNT=1000
        SEQUENCE:0
        UID:ac22036c-73e6-4020-b54a-80e580462749
        END:VEVENT
        END:VCALENDAR
        """;

    _ = Calendar.Load(new StringReader(iCalString));
}
```

```
BenchmarkDotNet v0.15.0, Windows 11 (10.0.26200.7623)
13th Gen Intel Core i7-13700K 3.40GHz, 1 CPU, 24 logical and 16 physical cores
.NET SDK 10.0.102
  [Host] : .NET 8.0.23 (8.0.2325.60607), X64 RyuJIT AVX2

Toolchain=InProcessNoEmitToolchain

| Method              | Mean       | Error     | StdDev    | Gen0    | Gen1    | Allocated |
|-------------------- |-----------:|----------:|----------:|--------:|--------:|----------:|
| version/6.0 Test    |  14.042 us | 0.0498 us | 0.0416 us |  3.1128 |  0.1526 |  47.89 KB |
| This PR Test        |  10.591 us | 0.1775 us | 0.1660 us |  2.2583 |  0.1068 |  34.67 KB |

```


maknapp:
I have looked only at the reading/unfolding part so far.

`ReadContentLines()` is buffering the entire reader before yielding anything, which is a major issue when handling large ics files. It should be deserializing to object as the buffer is read, with only folded lines needing to be copied to another buffer.

If a property's folded lines fit entirely within `buffer`, then the needed buffer size to unfold would be known and `unfoldedBuffer` could be rented directly with the required size.

Why is `ZCharArray` using `ArrayPool.Create()` instead of `ArrayPool.Shared`? I see no reason not to use shared.

axunonb:
> which is a major issue when handling large ics files

Question is what size is large. I checked with ~100 MB which I considered as large.
But anyway that's a simple change and reflects the intention of `TextReader.Read`, so I'll post an update.

axunonb:
> Why is `ZCharArray` using `ArrayPool.Create()` instead of `ArrayPool.Shared`? I see no reason not to use shared.

This question is related the first question, isn't it? If we expect filling 100MB+ `ArrayPool.Create()` is a good choice for configuring limits. If we rather have small allocations `ArrayPool.Shared` may be better. After refactoring Q1 I'll use `ArrayPool.Shared`.

axunonb:
> should be deserializing to object as the buffer is read, with only folded lines needing to be **copied to another buffer**.

This is a bit tricky when the goal is to avoid allocating a buffer per line on the heap.
`IEnumerable<ReadOnlyMemory<char>> ReadContentLinesUnsafe(TextReader reader)` now uses a *volatile buffer approach* for zero-allocation streaming:
Each yielded content line points to a shared buffer that is reused for subsequent lines. The caller must consume or copy the data before requesting the next line, as the buffer contents will be overwritten. This is ensured. I added a hint to the docs.
Implemented in [this commit](https://github.com/ical-org/ical.net/pull/912/changes/ad559880811143736f6dd6c94002945bfb3377a7).
Benchmark results shown above don't change.

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+
+namespace Ical.Net.Utility;
+
+/// <summary>
+/// A lightweight container that rents a char array as a buffer from an <see cref="ArrayPool{T}"/> and returns it when disposed.
+/// It simplifies passing around the buffer without intermediate memory allocations.
+/// <para/>
+/// Note that failing to dispose this struct after use will result in a memory leak.
+/// </summary>
+internal struct ZCharArray : IDisposable
```
maknapp: The name `ZCharArray` does not mean much to me. Does `PooledCharArray` or `PooledCharBuffer` sound better to you?

This class is similar to `System.Buffers.ArrayBufferWriter<T>` except using ArrayPool FYI.

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+{
+    // ArrayPool is thread-safe
+    private static readonly ArrayPool<char>
+        Pool = ArrayPool<char>.Shared;
+
+    private char[]? _bufferArray;
```
maknapp: This does not need to be nullable.

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+    private int _currentLength;
+
+    /// <summary>
+    /// The default capacity of the array.
+    /// </summary>
+    public const int DefaultBufferCapacity = 10_000_000;
```
maknapp: This can probably be `8_192` like the read buffer, or maybe even smaller. This is only used for one line at a time now, so it will rarely need to grow beyond that.

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +39,431 @@ internal SimpleDeserializer(
...
+    /// Each yielded ReadOnlyMemory&lt;char&gt; points to a shared buffer that is reused for subsequent lines.
+    /// Callers <b>MUST</b> consume or copy the data before requesting the next line, as the buffer contents
+    /// will be overwritten.
+    /// The caller is responsible for disposing the provided reader if necessary.
+    /// </remarks>
+    /// <param name="reader">The text reader from which to read the content lines. Cannot be null.</param>
+    /// <returns>
+    /// An enumerable collection of volatile Memory segments, each representing a logical content line with line folding removed.
+    /// The collection is empty if no lines are present. Each segment is valid only until the next iteration.</returns>
+    private static IEnumerable<ReadOnlyMemory<char>> ReadContentLinesUnsafe(TextReader reader)
```
maknapp: Having "Unsafe" in the name makes me immediately think there is an `unsafe` block within.

If the method were changed to something like:
```c#
private IEnumerable<CalendarProperty> DeserializeContentLines(TextReader reader)
{
   ...
   yield return ParseContentLine(unfoldedBuffer.Span);
   ...
}
```

That would:
* Remove the need for "unsafe" warnings
* Remove all usage of `ReadOnlyMemory`
* Remove need for `ZCharArray.UnderlyingArray`

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+    /// </summary>
+    public ZCharArray() : this(DefaultBufferCapacity)
+    {
+    }
+
+    /// <summary>
+    /// Creates a new <see cref="ZCharArray"/> with the specified length.
+    /// </summary>
+    /// <param name="length">The length of the array.</param>
+    public ZCharArray(int length)
```
maknapp: Could rename `length` to `initialLength` or `initialCapacity` and update summary comments of both constructors to say "initial length."

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+    {
+        _bufferArray = Pool.Rent(length);
+        _currentLength = 0;
+    }
+
+    /// <summary>
+    /// Gets the underlying character array used to store the buffer's contents.
+    /// </summary>
+    internal char[] UnderlyingArray
```
maknapp: Can remove if making the `ReadContentLinesUnsafe()` changes.

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+        data.CopyTo(_bufferArray!.AsSpan(_currentLength, data.Length));
+        _currentLength += data.Length;
+    }
+
+    /// <summary>
+    /// Writes the specified char to the array. Resizes the array if necessary.
+    /// </summary>
+    /// <param name="c">The char to write.</param>
+    /// <exception cref="ObjectDisposedException"></exception>
+    public void Write(char c)
```
maknapp: This can be removed since it is only used in tests and tests can use `Write(ReadOnlySpan<char>)`.

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+        GrowBufferIfNeeded(1);
+        _bufferArray![_currentLength++] = c;
+    }
+
+    private void GrowBufferIfNeeded(int dataLength)
+    {
+        var requiredLength = _currentLength + dataLength;
+        if (requiredLength <= Capacity) return;
+
+        Grow(requiredLength * 2); // Does nothing if the buffer is already large enough
```
maknapp: `Grow()` is only used here so it does not need to be a separate method.

Also the comment here makes it seem like `Grow()` will do nothing if it doesn't need to. Make more sense as the summary for `GrowBufferIfNeeded()`.

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -5,16 +5,30 @@
...
+/// The serializer parses unfolded iCalendar content lines and constructs a hierarchy of
+/// calendar components and their properties. It supports standard iCalendar line folding
+/// and parameter parsing rules.
+/// <para/>
+/// This class is thread-safe.
+/// Use the static Default instance for typical deserialization scenarios.
+/// </remarks>
 public class SimpleDeserializer
 {
+    private const int NotFound = -1;
```
maknapp: Comparing IndexOf results to -1 is pretty common, so this was more confusing to me than just using -1.

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +39,431 @@ internal SimpleDeserializer(
...
         }
-        if (current != null)
+
+        // Check for unclosed component
+        ThrowIfUnclosed(currentComponent);
+    }
+
+    private static void EnsureValueIsString(CalendarProperty calProperty, out string value)
+    {
+        value = string.Empty;
```
maknapp: This can go inside the below `if` statement.

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +39,431 @@ internal SimpleDeserializer(
...
+        finally
+        {
+            // Must be explicitly disposed, because 'using'
+            // is not allowed when passing the array by ref to methods.
+            unfoldedBuffer.Dispose();
+            System.Buffers.ArrayPool<char>.Shared.Return(buffer);
+        }
+    }
+
+    private static ParserState ProcessFoldCheck(
```
maknapp, line 347: Fix indentation of next several methods.

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +39,431 @@ internal SimpleDeserializer(
...
+    int charsRead,
+    ref ZCharArray unfoldedBuffer,
+    ref int i)
+{
+    // Look ahead to the next newline character so we can write chunks.
+    // This search is more efficient than writing one char at a time.
+    var searchSpan = buffer.AsSpan(i, charsRead - i);
+    var newlineIndex = searchSpan.IndexOfAny('\r', '\n');
+
+    switch (newlineIndex)
```
maknapp: Only two cases, change to `if (newlineIndex == -1)`?

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +39,431 @@ internal SimpleDeserializer(
...
-            property.AddParameter(parameter);
+            var c = remaining[i];
+            // Skip delimiter checks while inside quotes so values
+            // containing ,;: aren’t split prematurely
+            // Note: Quotes are expected to be balanced
+            inQuotes = ToggleInQuotes(c, inQuotes);
+            if (inQuotes) continue;
+
+            // Wait for a delimiter to extract the value
+            if (",;:".IndexOf(c) == NotFound) continue;
```
maknapp: Change this and the one below to `.Contains()`

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +39,431 @@ internal SimpleDeserializer(
...
-    private static void SetPropertyParameters(CalendarProperty property, CaptureCollection paramNames, CaptureCollection paramValues)
+    private static int ParseValues(ReadOnlySpan<char> remaining, CalendarParameter parameter, string nameEndChars)
     {
-        var paramValueIndex = 0;
-        for (var paramNameIndex = 0; paramNameIndex < paramNames.Count; paramNameIndex++)
+        var i = 0;
+        var inQuotes = false;
+        var valueStart = 0;
+
+        for (; i < remaining.Length; i++)
```
maknapp, line 206: Can this use `IndexOf()` and `IndexOfAny()` instead of looping each character?

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +39,431 @@ internal SimpleDeserializer(
...
-            {
-                var paramValue = paramValues[paramValueIndex].Value;
-                parameter.AddValue(paramValue);
-                paramValueIndex++;
-            }
-            property.AddParameter(parameter);
+            var c = remaining[i];
+            // Skip delimiter checks while inside quotes so values
+            // containing ,;: aren’t split prematurely
+            // Note: Quotes are expected to be balanced
```
maknapp, line 211: I don't see tests for bad quotes. Is this something that should be handled?

Ical.Net/Utility/ZCharArray.cs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// <summary>
+    /// Disposes the array, returning it to the <see cref="ArrayPool{T}"/>.
+    /// </summary>
+    public void Dispose()
+    {
+        if (IsDisposed) return;
+
+        Pool.Return(_bufferArray!);
```
maknapp: `System.Text.Json` clears rented arrays for security reasons. Should the array be cleared before being returned? Same with read buffer?

maknapp (changes requested):

Ical.Net/Utility/ZCharArray.cs
axunonb: zero-allocation character array

Ical.Net/Utility/ZCharArray.cs
axunonb: Since I'm using it to track disposal state via IsDisposed, I can use a non-nullable field and check for a sentinel value or add a separate _isDisposed bool flag.

Ical.Net/Utility/ZCharArray.cs
axunonb: That was left over from the previous implementation that accumulated all lines. Since we're now processing one line at a time, 8_192 (or even 4_096) makes much more sense. Most iCalendar lines are well under 1KB

Ical.Net/Serialization/SimpleDeserializer.cs
axunonb: Excellent, this is the million dollar hint! Moving the parsing inside the iterator would eliminate the volatile buffer concern entirely.

Ical.Net/Serialization/SimpleDeserializer.cs
axunonb: `Contains` does not work for `netstandard2.0`

Ical.Net/Serialization/SimpleDeserializer.cs
axunonb: This is tricky because we need to handle quoted strings where delimiters are escaped. The character-by-character loop tracks the inQuotes state to properly handle values like PARAM="value;with;semicolons". While I could use IndexOfAny() for the non-quoted portions, the state tracking makes it complex. I'd prefer to keep the current implementation.

Ical.Net/Serialization/SimpleDeserializer.cs
axunonb: Up to now we expect balanced quotes per RFC 5545. Malformed input like `PARAM="unclosed` would read beyond the intended value. Will add `SerializationException` if we reach the end of a line while still in quotes.

Ical.Net/Utility/ZCharArray.cs
axunonb: Okay, this has minimal performance impact and improves security posture.

Ical.Net/Utility/ZCharArray.cs
maknapp: `ArrayPool<T>.Rent` has a chance to allocate at any time though, so "zero-allocation" is not true.

Ical.Net/Serialization/SimpleDeserializer.cs
```diff
@@ -25,205 +37,428 @@ internal SimpleDeserializer(
...
+                as part of the same logical line
+        */
+
+        shouldYield = false;
+
+        // If the char after a newline is Space or Tab, it's a fold.
+        if (c == ' ' || c == '\t')
+        {
+            return ParserState.Normal;
+        }
```
maknapp: `shouldYield = false;` can go inside `if` block.

Ical.Net/Utility/PooledCharBuffer.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// </summary>
+    private void GrowBufferIfNeeded(int dataLength)
+    {
+        var requiredLength = _currentLength + dataLength;
+        if (requiredLength <= Capacity) return;
+
+        // Grow the buffer by renting a larger array and copying existing data
+        var newCapacity = requiredLength * 2;
+        var newArray = Pool.Rent(newCapacity);
+        Array.Copy(_bufferArray, newArray, Math.Min(_bufferArray.Length, newCapacity));
```
maknapp: `Array.Copy` length should only need to be `_currentLength`. Copies only the data being used and we already know it must be less than or equal to `bufferArray` and `newArray` lengths.

maknapp (changes requested):

Issue #913, opened 2026-01-25
Todo.IsCompleted evaluates all occurrences


maknapp:
If a TODO has an rrule without COUNT or UNTIL, `IsCompleted()` throws with out of range exception.

axunonb:
Good catch! Yes, should go into v5 as well.

PR #914, opened 2026-01-25
Stop todo evaluation at given datetime

maknapp:
This is a v6 fix for #913. Should probably be fixed for v5 too. I am still trying to make sense of the TODO stuff, but I think `TakeWhile` was the original intention here.

### DIFF
--- a/Ical.Net.Tests/TodoTest.cs
+++ b/Ical.Net.Tests/TodoTest.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Ical.Net.CalendarComponents;
 using Ical.Net.DataTypes;
+using NodaTime;
 using NUnit.Framework;
 
 namespace Ical.Net.Tests;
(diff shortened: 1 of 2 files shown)


Ical.Net.Tests/TodoTest.cs
```diff
@@ -238,4 +239,31 @@ public void Todo_WithFutureStart_AndNoDuration_ShouldSucceed()
...
+            .AtMidnight()
+            .InUtc()
+            .ToInstant());
+
+        var results = Enumerable.Range(14, 10)
+            .Select(x => new LocalDate(2026, 1, x))
+            .Select(x => todo.IsCompleted(x.AtMidnight().InUtc()))
+            .ToList();
+
+        List<bool> expected = [true, true, true, true, true, true, true, true, false, false];
```
axunonb, line 267: // Jan 14-21: true  (8 days) - TODO is considered completed
// Jan 22-23: false (2 days) - TODO is NOT completed

maknapp:
This can be closed if #922 is merged because it deletes these changes.

PR #923, opened 2026-02-01
Run file-based test cases with each instance as DTSTART

minichma:
This PR extends the file-based test cases to run each test case not only with the specified DTSTART but also with each of the expected recurrence instances as DTSTART as individual test case. This tests is helpful because in most test cases DTSTART is specified at the beginning of the respective interval. With this extension we also test for less common, yet valid patterns, i.e. DTSTART being specified not at the beginning of an interval.

The new test cases actually bring up some failing tests that should be fixed before this one can be merged.

The PR also contains some additional tests that cover similar cases where DTSTART is not at the beginning of an interval but run nicely.


axunonb:
@minichma All the tests seem to pass with v6 now. Would you like to update the PR?

minichma:
> @minichma All the tests seem to pass with v6 now. Would you like to update the PR?

@axunonb Thanks for the friendly reminder. I just pushed an update, should be good for review.

axunonb (approved):
Great, thanks a lot!

PR #925, opened 2026-02-02
Handle byweekno spanning months

### LINKED ISSUE #924: V6: Tests failing if DTSTART is not at the beginning of an interval
The following test cases fail:

```ics
# 30.12 in WEEK 52 with expansion on BYMONTH, BYMONTHDAY, BYWEEKNO
RRULE:FREQ=YEARLY;BYMONTH=12;BYMONTHDAY=1,30;BYWEEKNO=52;UNTIL=20340101
DTSTART:20291230
INSTANCES:20291230,20331230

# 30.12 in WEEK 52 with expansion on BYYEARDAY, BYWEEKNO
RRULE:FREQ=YEARLY;BYYEARDAY=-2;BYWEEKNO=52;UNTIL=20340101
DTSTART:20291230
INSTANCES:20291230,20331230
```

They've been identified as part of #923. The more general pattern I was trying to test for was the case that DTSTART is not specified at the beginning of an interval. E.g. we have `BYMONTH=1,2` and DTSTART is specified with month 2 rather than 1. We don't seem to cover this case explicitly in the code and that it mostly works seems to be a rather subtle (and undocumented) consequence of how recurrence evaluation is implemented. I didn't analyze the test cases above in more detail, so they may or may not be related to the described pattern. Actually the second test doesn't have any `BY` with multiple values, so from that perspective it seems to be a different problem.

The test cases seem to succeed in V5.

Version:
V6, based on Commit [#ed3d40d677ea924db594b1c9efa3e8ad42632eda](https://github.com/ical-org/ical.net/tree/ed3d40d677ea924db594b1c9efa3e8ad42632eda)

maknapp:
Changes BYMONTH + BYWEEKNO to start expanding by weeks and limiting months later.

Fixes #924


Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -452,11 +456,25 @@ private IEnumerable<ZonedDateTime> ByDaySpecialYearly()
         return ExpandDayFromYear();
     }
 
+    /// <summary>
+    /// Limits weeks by month.
+    /// </summary>
+    private IEnumerable<ZonedDateTime> LimitWeekByMonth()
+    {
+        foreach (var value in ExpandWeekNo())
```
axunonb: Add `//NOSONAR` - SonarCloud prefers to see LINQ, although we still prefer loops

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -922,7 +940,7 @@ private IEnumerable<ZonedDateTime> ByYearDay()
 
     private IEnumerable<ZonedDateTime> LimitYearDay()
     {
-        foreach (var value in ByWeekNo())
+        foreach (var value in ByMonth())
```
axunonb: Add `//NOSONAR` - SonarCloud prefers to see LINQ, although we often still prefer loops

axunonb (approved):
Quite a bit of refactoring to get #924 fixed. Great!
BTW: Till 25 Feb I'm OOO with limited availability.

axunonb:
@minichma Could you take a look, please?

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -407,26 +407,30 @@ private IEnumerable<ZonedDateTime> ExpandHour()
...
 
-        return _frequency switch
+        if (_rule.ByWeekNo)
         {
-            FrequencyType.Weekly => ExpandDayFromWeekWithoutOffsets(),
-            FrequencyType.Monthly => ByDaySpecialMonthly(),
-            FrequencyType.Yearly => ByDaySpecialYearly(),
-            _ => LimitDayOfWeek()
-        };
+            return _rule.ByMonth ? LimitWeekByMonth() : ExpandWeekNo();
```
minichma, line 430: This method seems to be one of the more complex ones. Some more comments on the individual cases would therefore be helpful, ideally citing the RFC.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
```diff
@@ -407,26 +407,30 @@ private IEnumerable<ZonedDateTime> ExpandHour()
...
+        }
+
+        return LimitMonthByReferenceDay();
```
minichma, line 433: The method structure is not always quite obvious. Usually I understand we go from one level to the next, according to the order imposed by the RFC. I.e. `BYMONTH, BYWEEKNO, BYYEARDAY, BYMONTHDAY, BYDAY, BYHOUR, BYMINUTE, BYSECOND and BYSETPOS` in opposite order. One level takes care of calling the next. In `ByDay()` we have some special cases according to the expand matrix. The individual levels take care of knowing whether to expand or to limit, e.g. `ByMonthDay()` calls `ByYearDay()`, etc. From that perspective it is not quite obvious why we have a `ByMonth()` and a `LimitMonthByReferenceDay()`. I feel that it could either be harmonized or alternatively the individual versions would deserve some more descriptive names and comments where they are called and why. But that's probably not a topic of this PR.

minichma (approved):
I was trying to review the change but had a little bit of a hard time because I'm not familiar with the rewritten RecurrencePatternEvaluator yet. I invested a few hours and was trying to find some more failing edge cases but didn't find any, so from that point the change LGTM. I haven't managed to go through all cases though, so the confidence of the review is somewhat limited.

A thought on what would have made the review somewhat easier. The changes seem to consist of multiple, logically independent steps. I.e. it seems that code has been refactored in order to apply subsequent functional changes. Code refactoring often consist of more lines being changed than the functional ones while they usually need less careful review, as the behavior of the refactored code should stay the same. It would therefore be helpful to structure the commits in more fine-grained manner, i.e. have each independent refactoring in a separate commit and make it clear in the commit message that the change shouldn't contain any functional change. Then keep the functional changes as small as possible, so they are easier to read and commit them separately. If there's just a single commit like in this PR, a reviewer has to read and understand all the changes at once, which is significantly harder, more error prone and takes accordingly longer.

Generally speaking I feel the code has improved a lot over the v5 version. I also feel that the level of comments has improved. Nevertheless I feel that it would be helpful to add more citations to the RFC, especially where branches might not be too obvious or where we made decisions due to a lack of clarity of the RFC. Maybe that's also something AI could help with.

Ical.Net/Evaluation/RecurrencePatternEvaluator.cs
axunonb: @maknapp Not sure whether you intend to modify this PR in terms of the latest comments, or squash merge for now?

maknapp: This is fine to merge now. I do agree that more comments could be added in places, but can be in new PR.

PR #926, opened 2026-02-08
Test: Update `icalrecur_test.txt` from libical master

minichma:
`icalrecur_test.out` contains test cases from the https://github.com/libical/libical project. Update the file to the latest upstream master.

Two test cases were failing (on master its 3) that are related to `BYWEEKNO`, which is implemented differently in libical than in Ical.Net. They have been disabled.


axunonb (approved):
Excellent, thanks for the update. Nice increase in coverage, too.

Issue #910, opened 2026-01-18
Drop support for more than one RRULE property


maknapp:
RFC 5545:
> The "RRULE" property SHOULD NOT occur more than once in a component.

The spec does not allow more than one rrule for any component. Can we remove support for multiple rules in v6?

axunonb:
RFC 5545 discourages multiple RRULEs because they make the calendar's behavior undefined.
Even though it's a "SHOULD", it makes sense to disallow in v6.

`RecurrenceRuleTests` in `Ical.Net.Tests/RecurrenceTests.cs` exercises a component that has two `RecurrenceRules` added (`eventA.RecurrenceRules.Add(five); eventA.RecurrenceRules.Add(ten);` and the same for `eventB` with swapped order) to ensure equality/serialization logic still holds when a single component hosts more than one `RRULE`.

maknapp:
Should `IList<RecurrencePattern> RecurrenceRules` be marked obsolete in v6 or replaced with `RecurrencePattern?`? Is it too late to mark as obsolete in v5?

If the list property needs to be kept in v6 and marked obsolete, should it still handle all rules in the list or just use the first rule in the list only?

axunonb:
We marked `Exceptionrules` as obsolte in v5.
I think we can do so for `IList<RecurrencePattern> RecurrenceRules` in v5, too.

Would you prefer `RecurrencePattern RecurrencePattern?` for the property instead of `RecurrencePattern RecurrenceRule?`
Maybe introduce the replacement in v5 and fully remove `IList<RecurrencePattern> RecurrenceRules` in v6?

maknapp:
I think the names should match. I would actually prefer the `RecurrencePattern` class to be renamed to `RecurrenceRule` since it is more commonly referred to as a rule instead of a pattern. If we do not want to go that route, then changing the property name to `RecurrencePattern` is the next best option.

axunonb:
Yes, the dilemma is that the property name `Exceptionrules` is fine, while the class name `RecurrencePattern` is not.
If we mark  `IList<RecurrencePattern> RecurrenceRules` as obsolete in favor of `RecurrenceRule? RecurrenceRule` in v5, we get a similar property name but with a different type. Only `RecurrenceRule? RecurrenceRule`  will exist in v6. 
How does that sound to you?

maknapp:
That sounds good to me.

PR #928, opened 2026-02-13
Deprecate RecurrenceRules and RecurrencePattern

maknapp:
Related #910

Deprecates `IList<RecurrencePattern> RecurrenceRules` in favor of `RecurrenceRule? RecurrenceRule`.

This is for preparing to rename `RecurrencePattern` and `RecurrencePatternSerializer` in v6 to use "Rule" instead of "Pattern". The only class left with "Pattern" in the name is `RecurrencePatternEvaluator` which is internal in v6.

Some of the changes made to deprecate `RecurrencePattern` may cause some breaking changes depending on usage. There are a _lot_ of public virtual methods and I am not sure how those can cause problems. I am also not sure if I handled `CalendarDataType` correctly.


Ical.Net/DataTypes/RecurrenceRule.cs
```diff
@@ -0,0 +1,183 @@
...
+        return serializer.SerializeToString(this);
+    }
+
+    /// <inheritdoc/>
+    public override void CopyFrom(ICopyable obj)
+    {
+        base.CopyFrom(obj);
+        if (obj is not RecurrenceRule r)
+        {
+            // Allow copying from rule wrapped by pattern
```
axunonb: Remove uncommented code?

axunonb (dismissed):
Excellent, thanks! Just a minor comment.

maknapp:
Extra commented code removed.

axunonb (approved):
👍

PR #930, opened 2026-02-28
Remove date and time arithmetic from CalDateTime

### LINKED ISSUE #849: `CalDateTime.AddDuration` vs. `CalDateTime.AddDay`

maknapp:
From discussion in #849 and #854, one way to resolve math issues is to leave math to `NodaTime` (or `System.DateTime`). This removes all `Add*()` methods from `CalDateTime` and replaces usage by converting to `NodaTime` and back.

Doing this adds more support towards having `NodaTime` in the public API. Making it easy to convert to and from `NodaTime` types is important.

While this change does make sense, it does complicate things significantly for users who use `Add*()`. For those unfamiliar with `NodaTime`, they would have to learn `NodaTime` while also understanding that `CalDateTime` can hold date, datetime, or datetime with time zone without offset. `CalDateTime` will need to be converted to the correct `NodaTime` type based on the data it holds and what date/time units are being added. This will need to be well documented.

Fixes #849

### DIFF
--- a/Ical.Net/DataTypes/CalDateTime.cs
+++ b/Ical.Net/DataTypes/CalDateTime.cs
@@ -464,24 +464,6 @@ public sealed class CalDateTime : IFormattable, IEquatable<CalDateTime>
         return new(ToZonedDateTime(otherTzId));
     }
 
-    /// <inheritdoc cref="DateTime.AddYears"/>
-    public CalDateTime AddYears(int years)
-    {
-        return new(_localDate.PlusYears(years), _localTime, _tzId);
-    }
-
-    /// <inheritdoc cref="DateTime.AddMonths"/>
-    public CalDateTime AddMonths(int months)
-    {
-        return new(_localDate.PlusMonths(months), _localTime, _tzId);
-    }
-
-    /// <inheritdoc cref="DateTime.AddDays"/>
-    public CalDateTime AddDays(int days)
-    {
-        return new(_localDate.PlusDays(days), _localTime, _tzId);
-    }
-
     /// <inheritdoc />
     public override string ToString() => ToString(null, null);
 
--- a/Ical.Net/DataTypes/CalDateTime.cs
+++ b/Ical.Net/DataTypes/CalDateTime.cs
@@ -464,64 +464,6 @@ public sealed class CalDateTime : IFormattable, IEquatable<CalDateTime>
         return new(ToZonedDateTime(otherTzId));
     }
 
-    /// <summary>
-    /// Add the specified <see cref="Duration"/> to this instance/>.
-    /// </summary>
-    /// <remarks>
-    /// In correspondence to RFC5545, the weeks and day fields of a duration are considered nominal durations while the time fields are considered exact values.
-    /// </remarks>
-    /// <exception cref="InvalidOperationException">
-    /// Thrown when attempting to add a time span to a date-only instance, 
-    /// and the time span is not a multiple of full days.
-    /// </exception>
-    public CalDateTime Add(Duration d)
-    {
-        // If NOOP
-        if (d.IsEmpty) return this;
-
-        if (!HasTime && d.HasTime)
-        {
-            throw new InvalidOperationException($"This instance represents a 'date-only' value '{ToString()}'. Only multiples of full days can be added to a 'date-only' instance, not '{d}'");
-        }
-
-        // RFC 5545 3.3.6:
-        // If the property permits, multiple "duration" values are specified by a COMMA-separated
-        // list of values.The format is based on the[ISO.8601.2004] complete representation basic
-        // format with designators for the duration of time.The format can represent nominal
-        // durations(weeks and days) and accurate durations(hours, minutes, and seconds).
-        // Note that unlike[ISO.8601.2004], this value type doesn't support the "Y" and "M"
-        // designators to specify durations in terms of years and months.
-        //
-        // The duration of a week or a day depends on its position in the calendar. In the case
-        // of discontinuities in the time scale, such as the change from standard time to daylight
-        // time and back, the computation of the exact duration requires the subtraction or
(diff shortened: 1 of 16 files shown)


axunonb:
> While this change does make sense, it does complicate things significantly for users who use `Add*()`. For those unfamiliar with `NodaTime`,

We already have a couple of breaking changes in terms of `CalDatetime`
- `Date` and `Time` now return NodaTime types
- Comparison operators and several methods removed
- Replaced `CalDateTime` for occurrences

So this step is just kind of finalizing the direction we took in v6.

As you know I was hoping for an easy migration path from v5 to v6. Now v6 requires a direct dependency on [NodaTime](https://nodatime.org/) and users will have to learn it and rewrite their current `ical.net` integration.

Ical.Net/DataTypes/CalDateTime.cs
axunonb, line 231: Could be removed in favor of `NodaTime`.

axunonb, line 374: Could be removed in favor of `NodaTime`.

axunonb, line 412: Redundant `else`

axunonb, line 424: Redundant `else`

axunonb, line 434: Maybe use expression body?

axunonb, line 443: Redundant `else`

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -464,103 +464,6 @@ public CalDateTime ToTimeZone(string? otherTzId)
         return new(ToZonedDateTime(otherTzId));
```
axunonb, line 464: Add a comment that this drops the timezone offset and converting back to ZonedDatetime will not resolve the same value?

Ical.Net/DataTypes/Duration.cs
```diff
@@ -59,7 +62,7 @@ public NodaTime.Period ToPeriod()
         return b.Build();
     }
 
-    public NodaTime.Period GetNominalPart()
+    public readonly NodaTime.Period GetNominalPart()
```
axunonb, line 65: Good catch - compiler knows no mutation occurs, no copy needed

Ical.Net/Utility/DateUtil.cs
axunonb, line 11: Do we still need this class?
`WeekOfMonth(DateTime d)` has just one reference

axunonb (dismissed):
See few minor comments

Ical.Net/DataTypes/Duration.cs
```diff
@@ -45,7 +45,10 @@ public Duration(int? weeks = null, int? days = null, int? hours = null, int? min
         Seconds = seconds;
     }
 
-    public NodaTime.Period ToPeriod()
+    public static Duration FromPeriod(NodaTime.Period p)
```
axunonb, line 48: Not for this PR: Should we remove the `Period` ambiguity between NodaTime and ical.net by using another name, e.g. `CalPeriod`?

maknapp: I would like all NodaTime name overlaps to be renamed - `CalPeriod`, `CalDuration`, and any others.

Ical.Net/DataTypes/CalDateTime.cs
maknapp: I think it is better to keep these. These are simple convenience methods that do not add complexity. For those who refuse to use NodaTime, these could be more useful than being stuck with only `DateTime`.

maknapp:
> Do we still need this class? `WeekOfMonth(DateTime d)` has just one reference

I have been looking at #749 and handling time zones defined in the calendar. This will remove the last major reference to `DateUtil`.

> Add a comment that this drops the timezone offset and converting back to ZonedDatetime will not resolve the same value?

This and the many redundant `else` issues can be in a separate PR. It might make more sense to move all `ToZonedDateTime()` related methods from `CalDateTime` to a `TimeZoneResolver` that can handle calendar time zones.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Those ical.net users who refuse to use NodaTime will have a difficult time ahead of them 😉

axunonb (approved):
LG, tnx

axunonb:
@maknapp After merging this PR I performed a rebase of `version/6.0`. (There's a temporary `version/6.0-backup` with a copy before the rebase.)
I added a commit [Fix file encoding: convert .cs files from UTF-16 LE to UTF-8](https://github.com/ical-org/ical.net/commit/58494b413642d19dc8928b268cdd673da61a6249). This mixed encoding gave me a hard time during the rebase.
Probably the `.gitattributes` and `.editorconfig` must be updated. Then this should not happen again. I'll submit a PR for that.

PR #932, opened 2026-03-16
Update encoding and line endings for consistency

axunonb:
`.editorconfig` now enforces UTF-8 encoding for all files. 
`.gitattributes` sets LF line endings and auto text normalization for C# files

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,10 @@
 # Remove the line below if you want to inherit .editorconfig settings from higher directories
 root = true
 
+[*]
+# editors create UTF-8 only
+charset = utf-8
+
 [*.{cs,xml,csproj,sln}]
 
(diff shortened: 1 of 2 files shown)


maknapp (approved):

maknapp:
@axunonb I think auto line endings in gitattributes is making my local tests fail now.  It seems some of the wiki tests depend on file line endings, and after git touches RecurrenceWikiTests.cs in Windows, the line endings are changed.

axunonb:
iCalenders should use CRLF. But with the LF line endings the expected results don't fit because of raw strings. So

- Either the wiki tests must normalize line endings before the assertions
- or we remove the wiki tests altogether. We have the examples from the wiki tests as part of the wiki. 

[Edit]
Fixed with #938

PR #934, opened 2026-03-22
Simplify and cleanup CalDateTime

maknapp:
The intent here is to remove more from CalDateTime while keeping it simple to use. See individual commits for changes. I can change to separate PRs if that is preferred.

### Breaking Changes:
1. `ToTimeZone(string?)` removed
2. `AsUtc` property removed
3. `Value` property removed
4. `CalDateTime(DateTime, string? tzId, bool? hasTime = null)` removed
5. `CalDateTime(DateTime, bool? hasTime = null)` removed
7. `CalDateTime(DateOnly date)` -> `CalDateTime FromDateOnly(DateOnly date)` (and only .NET 6+)
8. `CalDateTime(DateOnly date, TimeOnly? time, string? tzId = null)` removed
9. `Hour`, `Minute`, `Second` properties removed


maknapp:
@axunonb version/6.0 is protected and never supposed to move. Why was version/6.0 rebased? Losing all commit dates... can we change it back?

axunonb:
See https://github.com/ical-org/ical.net/pull/930#issuecomment-4064180875
There is a branch [version/6.0-backkup](https://github.com/ical-org/ical.net/tree/version/6.0-backup) that I created before the rebase.
If you think the **commit** dates are important, we can go back to this backup.
The **author** dates are untouched - see `git log --pretty=fuller`

maknapp:
I remember reading that comment now and was confused then but skipped over it and forgot to go back. I do not like changing history of any shared/major branches. Looking back to https://github.com/ical-org/ical.net/pull/854#issuecomment-3620708025, I was assuming this meant that version/6.0 was eventually going to be renamed to master or merged with a merge commit, not rebased. Rebasing a large branch like this repeatedly to keep up-to-date with master also makes links and history in github more confusing. I would strongly prefer to go back to the original version/6.0 branch and never rebase it.

axunonb:
Sure, feel free to recreate branch version/6.0 from [version/6.0-backkup](https://github.com/ical-org/ical.net/tree/version/6.0-backup) the way you prefer. The latest 3 commit are missing in the backup.

maknapp:
I cannot force push to version/6.0 branch because it is restricted. I pushed a v6 branch which is the original with 2 of the latest commits. The encoding commit 58494b41 is a noop for me and just produces an empty commit.

axunonb:
> cannot force push to version/6.0 branch

The assumption was you can as a maintainer, sorry. You're now entitled to force push, while it is generally blocked.
Btw. would like to become an owner? Would make sense.

maknapp:
Still blocked with:
> ! [remote rejected]   version/6.0 -> version/6.0 (protected branch hook declined)"

We could just switch to using "v6" (shorter name is better?) and mark it protected.

> Btw. would like to become an owner?

Sure, that would work too.

axunonb:
@maknapp It would be helpful to rebase this branch before the review.

maknapp:
See previous comment. I still cannot force push version/6.0 branch to revert it back to the previous base commit. This PR is based off of the original version/6.0 branch.

axunonb:
> See previous comment

image
That's the setting that should allow you to force-push. You're also admin of the repo. Maybe try again?

maknapp:
Okay pushing worked now. I think everything is back to what it was except commit 58494b41 which was empty for me. Verify your files are still UTF-8 after switching to current version/6.0 branch.

axunonb:
> files are still UTF-8 after switching to current version/6.0 branch

```bash
git ls-files | ForEach-Object { $b=[IO.File]::ReadAllBytes($_); $e=if($b.Length-ge 3-and $b[0] -eq 0xEF-and $b[1] -eq 0xBB-and $b[2] -eq 0xBF){"UTF-8-BOM"}elseif($b.Length-ge 2-and $b[0] -eq 0xFF-and $b[1] -eq 0xFE){"UTF-16LE"}elseif($b.Length-ge 2-and $b[0] -eq 0xFE-and $b[1] -eq 0xFF){"UTF-16BE"}else{try{[Text.UTF8Encoding]::new($false,$true).GetString($b)|Out-Null;"UTF-8"}catch{"OTHER"}}; if($e -ne "UTF-8"){[PSCustomObject]@{Encoding=$e;File=$_}} } | Format-Table -AutoSize
```
reported no files that are not `UTF-8`, so all are 'clean'. Thanks!

[Edit]
With this chore done, we should remove the backup branch, right?

~~#931 isn't included in v6 now, while it should. Maybe fix in this PR.~~  See #936

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -16,47 +16,49 @@
...
 /// <para/>
-/// Any <see cref="Time"/> values are always rounded to the nearest second.
-/// This is because RFC 5545, Section 3.3.5, does not allow for fractional seconds.
-/// </remarks>
+/// <see cref="Time"/> values are rounded to the nearest second.
+/// Time zone offset is <i>not</i> stored.
+/// <para/>
+/// See RFC 5545, Section 3.3.4 and 3.3.5.
 /// </summary>
 public sealed class CalDateTime : IFormattable, IEquatable<CalDateTime>
```
axunonb, line 26: The **breaking public API changes** with no deprecation step should be clearly called out with the break in the PR description. This will be helpful for any changelog/migration docs.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -16,47 +16,49 @@
...
 
     /// <summary>
-    /// Creates a new instance of the <see cref="CalDateTime"/> class
-    /// with the current date/time in the Coordinated Universal Time (UTC) timezone.
+    /// Creates a <see cref="CalDateTime"/> representing a DATE value.
     /// </summary>
-    public static CalDateTime UtcNow => new CalDateTime(DateTime.UtcNow, UtcTzId, true);
+    /// <param name="value">The value to copy the local date from.</param>
+    /// <returns>A new <see cref="CalDateTime"/> with same date as the specified <see cref="DateTime"/>.</returns>
+    public static CalDateTime FromDate(DateTime value) => new(LocalDate.FromDateTime(value));
```
axunonb: If DATE creation is moved to a static factory like with this method, it should follow the From* / To* convention already in use. The PR description mentions this but doesn't yet clarify the chosen name(s).

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -186,30 +158,36 @@ public CalDateTime(LocalDate date, LocalTime? time, string? tzId = null)
...
+    /// it cannot have a time zone, and the <see paramref="tzId"/> will be ignored.
     /// <para/>
     /// If the parsed string represents an RFC 5545, DATE-TIME value, the <see paramref="tzId"/> will be used.
     /// </param>
-    /// <param name="tzId">A timezone of <see langword="null"/> represents
-    /// a floating date/time, which is the same in all timezones. (<seealso cref="UtcTzId"/>) represents the Coordinated Universal Time.
-    /// Other values determine the timezone of the date/time.
-    /// </param>
+    /// <param name="tzId">The time zone ID.</param>
     public CalDateTime(string value, string? tzId = null)
```
axunonb, line 145: This allocates a `DateTimeSerializer` and an intermediate `CalDateTime` on every call, which is not addressed in this PR. Maybe track as a follow-up issue? (A static DateTimeSerializer instance or a TryParse/Parse(string) static method would be cleaner.)

XML doc on the `<see paramref=...>` tags use wrong syntax (paramref vs paramref); this is present in the base file but worth fixing now.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -385,83 +358,116 @@ public override bool Equals(object? obj)
...
     /// <para/>
-    /// If <see cref="IsFloating"/>==<see langword="true"/>
-    /// it means that the <see cref="Value"/> is considered as local time for every timezone:
-    /// The returned <see cref="Value"/> is unchanged and the <see paramref="otherTzId"/> is set as <see cref="TzId"/>.
+    /// DATE values will default to <see cref="LocalTime.Midnight"/>.
+    /// <para/>
+    /// Values without a time zone will be treated as being in the specified time zone.
+    /// If the local date and time is ambiguous due to the time zone, it will be resolved using <see cref="NodaTime.TimeZones.Resolvers.LenientResolver"/>.
     /// </summary>
-    public CalDateTime ToTimeZone(string? otherTzId)
```
axunonb, line 457: See also the general remark about breaking changes.
Even if only used in tests internally, ToTimeZone(string?) is a public method on a public sealed type. Removing it is a breaking change.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -68,115 +70,85 @@ private CalDateTime()
...
-    /// a floating date/time, which is the same in all timezones. (<seealso cref="UtcTzId"/>) represents the Coordinated Universal Time.
-    /// Other values determine the timezone of the date/time.
-    /// </param>
     /// <param name="year"></param>
     /// <param name="month"></param>
     /// <param name="day"></param>
     /// <param name="hour"></param>
     /// <param name="minute"></param>
     /// <param name="second"></param>
+    /// <param name="tzId">The time zone ID.</param>
```
axunonb, line 75: We can expect the `//NOSONAR` suppression on the (int, int, int, int, int, int, string?) constructor still applies?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -385,83 +358,116 @@ public override bool Equals(object? obj)
...
-    /// The returned <see cref="Value"/> is unchanged and the <see paramref="otherTzId"/> is set as <see cref="TzId"/>.
+    /// DATE values will default to <see cref="LocalTime.Midnight"/>.
+    /// <para/>
+    /// Values without a time zone will be treated as being in the specified time zone.
+    /// If the local date and time is ambiguous due to the time zone, it will be resolved using <see cref="NodaTime.TimeZones.Resolvers.LenientResolver"/>.
     /// </summary>
-    public CalDateTime ToTimeZone(string? otherTzId)
+    /// <param name="defaultZone">The time zone to use if this value has no time zone.</param>
+    /// <returns>A zoned date time representing this value in same time zone or the specified time zone.</returns>
+    public ZonedDateTime AsZonedOrDefault(DateTimeZone defaultZone)
```
axunonb: `AsZonedOrDefault(DateTimeZone)` has an inconsistent name relative to `ToZonedDateTime()`. 
As we're tidying the API, this might be a good time to rename it to something like `ToZonedDateTimeOrFloating`?

axunonb (dismissed):
Nice! See some mostly minor inline comments.

As for SonarCloud "[Refactor 'GetHashCode' to not reference mutable fields.](https://sonarcloud.io/project/issues?pullRequest=934&open=AZ0WVUdA-tgOj4n6RzTQ&id=ical-org_ical.net)" this is unfortunately the case for many other classes. Can be kept unchanged for now.

Maybe have a quick look at the [Codecov](https://app.codecov.io/gh/ical-org/ical.net/pull/934?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ical-org) Report, especially for `CalDateTime` decreasing coverage.

[Edit re `CalDateTime` coverage]:
`public ZonedDateTime ToZonedDateTime()`: Would be good to cover `InvalidOperationException` for `_tzId == null`?
`public CalDateTime(LocalDate date)` cover `ArgumentOutOfRangeException`?
The other uncovered methods are related to `DateOnly` / `TimeOnly` - not strictly necessary to cover

axunonb:
> Remove Hour/Minute/Second, and Year/Month/Day

If I remember correctly, this what I mentioned some time ago, and the reply was `NodaTime` supply these methods too.
So it would simplify the API on one side, but the simple properties don't harm on the other side. Take your choice.

Ical.Net/DataTypes/CalDateTime.cs
maknapp: I have been thinking about different options for this. It definitely cannot be `FromDateTime()` because it does not say a DATE value is being created. Some options:

```c#
// Clearly states the type and value used; will saying date-time-date confuse people?
public static CalDateTime FromDateTimeDate(DateTime value);

// Shortest name; implies date only but someone could assume time is included
public static CalDateTime FromDate(DateTime value);

// Change DateOnly constructor to static too?; Might confuse people into thinking value is DateOnly
public static CalDateTime FromDateOnly(DateTime value);
public static CalDateTime FromDateOnly(DateOnly value);

// Longest name; explicitly says type and that it will be a DATE value
public static CalDateTime FromDateTimeDateOnly(DateTime value);
```

Which do you like best, or do you have another name in mind?

Ical.Net/DataTypes/CalDateTime.cs
maknapp: Created issue #937 and fixing doc here.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: I'd say the cleanest approach is **`FromDate(...)` with overloads**:

The name describes the **result** (a DATE value), not the input type — which is the most important thing to communicate. It also mirrors the BCL precedent of `DateOnly.FromDateTime()`, where users already understand the pattern of "create this date-focused type from that input". The overload family removes any residual ambiguity:

```csharp
// Ical.Net\DataTypes\CalDateTime.cs
public static CalDateTime FromDate(int year, int month, int day)
    => new CalDateTime(new LocalDate(year, month, day), null, null);

public static CalDateTime FromDate(DateTime value)
    => new CalDateTime(LocalDate.FromDateTime(value), null, null);

public static CalDateTime FromDate(LocalDate value)
    => new CalDateTime(value, null, null);

#if NET6_0_OR_GREATER
public static CalDateTime FromDate(DateOnly value)
    => new CalDateTime(value.ToLocalDate(), null, null);
#endif
```

Once a user sees any one overload, all others are predictable.

Another consideration:

If this factory lands, the `DateTime` DATE-TIME constructor could become a paired `FromDateTime(...)` static factory:

```csharp
public static CalDateTime FromDateTime(DateTime value, string? tzId = null) { ... }
```

That would give a symmetric, discoverable pair: **`FromDate`** for RFC 5545 DATE, **`FromDateTime`** for RFC 5545 DATE-TIME - directly matching the spec's own type names.

Would you follow this idea?

Ical.Net/DataTypes/CalDateTime.cs
maknapp: I named this completely different on purpose to prevent people from confusing the intention of the argument. The time zone argument for the ToZonedDateTime methods is for converting to that time zone, whereas this one only adds the specified time zone if there is none.

I am open to other names, but I do not like `ToZonedDateTimeOrFloating` because my first thought is the result can somehow be floating.

Ical.Net/DataTypes/CalDateTime.cs
maknapp: `FromDate` goes against BCL precedent, not the same. All `From*` static constructors that I can think of describe the input type to convert from, never the result (a DATE value). I cannot think of any types with multiple `FromX` overloads like this.

I would like converting the `DateTime` constructor to `CalDateTime.FromDateTime()`. It is generally not good to have both.

If we wanted to match the BCL and NodaTime, it should be just `FromDateTimeDate` with no overloads. That would be most similar to things like `DateTimeOffset.FromUnixTime*` or `LocalTime.From*`.

```c#
public static CalDateTime FromDateTimeDate(DateTime value);
public static CalDateTime FromDateTime(DateTime value);
public static CalDateTime FromDateOnly(DateOnly value);
public static CalDateTime FromInstant(Instant value);
```

maknapp:
This latest commit [d887305](https://github.com/ical-org/ical.net/pull/934/commits/d887305d7603be3320c5d7fa03758a569fbc3a7a) removes the constructors with separate date and time in favor of just `DateTime` or `LocalDateTime`. It also changes some of the constructors to `From*` and moves them next to the associated `To*` to match https://github.com/ical-org/ical.net/pull/934#discussion_r3029665676. Thoughts?

Ical.Net/DataTypes/CalDateTime.cs
maknapp: I removed the NOSONAR comment and there are no sonar warnings.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: `From*` describes the input - my wrong.
So yes, then this makes sense.

axunonb:
> Thoughts

Few things come to my mind:

* ZonedDateTime/Instant got factories, while CalDateTime(LocalDateTime, string? tzId) stays as a public constructor. By intension?
* Generally: If we have a factory, there is no public constructor doing the same thing, right?
* public static CalDateTime Now => new(DateTime.Now); DateTime.Now has DateTimeKind.Local. Would using explicit floating be more clear: public static CalDateTime Now => new(LocalDateTime.FromDateTime(DateTime.Now)); or alike?

Ical.Net/DataTypes/CalDateTime.cs
axunonb: Maybe `ToZonedDateTimeWithDefault`? Else leave as is.

maknapp: Another option is to remove all `ToZonedDateTime()` and just rename `AsZonedOrDefault()` to `ToZonedDateTime()`. This would remove all ways to directly convert to another time zone. Users can make their own extension methods if they convert time zones all the time.

Will people still get confused and think `ToZonedDateTime(DateTimeZone defaultZone)` will always output a value in that time zone? Would it still need to be named something like `ToZonedDateTimeWithDefault()`?

```c#
dt.ToZonedDateTime() == dt.AsZonedOrDefault(DateTimeZone.Utc);

dt.ToZonedDateTime(dateTimeZone) == dt.AsZonedOrDefault(dateTimeZone).WithZone(dateTimeZone);

// This one is only used in tests and mostly to prevent having to change tests too much
dt.ToZonedDateTime("America/New_York") == dt.AsZonedOrDefault(tzdb["America/New_York"]).WithZone(tzdb["America/New_York"]);
```

maknapp:
>     * ZonedDateTime/Instant got factories, while CalDateTime(LocalDateTime, string? tzId) stays as a public constructor. By intension?

`CalDateTime` stores `LocalDate` and `LocalTime` values, so the `LocalDateTime` and `LocalDate` constructors felt better to me as constructors since they copy values (mostly) directly. `Instant` "converts" to UTC and `ZonedDateTime` drops the offset.

That being said, I'm on the fence about being split like this - should probably be all or nothing. All types beyond `int` constructors being factories would be explicit and easy to read, and would allow for adding more later. Using constructors with only required factory methods (`FromDateTimeDate` and `FromDateTime`) would be easier to use. Do you have an opinion on which way to go?

>     * Generally: If we have a factory, there is no public constructor doing the same thing, right?
Yes.

>     * public static CalDateTime Now => new(DateTime.Now); DateTime.Now has DateTimeKind.Local. Would using explicit floating be more clear: public static CalDateTime Now => new(LocalDateTime.FromDateTime(DateTime.Now)); or alike?
Yes.

maknapp:
`CalDateTime` should only represent a calendar date/time in UTC and not an `Instant`, so I removed the To/From for `Instant`.

I tried to remove all `ToZonedDateTime` methods and rename `AsZonedOrDefault` to `ToZonedDateTime` but I ran into two issues:
1. There are many places (mostly in tests) where it is known that there is a time zone, so being required to give a default was annoying. Having an automatic default of UTC seemed wrong, so I kept a `ToZonedDateTime()` that throws when there is no time zone.
2. I changed `AsZonedOrDefault` to `ToZonedOrDefault`. `ToZonedDateTimeWithDefault` is pretty long, and using `*OrDefault` is more similar to things like `FirstOrDefault`. This way both start with `ToZoned*` to easily see both options in autocomplete.

axunonb:
> `ToZonedDateTimeWithDefault` is pretty long, and using `*OrDefault` is more similar to things like `FirstOrDefault`. This way both start with `ToZoned*` to easily see both options in autocomplete.

Yes, the sound like a good way to go.

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -16,47 +16,42 @@
...
     public const string UtcTzId = "UTC";
 
     /// <summary>
-    /// Creates a new instance of the <see cref="CalDateTime"/> class
-    /// with the current date/time and sets the <see cref="TzId"/> to <see langword="null"/>.
+    /// Creates a <see cref="CalDateTime"/>
+    /// with the current local date/time and no time zone.
     /// </summary>
-    public static CalDateTime Now => new CalDateTime(DateTime.Now, null, true);
+    public static CalDateTime Now => new(DateTime.Now);
```
axunonb: This produces a floating `CalDateTime`, although it looks like current moment.
Should we remove `CalDateTime.Now` altogether?

Ical.Net/DataTypes/CalDateTime.cs
```diff
@@ -68,115 +63,64 @@ private CalDateTime()
...
-    /// <exception cref="System.ArgumentException">If the specified value's kind is <see cref="DateTimeKind.Local"/></exception>
-    public CalDateTime(DateTime value, bool hasTime = true) : this(
-        value,
-        value.Kind switch
+    /// Time zone will be UTC if <paramref name="tzId"/> is <see langword="null"/> and
+    /// <paramref name="value"/> kind is <see cref="DateTimeKind.Utc"/>.
+    /// </summary>
+    /// <param name="value">The value to copy the local date and time from.</param>
+    /// <param name="tzId">The time zone ID.</param>
+    public CalDateTime(DateTime value, string? tzId = null) : this(
```
axunonb: The previous version threw on `DateTimeKind.Local`. Now we silently get a floating `CalDateTime`.
If we go this way, XML doc should at minimum call this out explicitly, even if throwing is considered too strict.

Ical.Net/DataTypes/CalDateTime.cs
axunonb:
```cs
// Current ToZonedDateTime(DateTimeZone) — ALWAYS converts to the target zone
dt.ToZonedDateTime(americaNewYork); // result is always in America/New_York

// AsZonedOrDefault — only uses the zone if floating
dt.AsZonedOrDefault(americaNewYork); // result MAY NOT be in America/New_York
```
User may expect a conversion-to-zone, not a fallback-if-floating.

Ical.Net/DataTypes/CalDateTime.cs
axunonb: The breaking changes I collected so far are:

- `ToTimeZone(string?)` removed
- `AsUtc `property removed
- `Value `property removed
- `CalDateTime(DateTime, bool hasTime, ...)` constructor shape changed
- `CalDateTime(ZonedDateTime)` / `CalDateTime(Instant)` constructors replaced by factories

Ical.Net/DataTypes/CalDateTime.cs
maknapp: I changed the constructor to `FromDateTime` and changed the summary to say "Converts a DateTime of any kind to..." (similar to `NodaTime.FromDateTime` summary). Is that good?

Ical.Net/DataTypes/CalDateTime.cs
maknapp: Since users have the option of using NodaTime entirely, I think it is fine leaving this as-is for those who want to keep using DateTime.

Ical.Net/DataTypes/CalDateTime.cs
maknapp: Updated PR with breaking changes from main branch (NodaTime constructors are not on main).

maknapp:
I think this is ready to merge now. We can see how these changes feel and remove/change more later. I removed the time properties to force users to check for null themselves. While adding tests I noticed that `GetHashCode` did not match `Equals` so I changed that too.

maknapp:
@axunonb, I rebased on v6 and didn't make any other changes.

axunonb (approved):
LGTM, thanks!

PR #940, opened 2026-04-12
Add alarm poll end time

### LINKED ISSUE #939: PollAlarms(startTime, endTime) Throws EvaluationOutOfRangeException
When polling alarms for an event with an unbounded recurrence rule, `PollAlarms(startTime, endTime)` throws an `Ical.Net.Evaluation.EvaluationOutOfRangeException` exception, even when passing an `endTime`.

This appears to stem from `Alarm.Poll` evaluating all occurrences to return an `IList` rather than following the pattern of `RecurringComponent` and returning an `IEnumerable`.

**To Reproduce**

``` csharp
using System.Linq;
using Ical.Net;
using Ical.Net.DataTypes;

namespace iCalTest
{
	class Program
	{
		static void Main()
		{
			CalendarEvent e = new()
			{
				Start = new CalDateTime(2026, 04, 07),
			};
			e.RecurrenceRules.Add(new RecurrencePattern(FrequencyType.Weekly, 1));
			e.Alarms.Add(new Alarm()
			{
				Trigger = new Trigger(new Duration(days: -1))
			});

			// Generates 9 occurrences
			var eventOccurrences = e.GetOccurrences(CalDateTime.Now).TakeWhileBefore(CalDateTime.Now.AddDays(60)).ToList();

			// Throws Ical.Net.Evaluation.EvaluationOutOfRangeException
			var alarmOccurrences = e.PollAlarms(CalDateTime.Now, CalDateTime.Now.AddDays(60));
		}
	}
}
```

**Expected behavior**
`PollAlarms` returns alarm occurrences between the specified dates.

- OS: `Windows 11`
- .NET version: `.NET Framework 4.8.1`
- ical.net version: `5.2.1`

**Additional context**
Exception Message: `An out-of-range value was encountered while evaluating occurrences. This commonly happens when trying to enumerate an unbounded RRULE to its end. Consider applying the .TakeWhile() operator.`

Stack Trace:
```
   at Ical.Net.Evaluation.RecurrenceUtil.<>c__2`1.<HandleEvaluationExceptions>b__2_0(ArgumentOutOfRangeException _)
   at Ical.Net.Utility.CollectionHelpers.<Catch>d__13`2.MoveNext()
   at Ical.Net.Utility.CollectionHelpers.<Catch>d__13`2.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at Ical.Net.CalendarComponents.Alarm.GetOccurrences(IRecurringComponent rc, CalDateTime fromDate, EvaluationOptions options)
   at Ical.Net.CalendarComponents.Alarm.Poll(CalDateTime start, EvaluationOptions options)
   at Ical.Net.CalendarComponents.RecurringComponent.<>c__DisplayClass78_0.<PollAlarms>b__0(Alarm a)
   at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Ical.Net.CalendarComponents.RecurringComponent.PollAlarms(CalDateTime startTime, CalDateTime endTime)
   at iCalTest.Program.Main() in ***\iCalTest\iCalTest\Program.cs:line 26
```

maknapp:
Alarms were partially missed when switching evaluation to IEnumerable. The end time argument was removed in several places causing evalution until the end or an exception.

This adds the end time argument back in to avoid having to make significant or breaking changes.

Fixes #939

### DIFF
--- a/Ical.Net.Tests/AlarmTest.cs
+++ b/Ical.Net.Tests/AlarmTest.cs
@@ -4,94 +4,217 @@
 //
 
 #nullable enable
-using Ical.Net.CalendarComponents;
-using Ical.Net.DataTypes;
-using NUnit.Framework;
-using System.Collections.Generic;
+using System;
 using System.IO;
 using System.Linq;
-using Ical.Net.Serialization.DataTypes;
+using Ical.Net.CalendarComponents;
+using Ical.Net.DataTypes;
+using Ical.Net.Serialization;
+using NUnit.Framework;
 
 namespace Ical.Net.Tests;
 
 [TestFixture]
-public class AttendeeTest
+public class AlarmTests
 {
-    internal static CalendarEvent VEventFactory() => new CalendarEvent
+    #region Examples from RFC 5545
+
+    [Test]
+    public void ExactTimeAlarmWithRepeat()
     {
-        Summary = "Testing",
-        Start = new CalDateTime(2010, 3, 25),
-        End = new CalDateTime(2010, 3, 26)
-    };
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(1997, 3, 18)
+        };
 
-    private static readonly IList<Attendee> _attendees = new List<Attendee>
+        var valarm = """
+            BEGIN:VALARM
+            TRIGGER;VALUE=DATE-TIME:19970317T133000Z
+            REPEAT:4
+            DURATION:PT15M
+            ACTION:AUDIO
+            ATTACH;FMTTYPE=audio/basic:ftp://example.com/pub/
+                sounds/bell-01.aud
+            END:VALARM
+            """;
+
+        var alarm = SimpleDeserializer.Default
+            .Deserialize(new StringReader(valarm))
+            .Cast<Alarm>()
+            .Single();
+
+        e.Alarms.Add(alarm);
+
+        var results = e.PollAlarms(new CalDateTime(1997, 3, 10), null)
+            .Select(x => x.DateTime)
+            .ToList();
+
+        var expectedAlarms = new[]
+        {
+            new CalDateTime(new DateTime(1997, 3, 17, 13, 30, 0, DateTimeKind.Utc)),
+            new CalDateTime(new DateTime(1997, 3, 17, 13, 45, 0, DateTimeKind.Utc)),
+            new CalDateTime(new DateTime(1997, 3, 17, 14, 0, 0, DateTimeKind.Utc)),
+            new CalDateTime(new DateTime(1997, 3, 17, 14, 15, 0, DateTimeKind.Utc)),
+            new CalDateTime(new DateTime(1997, 3, 17, 14, 30, 0, DateTimeKind.Utc)),
+        };
+
+        Assert.That(results, Is.EquivalentTo(expectedAlarms));
+    }
+
+    [Test]
+    public void RelativeAlarmWithRepeat()
     {
-        new Attendee("MAILTO:james@example.com")
+        CalendarEvent e = new()
         {
-            CommonName = "James James",
-            Role = ParticipationRole.RequiredParticipant,
-            Rsvp = true,
-            ParticipationStatus = EventParticipationStatus.Tentative
-        },
-        new Attendee("MAILTO:mary@example.com")
+            Start = new CalDateTime(1997, 3, 18, 8, 30, 0, "America/New_York")
+        };
+
+        var valarm = """
+            BEGIN:VALARM
+            TRIGGER:-PT30M
+            REPEAT:2
+            DURATION:PT15M
+            ACTION:DISPLAY
+            DESCRIPTION:Breakfast meeting with executive
+                team at 8:30 AM EST.
+            END:VALARM
+            """;
+
+        var alarm = SimpleDeserializer.Default
+            .Deserialize(new StringReader(valarm))
+            .Cast<Alarm>()
+            .Single();
+
+        e.Alarms.Add(alarm);
+
+        var results = e.PollAlarms(new CalDateTime(1997, 3, 18), null)
+            .Select(x => x.DateTime)
+            .ToList();
+
+        var expectedAlarms = new[]
         {
-            CommonName = "Mary Mary",
-            Role = ParticipationRole.RequiredParticipant,
-            Rsvp = true,
-            ParticipationStatus = EventParticipationStatus.Accepted
-        }
-    }.AsReadOnly();
+            new CalDateTime(1997, 3, 18, 8, 0, 0, "America/New_York"),
+            new CalDateTime(1997, 3, 18, 8, 15, 0, "America/New_York"),
+            new CalDateTime(1997, 3, 18, 8, 30, 0, "America/New_York"),
+        };
 
+        Assert.That(results, Is.EquivalentTo(expectedAlarms));
+    }
 
-    /// <summary>
-    /// Ensures that attendees can be properly added to an event.
-    /// </summary>
-    [Test, Category("Attendee")]
-    public void Add1Attendee()
+    [Test]
+    public void RelativeAlarmDaysBefore()
     {
-        var evt = VEventFactory();
-        Assert.That(evt.Attendees.Count, Is.EqualTo(0));
-
-        evt.Attendees.Add(_attendees[0]);
-        Assert.That(evt.Attendees, Has.Count.EqualTo(1));
-
-        Assert.Multiple(() =>
+        Todo todo = new()
         {
-            //the properties below had been set to null during the Attendees.Add operation in NuGet version 2.1.4
-            Assert.That(evt.Attendees[0].Role, Is.EqualTo(ParticipationRole.RequiredParticipant));
-            Assert.That(evt.Attendees[0].ParticipationStatus, Is.EqualTo(EventParticipationStatus.Tentative));
+            Start = new CalDateTime(1997, 3, 18, 7, 30, 0, "America/New_York"),
+            Due = new CalDateTime(1997, 3, 18, 8, 30, 0, "America/New_York"),
+        };
+
+        var valarm = """
+            BEGIN:VALARM
+            TRIGGER;RELATED=END:-P2D
+            ACTION:EMAIL
+            ATTENDEE:mailto:john_doe@example.com
+            SUMMARY:*** REMINDER: SEND AGENDA FOR WEEKLY STAFF MEETING ***
+            DESCRIPTION:A draft agenda needs to be sent out to the attendees
+              to the weekly managers meeting (MGR-LIST). Attached is a
+              pointer the document template for the agenda file.
+            ATTACH;FMTTYPE=application/msword:http://example.com/
+             templates/agenda.doc
+            END:VALARM
+            """;
+
+        var alarm = SimpleDeserializer.Default
+            .Deserialize(new StringReader(valarm))
+            .Cast<Alarm>()
+            .Single();
+
+        todo.Alarms.Add(alarm);
+
+        var results = todo.PollAlarms(new CalDateTime(1997, 3, 10), new CalDateTime(1997, 3, 20))
+            .Select(x => x.DateTime)
+            .ToList();
+
+        var expectedAlarms = new[]
+        {
+            new CalDateTime(1997, 3, 16, 8, 30, 0, "America/New_York"),
+        };
+
+        Assert.That(results, Is.EquivalentTo(expectedAlarms));
+    }
+
+    #endregion
+
+
+    [Test]
+    public void AlarmWithExactTime()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(2026, 4, 7)
+        };
+
+        e.Alarms.Add(new Alarm()
+        {
+            Trigger = new Trigger
+            {
+                DateTime = new CalDateTime(new DateTime(2026, 4, 5, 0, 0, 0, DateTimeKind.Utc))
+            }
         });
+
+        var alarmOccurrences = e.PollAlarms(e.Start, e.Start.AddDays(1))
+            .Select(x => x.DateTime!)
+            .ToList();
+
+        var expectedAlarms = new[]
+        {
+            new CalDateTime(new DateTime(2026, 4, 5, 0, 0, 0, DateTimeKind.Utc)),
+        };
+
+        Assert.That(alarmOccurrences, Is.EquivalentTo(expectedAlarms));
     }
 
-    [Test, Category("Attendee")]
-    public void Add2Attendees()
+    [Test]
+    public void RecurringAlarm()
     {
-        var evt = VEventFactory();
-        Assert.That(evt.Attendees.Count, Is.EqualTo(0));
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(2026, 4, 7),
+            RecurrenceRule = new RecurrencePattern(FrequencyType.Weekly, 1)
+        };
 
-        evt.Attendees.Add(_attendees[0]);
-        evt.Attendees.Add(_attendees[1]);
-        Assert.That(evt.Attendees, Has.Count.EqualTo(2));
-        Assert.That(evt.Attendees[1].Role, Is.EqualTo(ParticipationRole.RequiredParticipant));
+        e.Alarms.Add(new Alarm()
+        {
+            Trigger = new Trigger(new Duration(days: -1))
+        });
+
+        var alarmOccurrences = e.PollAlarms(e.Start, e.Start.AddDays(21))
+            .Select(x => x.DateTime!)
+            .ToList();
+
+        var expectedAlarms = new[]
+        {
+            new CalDateTime(2026, 4, 6),
+            new CalDateTime(2026, 4, 13),
+            new CalDateTime(2026, 4, 20),
+        };
+
+        Assert.That(alarmOccurrences, Is.EquivalentTo(expectedAlarms));
     }
 
-    /// <summary>
-    /// Ensures that attendees can be properly removed from an event.
-    /// </summary>
-    [Test, Category("Attendee")]
-    public void Remove1Attendee()
+    [Test]
+    public void AlarmWithoutParentIsEmpty()
     {
-        var evt = VEventFactory();
-        Assert.That(evt.Attendees.Count, Is.EqualTo(0));
+        var alarm = new Alarm()
+        {
+            Trigger = new()
+            {
+                DateTime = new CalDateTime(new DateTime(2026, 4, 5, 0, 0, 0, DateTimeKind.Utc))
+            }
+        };
 
-        var attendee = _attendees.First();
-        evt.Attendees.Add(attendee);
-        Assert.That(evt.Attendees, Has.Count.EqualTo(1));
+        var occurrences = alarm.Poll(null, null);
 
-        evt.Attendees.Remove(attendee);
-        Assert.That(evt.Attendees.Count, Is.EqualTo(0));
-
-        evt.Attendees.Remove(_attendees.Last());
-        Assert.That(evt.Attendees.Count, Is.EqualTo(0));
+        Assert.That(occurrences, Is.Empty);
     }
 }
--- a/Ical.Net/CalendarComponents/Alarm.cs
+++ b/Ical.Net/CalendarComponents/Alarm.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // Copyright ical.net project maintainers and contributors.
 // Licensed under the MIT license.
 //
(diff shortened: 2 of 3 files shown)


maknapp:
1. What is the purpose of `Alarm.Poll()`? I do not use alarms, but if I did I would want to get the "next" alarm in the calendar after "now" but that is not what this does. The start/end times only filter on event/todo occurrence and then gets all alarms for those items. Filtering by actual alarm time is more complicated.
2. Why is it called "Poll" instead of just "GetOccurrences" like everything else? "PollAlarms" could be "GetAlarmOccurrences".

axunonb:
I'm not deep in `Alarm` either.

The mentioned methods work together in a call hierarchy:

```
PollAlarms()  ← public API on RecurringComponent / IAlarmContainer
    └── Poll()    ← per-Alarm entry point
            └── GetOccurrences()  ← does the actual work
```

So `Poll` is the **self-contained entry point** on a single `Alarm` instance. It:

1. Resolves the **parent** `RecurringComponent` via `Parent` — since an `Alarm` is always a child of a component like a `VEvent` or `Todo`.
2. Guards against the case where `Parent` is not a `RecurringComponent` (returns empty).
3. Delegates all the real work to `GetOccurrences`, passing the resolved parent along.

It exists so callers don't need to know or pass the parent component manually.

External code will most likely call 
`PollAlarms(DateTimeZone, Instant?, Instant?)` on the `RecurringComponent` (e.g., a `VEvent`), 
instead of calling
`Poll(DateTimeZone, Instant?, EvaluationOptions?)` or `GetOccurrences(IRecurringComponent, DateTimeZone, Instant?, EvaluationOptions?)` directly:, e.g.:

```cs
// RecurringComponent.cs
public virtual IList<AlarmOccurrence> PollAlarms(DateTimeZone timeZone, Instant? startTime, Instant? endTime)
    => Alarms.SelectMany(a => a.Poll(timeZone, startTime)
                               .TakeWhile(p => endTime == null || p.Start.ToInstant() < endTime))
             .ToList();
```
Does this make sense to you?

Ical.Net/CalendarComponents/Alarm.cs
```diff
@@ -93,30 +105,20 @@ public virtual IList<AlarmOccurrence> GetOccurrences(IRecurringComponent rc, Cal
                 fromDate = rc.Start?.Copy();
             }
 
-            Duration? duration = null;
-            foreach (var o in rc.GetOccurrences(fromDate, options))
+            var componentOccurrences = rc.GetOccurrences(fromDate, options);
+            if (endDate != null)
+            {
+                componentOccurrences = componentOccurrences.TakeWhileBefore(endDate);
```
axunonb, line 111: `TakeWhileBefore(endDate)` filters on component start times. For any relative trigger, an alarm fires at `componentStart + triggerDuration`, not at `componentStart`. The correct upper bound for the component window is endDate - triggerDuration.
```
Weekly event: Apr 7, Apr 14, Apr 21, Apr 28, etc.
Trigger:      -P1D  (triggerDuration = -1 day)
Alarms fire:  Apr 6, Apr 13, Apr 20, Apr 27, etc.
endDate:      Apr 28
```
So shouldn't it rather read
`componentOccurrences = componentOccurrences.TakeWhileBefore(endDate.Add(-Trigger.Duration!.Value));` ?

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -220,5 +220,5 @@ public virtual IEnumerable<Occurrence> GetOccurrences(CalDateTime? startTime = n
     public virtual IList<AlarmOccurrence> PollAlarms() => PollAlarms(null, null);
 
     public virtual IList<AlarmOccurrence> PollAlarms(CalDateTime? startTime, CalDateTime? endTime)
-        => Alarms.SelectMany(a => a.Poll(startTime).TakeWhile(p => (endTime == null) || (p.Period?.StartTime < endTime))).ToList();
+        => Alarms.SelectMany(a => a.Poll(startTime, endTime).TakeWhile(p => (endTime == null) || (p.Period?.StartTime < endTime))).ToList();
```
axunonb, line 223: `AddRepeatedItems(List<AlarmOccurrence>)` appends repetitions after base occurrences. For a recurring event with REPEAT:N, the list order is:
`[base₁, base₂, base₃, base₁+Δ, base₁+2Δ, base₂+Δ, base₂+2Δ, etc.]`
`base₁+Δ` may be greater than base₂, so `TakeWhile` expects sorted elements and stops early, dropping valid alarm occurrences. **So I'd vote for `where` here**.

Even with the change: don't  we get redundant double-filtering?

Ical.Net.Tests/AlarmTest.cs
```diff
@@ -4,94 +4,217 @@
...
 
-        var attendee = _attendees.First();
-        evt.Attendees.Add(attendee);
-        Assert.That(evt.Attendees, Has.Count.EqualTo(1));
+        var expectedAlarms = new[]
+        {
+            new CalDateTime(2026, 4, 6),
+            new CalDateTime(2026, 4, 13),
+            new CalDateTime(2026, 4, 20),
+        };
```
axunonb, line 200: With the changes mentioned in `Alarm` and `RecurringComponent`
`new CalDateTime(2026, 4, 27)`
must be added to `expectedAlarms`, while CalDateTime(2026, 4, 6) must be removed

Ical.Net.Tests/AlarmTest.cs
```diff
@@ -4,94 +4,217 @@
...
 
-        evt.Attendees.Remove(attendee);
-        Assert.That(evt.Attendees.Count, Is.EqualTo(0));
+        var occurrences = alarm.Poll(null, null);
 
-        evt.Attendees.Remove(_attendees.Last());
-        Assert.That(evt.Attendees.Count, Is.EqualTo(0));
+        Assert.That(occurrences, Is.Empty);
     }
 }
```
axunonb, line 220: [Edit: add more unit tests]

```cs
[Test]
public void RecurringAlarm_NegativeTrigger_BoundaryOccurrenceShouldBeIncluded()
{
    // Verifies that a component occurrence whose alarm fire time falls within the
    // polling window [startTime, endTime) is correctly included even when the
    // component's own DTSTART coincides with endTime (boundary case).
    //
    // Weekly occurrences: Apr 7, Apr 14, Apr 21, Apr 28, ...
    // Alarms (TRIGGER:-P1D):     Apr 6, Apr 13, Apr 20, Apr 27, ...
    //
    // The Apr 28 occurrence fires its alarm on Apr 27 — within [Apr 7 09:00, Apr 28 09:00).
    // The Apr 6 alarm fires before startTime (Apr 7 09:00) and is excluded per RFC 5545.

    CalendarEvent e = new()
    {
        Start = new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
        RecurrenceRule = new RecurrencePattern(FrequencyType.Weekly, 1)
    };

    e.Alarms.Add(new Alarm
    {
        Trigger = new Trigger(new Duration(days: -1))
    });

    var results = e.PollAlarms(
            new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
            new CalDateTime(2026, 4, 28, 9, 0, 0, "UTC"))
        .Select(x => x.DateTime!)
        .OrderBy(x => x)
        .ToList();

    Assert.That(results, Is.EqualTo(new[]
    {
        new CalDateTime(2026, 4, 13, 9, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 20, 9, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 27, 9, 0, 0, "UTC"),
    }));
}

[Test]
public void RecurringAlarm_WithRepeat_EndTimeCutoffDoesNotDropEarlierRepetitions()
{
    // Verifies that alarm repetitions for an earlier component occurrence are included
    // even when a later occurrence's base alarm falls at or after endTime.
    // The streaming implementation (OrderedNestedMergeMany) produces a globally
    // time-sorted stream, so TakeWhile correctly stops only when fire times exceed endTime.

    CalendarEvent e = new()
    {
        Start = new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
        RecurrenceRule = new RecurrencePattern(FrequencyType.Weekly, 1)
    };

    // TRIGGER:+PT1H, REPEAT:2, DURATION:PT1H
    // Apr  7 09:00 occurrence → alarms: Apr  7 10:00, 11:00, 12:00
    // Apr 14 09:00 occurrence → alarms: Apr 14 10:00, 11:00, 12:00
    e.Alarms.Add(new Alarm
    {
        Trigger = new Trigger(new Duration(hours: 1)),
        Repeat = 2,
        Duration = new Duration(hours: 1)
    });

    // endTime = Apr 14 09:30. The sorted stream is:
    //   [Apr 7 10:00, Apr 7 11:00, Apr 7 12:00, Apr 14 10:00, ...]
    // TakeWhile(< Apr 14 09:30) includes all three Apr 7 alarms and stops at Apr 14 10:00.
    var results = e.PollAlarms(
            new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
            new CalDateTime(2026, 4, 14, 9, 30, 0, "UTC"))
        .Select(x => x.DateTime!)
        .OrderBy(x => x)
        .ToList();

    Assert.That(results, Is.EqualTo(new[]
    {
        new CalDateTime(2026, 4, 7, 10, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 7, 11, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 7, 12, 0, 0, "UTC"),
    }));
}

[Test]
public void RecurringAlarm_WithRepeat_LongRepDuration_DoesNotDropLaterOccurrenceRepetitions()
{
    CalendarEvent e = new()
    {
        Start = new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
        RecurrenceRule = new RecurrencePattern(FrequencyType.Weekly, 1)
    };

    // TRIGGER:PT0S (fires at component start), REPEAT:2, DURATION:P10D
    // Apr  7 occurrence => alarms: Apr  7, Apr 17, Apr 27
    // Apr 14 occurrence => alarms: Apr 14, Apr 24, May  4
    // Apr 21 occurrence => alarms: Apr 21, May  1, May 11
    // The streaming implementation merges these into a single globally time-sorted
    // sequence; TakeWhile(< Apr 25) correctly includes Apr 7, 14, 17, 21, 24.
    e.Alarms.Add(new Alarm
    {
        Trigger = new Trigger(Duration.Zero),
        Repeat = 2,
        Duration = new Duration(days: 10)
    });

    var results = e.PollAlarms(
            new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
            new CalDateTime(2026, 4, 25, 9, 0, 0, "UTC"))
        .Select(x => x.DateTime!)
        .OrderBy(x => x)
        .ToList();

    Assert.That(results, Is.EqualTo(new[]
    {
        new CalDateTime(2026, 4,  7, 9, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 14, 9, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 17, 9, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 21, 9, 0, 0, "UTC"),
        new CalDateTime(2026, 4, 24, 9, 0, 0, "UTC"),
    }));
}

[Test]
public void RecurringAlarm_RelatedEnd_ReturnsAlarmsRelativeToOccurrenceEnd()
{
    // RELATED=END triggers fire relative to the END of each occurrence, not the start.
    // GetOccurrences uses (endDate - triggerDuration) as the component window, which is
    // only an approximation for RELATED=END (exact for RELATED=START).
    // This test verifies alarms are computed from EffectiveEndTime, not StartTime.

    CalendarEvent e = new()
    {
        Start = new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
        Duration = new Duration(hours: 1), // each occurrence ends at 10:00 UTC
        RecurrenceRule = new RecurrencePattern(FrequencyType.Weekly, 1)
    };

    // Alarm fires 30 min before each occurrence END → each occurrence end - PT30M
    // Apr  7: end=10:00 => alarm  9:30
    // Apr 14: end=10:00 => alarm  9:30
    // Apr 21: end=10:00 => alarm  9:30
    e.Alarms.Add(new Alarm
    {
        Trigger = new Trigger(new Duration(minutes: -30))
        {
            Related = TriggerRelation.End
        }
    });

    var results = e.PollAlarms(
            new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"),
            new CalDateTime(2026, 4, 22, 0, 0, 0, "UTC"))
        .Select(x => x.DateTime!)
        .OrderBy(x => x)
        .ToList();

    Assert.That(results, Is.EqualTo(new[]
    {
        new CalDateTime(2026, 4,  7, 9, 30, 0, "UTC"),
        new CalDateTime(2026, 4, 14, 9, 30, 0, "UTC"),
        new CalDateTime(2026, 4, 21, 9, 30, 0, "UTC"),
    }));
}

[Test]
public void AlarmWithRepeatButNoDuration_ShouldNotProduceDuplicates()
{
    // RFC 5545 section 3.8.6.2: REPEAT and DURATION must appear together.
    // AddRepeatedItems skips the Add() call when Duration is null, but still executes
    // the outer loop, appending `Repeat` identical copies of the base alarm time.
    // This test documents the expected behavior: no duplicates should be produced.

    CalendarEvent e = new()
    {
        Start = new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC")
    };

    e.Alarms.Add(new Alarm
    {
        Trigger = new Trigger(new Duration(minutes: -15)),
        Repeat = 3
        // Duration intentionally omitted
    });

    var results = e.PollAlarms(null, null)
        .Select(x => x.DateTime!)
        .ToList();

    // Currently returns 4 alarms (base + 3 duplicates); should return exactly 1.
    Assert.That(results, Has.Count.EqualTo(1));
    Assert.That(results[0], Is.EqualTo(new CalDateTime(2026, 4, 7, 8, 45, 0, "UTC")));
}

[Test]
public void PollAlarms_StartTime_FiltersAlarmFireTimes()
{
    // Per RFC 5545, PollAlarms(startTime, endTime) treats startTime as a lower bound
    // on alarm FIRE TIMES. An alarm that fires before startTime is excluded, even when
    // its component occurrence starts at or after startTime.

    CalendarEvent e = new()
    {
        Start = new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC")
    };

    e.Alarms.Add(new Alarm
    {
        Trigger = new Trigger(new Duration(days: -1)) // fires Apr 6 for Apr 7 occurrence
    });

    var results = e.PollAlarms(
            new CalDateTime(2026, 4, 7, 9, 0, 0, "UTC"), // startTime = Apr 7
            new CalDateTime(2026, 4, 8, 9, 0, 0, "UTC"))
        .Select(x => x.DateTime!)
        .ToList();

    // The Apr 6 alarm fires before startTime (Apr 7 09:00) => correctly excluded.
    Assert.That(results, Is.Empty);
}
```

axunonb (commented):
Not sure whether the comments are fully correct, they're just what I figured out.

Ical.Net/CalendarComponents/Alarm.cs
```diff
@@ -93,30 +105,20 @@ public virtual IList<AlarmOccurrence> GetOccurrences(IRecurringComponent rc, Cal
...
-                    {
-                        dt = o.Period.StartTime.Add(duration.Value);
-                    }
-                    else
-                    {
-                        throw new ArgumentException(
-                            "Alarm trigger is relative to the START of the occurrence; however, the occurence has no discernible end.");
-                    }
+                    dt = o.Period.EffectiveEndTime ?? throw new ArgumentException(
+                        "Alarm trigger is relative to the START of the occurrence; however, the occurence has no discernible end.");
```
axunonb: Shouldn't it read: `Alarm trigger is relative to the END of the occurrence; however, the occurrence has no discernible end.` for `TriggerRelation.End`?

Ical.Net/CalendarComponents/Alarm.cs
axunonb, line 183:
```cs
// RFC 5545 section 3.8.6.2: REPEAT and DURATION must appear together.
// If either is absent there are no repetitions to generate.
// Note: Properties.Get<Duration> returns default(Duration) for an unset property,
// which is non-null, so we check property existence rather than the value.
// This should be checked:
if (Repeat <= 0 || !Properties.ContainsKey("DURATION"))
{
    return;
}
```

axunonb:
@minichma It would be helpful if you had a look at this PR, too. I'm not feeling too confident about the changes I suggested.

minichma:
> The end time argument was removed in several places causing evalution until the end or an exception.

Yes, this was certainly a mistake. I also don't know much about Alarms, when switching to IEnumerable I basically just wanted to keep the previous behaviour.
I guess I started with switching to IEnumerable but keeping the endTime at first. I stopped in the Alarm class as it was not exactly straight forward to switch to IEnumerable and seemingly not necessary. But this obviously wasn't true anymore after removing the endDate.

I'll try to have a closer look hopefully tomorrow evening or on Wednesday at the latest.

minichma:
I was playing a little with the code this morning and drafted an approach for also iterating the alarms in a streaming manner, which I think would be nice. See here:

https://github.com/minichma/ical.net/tree/xp/alarm_streaming

Its just about the streaming but doesn't implement the other improvements included in this PR. Ran Mark's tests from this PR and one fails, but that's probably not related to the streaming.

What do you guys think?

maknapp:
I have only skimmed the above comments really quickly. This PR was mostly focused on preventing the unexpected exception without changing (bad) alarm evaluation. The tests were only to document current behavior and add test coverage since there was 0%. Some of the tests are likely wrong because handling relative triggers appears to be incorrect.

I was not sure how much we wanted to fix in v5. Any changes towards making alarm IEnumerable should only be done in v6 since it is a breaking change. We could ignore any further alarm fixes for v5 unless they are easy copy from v6.

axunonb:
> What do you guys think?

Wow, that looks pretty cool, and it's a nice step forward having the streaming pattern also for `Alarm`s.
For simplicity, bringing Mark's code, my proposals and Markus' streaming implementation (on top of main) together one by one:
https://github.com/axunonb/ical.net/tree/pr/fix-alarm-eval-range-modified
~~@maknapp Feel free to merge into your PR.~~ Didn't read your [comment](https://github.com/ical-org/ical.net/pull/940#issuecomment-4243872295) before mine.
All tests pass.

Still, my changes and unit tests require a review in terms of RFC5545 compliance.

axunonb:
> We could ignore any further alarm fixes for v5 unless they are easy copy from v6.

Yes, @minichma's code should rather go into version/6.0 because of the breaking changes.
For further fixes we can cherry pick the simple ones to go into v5, too.

minichma:
Very nice!

> Still, my changes and unit tests require a review in terms of RFC5545 compliance.

I'll have a look, don't know much about Alarms though. Could you include it in a PR  (this one or a new one?) so we can comment?

minichma:
There's one major issue with my draft that will be tricky to solve. The code builds on the assumption that GetOccurrences() returns an ordered sequence. As the DURATION added would be the same for every recurrence, the resulting alarm recurrences would still be ordered. Unfortunately this the assumption doesn't necessarily hold true, namely in case of `RELATED=END`, because the original sequence is ordered by the start value, not the end value (which may make a difference, e.g. when using something like `RDATE;VALUE=PERIOD:19960403/P5D,19960404/P1D`. 

Properly dealing with this difference would require generating the original sequence with configurable ordering, which would affect quite some part of the evaluation logic. It would probably require explicitly passing an IComparer through the evaluation functionality or include a corresponding param in the EvaluationOptions. It would also imply that there is no 'correct' or 'natural' order for (e.g.) `Period`s (I feel we discussed this before), so they should not implement `IComparable<T>` anymore. Instead they should provide a number of Comparers (e.g. as static properties) that could be used for different purposes. 

Sounds like quite some work just to get this seemingly tiny alarm feature working. On the other hand, I feel moving the comparison logic out of the core entity classes would be a good idea anyways. And making the comparer configurable would be nice as well, so maybe worth it.

axunonb:
> Could you include it in a PR (this one or a new one?) so we can comment?

I just created a PR on my fork in order review and comment. We probably won't merge it in to main.
https://github.com/axunonb/ical.net/pull/2

maknapp:
Event/todo occurrences are being evaluated multiple times if there are multiple alarms attached. Why even have `GetOccurrences(...)` and `Poll(...)` on an `Alarm` when it produces no alarm occurrences by itself? Does it make sense to _only_ have `RecurringComponent.GetAlarmOccurrences(...)` (renamed from `PollAlarms`) that evaluates the event/todo first and then looks at child alarms?

VALARM can only belong to VEVENT or VTODO. Currently `Journal` has `PollAlarms(...)` because it extends `RecurringComponent`.

axunonb:
> Why even have `GetOccurrences(...)` and `Poll(...)` on an `Alarm`

`RecurringComponent.GetAlarmOccurrences(...)` Excellent! 
Every call runs the `RecurrencePatternEvaluator` pipeline from scratch. A component with 3 alarms reprocesses its entire `RRULE` chain 3 times.  `Alarm` is not a unit of evaluation - it's a mapping function.
The `FIXME` comment in the source already acknowledges that this should be improved.

[I committed a POC here](https://github.com/axunonb/ical.net/pull/2/commits/a9829cb6e6514b63d5e8871ae9fd27f46c5e9ad0)

axunonb (dismissed):
Let's merge this hot fix as suggested and publish a new minor release.
Move any improvements that will cause breaking changes to v6.

maknapp:
I updated the exception message mentioned above since that is not a breaking change.

axunonb (approved):

axunonb:
Merge this PR and publish a bug fix release?

maknapp:
Sounds good to me.

Issue #903, opened 2026-01-08
Recurrence rule validation


maknapp:
For the various `EvaluationException` instances related to rrule issues, would it make sense to validate rrules before evaluation? Evaluating a Calendar would throw if just one event was invalid. If rrules were validated on creation, it could be ignored before being added to the calendar. Can this be done without breaking serialize/deserialize backwards compatibility?

axunonb:
On a first glance, `RRULE` validation before evaluation is appealing: Evaluation failures that stem from invalid recurrence semantics could be handled early.
But should we evaluate leniently or strictly? `RULE:FREQ=MONTHLY;BYWEEKNO=15` is invalid as of RFC5545. Should evaluation fail here? Currently we just return no occurrences.

We could introduce an explicit validation step that clients can **opt into** before adding items to a calendar, like `TryValidateRecurrencePattern`. Provide constructors/overloads or a `ValidationOptions` flag so that libraries that need strict RFC compliance can influence validation. This will add complexity.

With optional validation, however, we still need to surface the existing `EvaluationException`s — validation is just a complementary step.

byteorbix:
I opened this [issue on rrule-temporal](https://github.com/ggaabe/rrule-temporal/issues/95) re strict vs lenient validation. They now added an option for strict validation, while the default is lenient.

axunonb:
See also #734

PR #942, opened 2026-04-16
feat(alarms)!: Introduce streaming for alarm occurrences

### LINKED ISSUE #943: Introduce streaming for alarm occurrences
Introduce `GetAlarmOccurrences` for streaming, ordered alarm fire times with optional time bounds, replacing current `PollAlarms` and `Poll` in v6.

axunonb:
Changes:
- Introduce GetAlarmOccurrences for streaming, ordered alarm fire times with optional time bounds, replacing PollAlarms/Poll.
- Refactor Alarm to use GetFireTimes for RFC 5545-compliant repetition logic.
- Merge and order alarm occurrences across all alarms, handling RELATED=END and RDATE periods correctly.
- Add OrderedNestedMergeMany for efficient streaming merge of ordered sequences.
- Make AlarmOccurrence comparable by fire time.
- Add more AlarmTests

BREAKING CHANGE: `IAlarmContainer.PollAlarms(DateTimeZone, Instant?, Instant?)` has been removed. Replace calls with `GetAlarmOccurrences(timeZone, startTime, endTime, options)`. The new method returns `IEnumerable<AlarmOccurrence>` ordered by fire time, accepts an optional `EvaluationOptions` argument, and correctly handles `RELATED=END` triggers on PERIOD RDATEs with varying durations.

Credits to @minichma for providing the code for streaming `Alarm` occurrences (https://github.com/minichma/ical.net/tree/xp/alarm_streaming)

Resolves #943

### DIFF
--- a/Ical.Net.Tests/AlarmTests.cs
+++ b/Ical.Net.Tests/AlarmTests.cs
@@ -311,6 +311,37 @@ public class AlarmTests
     }
 
     [Test]
+    public void RecurringAlarm_RelatedEnd_Indefinite()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(2026, 4, 7, 9, 0, 0, CalDateTime.UtcTzId),
+            Duration = new Duration(hours: 1),
+            RecurrenceRule = new RecurrenceRule(FrequencyType.Daily)
+        };
+
+        e.Alarms.Add(new Alarm
+        {
+            Trigger = new Trigger(new Duration(minutes: -30))
+            {
+                Related = TriggerRelation.End
+            }
+        });
+
+        var results = e.GetAlarmOccurrences(DateTimeZone.Utc)
+            .Select(x => x.Start.ToInstant())
+            .Take(3)
+            .ToList();
+
+        Assert.That(results, Is.EqualTo(new[]
+        {
+            Instant.FromUtc(2026, 4,  7, 9, 30, 0),
+            Instant.FromUtc(2026, 4, 14, 9, 30, 0),
+            Instant.FromUtc(2026, 4, 21, 9, 30, 0),
+        }));
+    }
+
+    [Test]
     public void AlarmWithRepeatButNoDuration_ShouldNotProduceDuplicates()
     {
         // RFC 5545 §3.8.6.2: REPEAT and DURATION must appear together.
--- a/Ical.Net.Tests/AlarmTests.cs
+++ b/Ical.Net.Tests/AlarmTests.cs
@@ -1,0 +1,421 @@
+//
+// Copyright ical.net project maintainers and contributors.
+// Licensed under the MIT license.
+//
+
+using System.IO;
+using System.Linq;
+using Ical.Net.CalendarComponents;
+using Ical.Net.DataTypes;
+using Ical.Net.Serialization;
+using NodaTime;
+using NUnit.Framework;
+using Duration = Ical.Net.DataTypes.Duration;
+using Period = Ical.Net.DataTypes.Period;
+
+namespace Ical.Net.Tests;
+
+[TestFixture]
+public class AlarmTests
+{
+    #region Examples from RFC 5545
+
+    [Test]
+    public void ExactTimeAlarmWithRepeat()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(1997, 3, 18)
+        };
+
+        var valarm = """
+            BEGIN:VALARM
+            TRIGGER;VALUE=DATE-TIME:19970317T133000Z
+            REPEAT:4
+            DURATION:PT15M
+            ACTION:AUDIO
+            ATTACH;FMTTYPE=audio/basic:ftp://example.com/pub/
+                sounds/bell-01.aud
+            END:VALARM
+            """;
+
+        var alarm = SimpleDeserializer.Default
+            .Deserialize(new StringReader(valarm))
+            .Cast<Alarm>()
+            .Single();
+
+        e.Alarms.Add(alarm);
+
+        var results = e.GetAlarmOccurrences(DateTimeZone.Utc)
+            .Select(x => x.Start.ToInstant())
+            .ToList();
+
+        Assert.That(results, Is.EqualTo(new[]
+        {
+            Instant.FromUtc(1997, 3, 17, 13, 30, 0),
+            Instant.FromUtc(1997, 3, 17, 13, 45, 0),
+            Instant.FromUtc(1997, 3, 17, 14,  0, 0),
+            Instant.FromUtc(1997, 3, 17, 14, 15, 0),
+            Instant.FromUtc(1997, 3, 17, 14, 30, 0),
+        }));
+    }
+
+    [Test]
+    public void RelativeAlarmBeforeStart()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(1997, 3, 17, 13, 30, 0, CalDateTime.UtcTzId),
+            Duration = new Duration(hours: 1)
+        };
+
+        var valarm = """
+            BEGIN:VALARM
+            TRIGGER:-PT30M
+            REPEAT:2
+            DURATION:PT15M
+            ACTION:AUDIO
+            ATTACH;FMTTYPE=audio/basic:ftp://example.com/pub/
+                sounds/bell-01.aud
+            END:VALARM
+            """;
+
+        var alarm = SimpleDeserializer.Default
+            .Deserialize(new StringReader(valarm))
+            .Cast<Alarm>()
+            .Single();
+
+        e.Alarms.Add(alarm);
+
+        var results = e.GetAlarmOccurrences(DateTimeZone.Utc)
+            .Select(x => x.Start.ToInstant())
+            .ToList();
+
+        Assert.That(results, Is.EqualTo(new[]
+        {
+            Instant.FromUtc(1997, 3, 17, 13,  0, 0),
+            Instant.FromUtc(1997, 3, 17, 13, 15, 0),
+            Instant.FromUtc(1997, 3, 17, 13, 30, 0),
+        }));
+    }
+
+    [Test]
+    public void RelativeAlarmAfterEnd()
+    {
+        CalendarEvent e = new()
+        {
+            Start    = new CalDateTime(1997, 3, 17, 13, 30, 0, CalDateTime.UtcTzId),
+            Duration = new Duration(hours: 1)
+        };
+
+        var valarm = """
+            BEGIN:VALARM
+            TRIGGER;RELATED=END:PT5M
+            REPEAT:2
+            DURATION:PT5M
+            ACTION:AUDIO
+            ATTACH;FMTTYPE=audio/basic:ftp://example.com/pub/
+                sounds/bell-01.aud
+            END:VALARM
+            """;
+
+        var alarm = SimpleDeserializer.Default
+            .Deserialize(new StringReader(valarm))
+            .Cast<Alarm>()
+            .Single();
+
+        e.Alarms.Add(alarm);
+
+        var results = e.GetAlarmOccurrences(DateTimeZone.Utc)
+            .Select(x => x.Start.ToInstant())
+            .ToList();
+
+        Assert.That(results, Is.EqualTo(new[]
+        {
+            Instant.FromUtc(1997, 3, 17, 14, 35, 0),
+            Instant.FromUtc(1997, 3, 17, 14, 40, 0),
+            Instant.FromUtc(1997, 3, 17, 14, 45, 0),
+        }));
+    }
+
+    [Test]
+    public void EmailAlarmBeforeEnd()
+    {
+        var valarm = """
+            BEGIN:VALARM
+            TRIGGER;RELATED=END:-P2D
+            ACTION:EMAIL
+            END:VALARM
+            """;
+
+        var alarm = SimpleDeserializer.Default
+            .Deserialize(new StringReader(valarm))
+            .Cast<Alarm>()
+            .Single();
+
+        CalendarEvent e = new()
+        {
+            Start    = new CalDateTime(1998, 1,  1, 0, 0, 0, CalDateTime.UtcTzId),
+            Duration = new Duration(days: 5)
+        };
+        e.Alarms.Add(alarm);
+
+        var results = e.GetAlarmOccurrences(DateTimeZone.Utc)
+            .Select(x => x.Start.ToInstant())
+            .ToList();
+
+        // end = Jan 6, trigger = end - 2 days = Jan 4
+        Assert.That(results, Is.EqualTo(new[]
+        {
+            Instant.FromUtc(1998, 1, 4, 0, 0, 0)
+        }));
+    }
+
+    #endregion
+
+    #region Recurring component alarms
+
+    [Test]
+    public void RecurringAlarm_RelativeStart_ReturnsAlarmsForEachOccurrence()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(2026, 4, 7, 9, 0, 0, CalDateTime.UtcTzId),
+            RecurrenceRule = new RecurrenceRule(FrequencyType.Weekly, 1) { Count = 4 }
+        };
+
+        e.Alarms.Add(new Alarm
+        {
+            Trigger = new Trigger(new Duration(minutes: -15))
+        });
+
+        var results = e.GetAlarmOccurrences(DateTimeZone.Utc)
+            .Select(x => x.Start.ToInstant())
+            .ToList();
+
+        Assert.That(results, Is.EqualTo(new[]
+        {
+            Instant.FromUtc(2026, 4,  7, 8, 45, 0),
+            Instant.FromUtc(2026, 4, 14, 8, 45, 0),
+            Instant.FromUtc(2026, 4, 21, 8, 45, 0),
+            Instant.FromUtc(2026, 4, 28, 8, 45, 0),
+        }));
+    }
+
+    [Test]
+    public void RecurringAlarm_WithRepeatDuration_ProducesRepetitions()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(2026, 4, 7, 9, 0, 0, CalDateTime.UtcTzId),
+            RecurrenceRule = new RecurrenceRule(FrequencyType.Weekly, 1) { Count = 2 }
+        };
+
+        e.Alarms.Add(new Alarm
+        {
+            Trigger  = new Trigger(new Duration(minutes: -15)),
+            Repeat   = 1,
+            Duration = new Duration(minutes: 5)
+        });
+
+        var results = e.GetAlarmOccurrences(DateTimeZone.Utc)
+            .Select(x => x.Start.ToInstant())
+            .ToList();
+
+        Assert.That(results, Is.EqualTo(new[]
+        {
+            Instant.FromUtc(2026, 4,  7, 8, 45, 0),
+            Instant.FromUtc(2026, 4,  7, 8, 50, 0),
+            Instant.FromUtc(2026, 4, 14, 8, 45, 0),
+            Instant.FromUtc(2026, 4, 14, 8, 50, 0),
+        }));
+    }
+
+    [Test]
+    public void RecurringAlarm_MultipleAlarms_ReturnsMergedOrdered()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(2026, 4, 7, 9, 0, 0, CalDateTime.UtcTzId),
+            RecurrenceRule = new RecurrenceRule(FrequencyType.Weekly, 1) { Count = 2 }
+        };
+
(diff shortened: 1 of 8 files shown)


minichma:
@axunonb Is it ok if I push something onto your branch?

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -275,4 +275,123 @@ public static IEnumerable<T> Catch<T, TException>(this IEnumerable<T> source, Ac
...
+    /// meaning it only enumerates the input sequences while the output sequence is being enumerated.
+    /// Both, the outer as well as the inner sequences may therefore be unbounded.
+    ///
+    /// The behavior is similar to that of <see cref="OrderedMergeMany{T}(IEnumerable{IEnumerable{T}}, IComparer{T})"/>,
+    /// with following differences:
+    /// * Additionally to the inner sequences, also the outer sequences must be ordered. They must be ordered
+    ///   by the head of the inner sequence according to the given comparer.
+    /// * The outer sequence is enumerated in a streaming manner, only as far as the outer is enumerated.
+    /// </remarks>
+    public static IEnumerable<T> OrderedNestedMergeMany<T>(this IEnumerable<IEnumerable<T>> sequences, IComparer<T> comparer)
```
minichma, line 326: Was trying to reduce the complexity of this method to make SonarCloud happy but that's not really helpful. Splitting into multiple methods just makes the code less readable instead. I suggest to suppress the warning.

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -209,8 +210,72 @@ public virtual IEnumerable<Occurrence> GetOccurrences(DateTimeZone timeZone, Ins
...
+                // order when occurrence durations differ (e.g. PERIOD RDATEs). Sort explicitly.
+                // This is safe because occurrences is always a materialized list.
+                return occurrences
+                    .SelectMany(o =>
+                    {
+                        var baseFireTime = o.End.LocalDateTime.Plus(period).InZoneLeniently(o.End.Zone);
+                        return alarm.GetFireTimes(baseFireTime)
+                            .Select(ft => new AlarmOccurrence(alarm, ft, this));
+                    })
+                    .OrderBy(ao => ao.Start.ToInstant());
```
minichma, line 262: The .OrderBy() forces a full enumeration, which will fail in the case of indefinite recurrences.

Ical.Net.Tests/AlarmTests.cs
```diff
@@ -0,0 +1,421 @@
...
+using Ical.Net.Serialization;
+using NodaTime;
+using NUnit.Framework;
+using Duration = Ical.Net.DataTypes.Duration;
+using Period = Ical.Net.DataTypes.Period;
+
+namespace Ical.Net.Tests;
+
+[TestFixture]
+public class AlarmTests
```
minichma, line 19: There's also a file named `AlarmTest.cs` but it contains the class `AtendeeTest`. Seems to be some leftover.

axunonb: yes, oversaw to delete it

minichma:
I pushed some refinement of the `OrderedNestedMergeMany` functionality (find any better name?) and a test reproducing the problem I commented with indefinite recurrences having a 'related to end' trigger. Please feel free to squash or discard as it is helpful.

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -209,8 +210,72 @@ public virtual IEnumerable<Occurrence> GetOccurrences(DateTimeZone timeZone, Ins
...
+    /// <remarks>
+    /// Component occurrences are evaluated once and shared across all alarms.
+    /// When <paramref name="endTime"/> is <c>null</c> and the component has an unbounded recurrence rule,
+    /// evaluation will continue indefinitely; callers must apply their own termination condition.
+    /// </remarks>
+    public virtual IEnumerable<AlarmOccurrence> GetAlarmOccurrences(DateTimeZone timeZone, Instant? startTime = null, Instant? endTime = null, EvaluationOptions? options = null)
+    {
+        if (!Alarms.Any()) return [];
+
+        var rawOccurrences = GetOccurrences(timeZone, startTime, options);
```
minichma, line 229: When filtering by the startTime, we need to consider the trigger duration/offset. If it is positive, the the alarm will fire *after*  the event's occurrence. If the filter startTime is in between, the following alarm won't be returned although it probably should.

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -209,8 +210,72 @@ public virtual IEnumerable<Occurrence> GetOccurrences(DateTimeZone timeZone, Ins
...
+    /// </summary>
+    /// <param name="timeZone">The time zone used to evaluate component occurrences.</param>
+    /// <param name="startTime">Lower bound (inclusive) on alarm fire times, or <c>null</c> for no lower bound.</param>
+    /// <param name="endTime">Upper bound (exclusive) on alarm fire times, or <c>null</c> for no upper bound.</param>
+    /// <remarks>
+    /// Component occurrences are evaluated once and shared across all alarms.
+    /// When <paramref name="endTime"/> is <c>null</c> and the component has an unbounded recurrence rule,
+    /// evaluation will continue indefinitely; callers must apply their own termination condition.
+    /// </remarks>
+    public virtual IEnumerable<AlarmOccurrence> GetAlarmOccurrences(DateTimeZone timeZone, Instant? startTime = null, Instant? endTime = null, EvaluationOptions? options = null)
```
minichma, line 225: If we manage to properly implement streaming, we wouldn't need the `endTime`. To be consistent with GetOccurrence() et al, it would be nice if we could avoid it.

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -209,8 +210,72 @@ public virtual IEnumerable<Occurrence> GetOccurrences(DateTimeZone timeZone, Ins
...
+    {
+        if (!Alarms.Any()) return [];
+
+        var rawOccurrences = GetOccurrences(timeZone, startTime, options);
+        // materialize the component occurrences once and share the list across all alarms
+        var occurrences = endTime == null
+            ? rawOccurrences.ToList()
+            : rawOccurrences.TakeWhile(o => o.Start.ToInstant() <= endTime).ToList();
```
minichma, line 233: Hm, if we materialize here, we won't be able to implement this without the endTime filter. The reason is that we want to avoid iterating the sequence multipe times in case of multiple alarms, right? I wouldn't expect multiple Alarms per event to be a very common use case. If it *does* happen, we iterate multiple times, but I don't see that's a problem. It scales linearly with the number of AlarmOccurrences.

If it should be iterated only once, we could also flip the logic to use the `occurrrences` as the outer sequence and select the alarms per occurrence as the inner sequence, but that's not a critical aspect.

Ical.Net/CalendarComponents/RecurringComponent.cs
minichma: Maybe I just misunderstood the approach. By having the endTime filter, we avoid the need for supporting indefinite sequences. Fair enough. Its just somewhat inconsistent with GetOccurrences() where we teach people to trust in sequences being streaming. A fully streaming solution would require the original sequence to be pre-ordered by the end time. I started drafting such a solution but at this time that exceeds my time budget. Its not overly complex though. It would just require explicitly passing a comparer for `Period`s and `Occurrence`s to all the sorting operators (`OrderedMerge*`, `OrderedExclude`, `SortedSet`, `OrderBy*`, etc.).

Ical.Net/CalendarComponents/RecurringComponent.cs
minichma, line 233: Similarly to the startTime, also when applying the endTime, the trigger's duration must be considered, particularly if it is negative.

Ical.Net/CalendarComponents/RecurringComponent.cs
maknapp: I have been leaning more and more towards the idea of adding end time back in for all occurrence evaluation because it just makes sense to be able to stop evaluating at the end time instead of always having to evaluate to end + 1.

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -275,4 +275,123 @@ public static IEnumerable<T> Catch<T, TException>(this IEnumerable<T> source, Ac
...
     }
+
+    private readonly struct DictKey<T>(T value, long idx)
+    {
+        public readonly T Value = value;
+
+        /// <summary>
+        /// The index is there, because the key of a SortedDictionary or SortedList must be unique.
+        /// </summary>
+        public readonly long Idx = idx;
```
maknapp, line 286: Calling this an "index" is misleading since it is used as a unique ID and not an index into a list. Rename this (and any other references) to `Id`?

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -275,4 +275,123 @@ public static IEnumerable<T> Catch<T, TException>(this IEnumerable<T> source, Ac
...
+
+        var active = new SortedDictionary<DictKey<T>, IEnumerator<T>>(keyComparer);
+
+        var pending = NextNonEmpty();
+        if (pending == null)
+            yield break;
+
+        try
+        {
+            long uniqueIdx = 0;
```
maknapp, line 344: Same here, `uniqueId`?

Ical.Net/Utility/CollectionHelpers.cs
```diff
@@ -275,4 +275,123 @@ public static IEnumerable<T> Catch<T, TException>(this IEnumerable<T> source, Ac
...
+
+        IEnumerator<T>? NextNonEmpty()
+        {
+            while (outer.MoveNext())
+            {
+                var enumerator = outer.Current?.GetEnumerator();
+                if (enumerator?.MoveNext() == true)
+                    return enumerator;
+
+                enumerator?.Dispose();
```
maknapp, line 391: Replace repeated null checks with one `if (outer.Current != null)`

Ical.Net/CalendarComponents/Alarm.cs
```diff
@@ -70,114 +68,19 @@ public Alarm()
...
+    /// <c>REPEAT</c> and <c>DURATION</c>.
     /// </summary>
-    private void AddRepeatedItems(List<AlarmOccurrence> occurrences)
+    internal IEnumerable<ZonedDateTime> GetFireTimes(ZonedDateTime baseFireTime)
     {
-        var len = occurrences.Count;
-        for (var i = 0; i < len; i++)
+        yield return baseFireTime;
+        if (Repeat <= 0 || !Properties.ContainsKey("DURATION")) yield break;
+        var period = Duration!.Value.ToPeriod();
```
maknapp, line 78: I believe `Duration` still needs to be applied using the nominal and time parts, with `InZoneRelativeTo` to handle daylight saving: https://github.com/ical-org/ical.net/blob/36ef4161528170b627c738e6e58590b08c86e7e3/Ical.Net/Evaluation/TodoEvaluator.cs#L27-L30

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -209,8 +210,72 @@ public virtual IEnumerable<Occurrence> GetOccurrences(DateTimeZone timeZone, Ins
...
+                {
+                    var baseFireTime = o.Start.LocalDateTime.Plus(period).InZoneLeniently(o.Start.Zone);
+                    return alarm.GetFireTimes(baseFireTime)
+                        .Select(ft => new AlarmOccurrence(alarm, ft, this));
+                })
+                .OrderedNestedMergeMany();
+        }
+
+        // else the trigger is absolute, so we ignore the component occurrences
+        var dt = alarm.Trigger.DateTime?.ToZonedDateTime();
```
maknapp, line 276: Absolute triggers do not require evaluating component occurrence, so can this be moved outside of here? An event with only absolute alarms should not need to start evaluating event occurrences at all.

maknapp (changes requested):

maknapp:
Does it make sense to "stream" alarms 100% like this? The common case (looking at Google and Microsoft) is probably only a few non-repeating relative-to-start alarms. Alarms for a single event/todo occurrence can be evaluated without further expansion, unlike rrule evaluation where every step can expand further. I do not think an alarm that repeats 10,000,000 times should be a concern.

Ical.Net.Tests/AlarmTests.cs
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    [Test]
+    public void RecurringAlarm_RelatedEnd_Indefinite()
+    {
+        CalendarEvent e = new()
+        {
+            Start = new CalDateTime(2026, 4, 7, 9, 0, 0, CalDateTime.UtcTzId),
+            Duration = new Duration(hours: 1),
+            RecurrenceRule = new RecurrenceRule(FrequencyType.Daily)
```
maknapp, line 320: Event is daily but expected results are weekly.

Ical.Net/CalendarComponents/RecurringComponent.cs
```diff
@@ -209,8 +210,72 @@ public virtual IEnumerable<Occurrence> GetOccurrences(DateTimeZone timeZone, Ins
...
 
-    public virtual IList<AlarmOccurrence> PollAlarms(DateTimeZone timeZone, Instant? startTime, Instant? endTime)
-        => Alarms.SelectMany(a => a.Poll(timeZone, startTime).TakeWhile(p => (endTime == null) || (p.Start.ToInstant() < endTime))).ToList();
+    private IEnumerable<AlarmOccurrence> GetAlarmOccurrenceStream(Alarm alarm, IList<Occurrence> occurrences)
+    {
+        if (alarm.Trigger == null) return [];
+
+        if (alarm.Trigger.IsRelative)
+        {
+            var period = alarm.Trigger.Duration!.Value.ToPeriod();
```
maknapp, line 248: Should use nominal and time parts here too. ~~Probably cannot use `InZoneRelativeTo` as it is right now because it is only for positive duration values.~~

Edit: Actually this should use `InZoneLeniently` because it is not adding from previous values like REPEAT. Maybe `InZoneRelativeTo` is being used in too many places...

Ical.Net/Utility/CollectionHelpers.cs
minichma: Its actually the index into the sequence of non-empty sequences, but I agree, that's neither obvious nor relevant here.

Ical.Net/CalendarComponents/RecurringComponent.cs
minichma: Adding endTime might save a few clock cycles, but probably not in a relevant quantity as the code basically scales quite nicely. (Actually the evaluation should have a computational complexity of O(N) (N being proportional to the number of recurrences returned), which doesn't change if returning one additional item.) However, adding endTime *does* increase the code complexity if you don't want to drop the streaming capabilities altogether, which is the reason why we removed it in v5. I feel that dropping the streaming capabilities altogether wouldn't be a good idea though, because it would prevent users from simply asking for the next occurrence. I think it is very reasonable to expect a calendaring system being able to answer this question. And for the alarms its basically the same. Its very reasonable to ask for _the next alarm_. While I share your assumption that people will *usually* be interested in the alarms *only* within a certain time frame, I do think that it would be a good idea to be able to return _just the next_ alarm. But this is tricky without a fully streaming implementation. So in the end it's tricky in any case :smile:.

But the streaming approach wouldn't be too complex once a custom ordering of occurrences by EndDate would be implemented. And that's not to complex either but certainly helpful, not only for alarms. Its just a decent amount of work (that could probably done by AI though).

Ical.Net/CalendarComponents/RecurringComponent.cs
minichma: Actually, when the trigger is relative to the end time, we also must consider the duration of the component (Event), because a occurrence that ends after `startTime` could have started before.

Ical.Net/CalendarComponents/RecurringComponent.cs
axunonb: From my viewpoint "Next occurrence" queries must remain cheap and unbounded.

A calendaring library is fundamentally expected to answer *"when is the next alarm?"* -  without the caller knowing anything about the future. With an `endTime`-bounded approach, the caller is **forced to guess** an upper bound:

- Too small → misses the next occurrence entirely, returning nothing
- Too large → wastes evaluation over a potentially huge recurrence expansion

With streaming, the answer is always just:

```csharp
var nextAlarm = calendar.GetAlarms(startTime).FirstOrDefault();
```

No guessing. No over-evaluation. No retry logic if the first window returns empty.

---

This matters especially for **sparse or irregular recurrence rules** (e.g. an alarm that fires once a year, or one with `COUNT=1` far in the future), where a bounded caller has no reasonable way to choose `endTime` without either:

1. Hardcoding an arbitrary large value (fragile), or
2. Looping with expanding windows (reinventing streaming, badly)

Streaming makes the library **correct by default** for the most fundamental calendaring question. Everything else — bounded range queries, `endTime` filtering — can be layered on top with a simple `TakeWhile` or `Where` on the stream.

The pragmatic middle ground would be to implement streaming/ordering first (since it unblocks the "next alarm" use case), and then `EndTime` can be added as a cheap post-filter on the stream - giving us both capabilities without sacrificing either.

> But the streaming approach wouldn't be too complex once a custom ordering of occurrences by EndDate would be implemented. And that's not to complex either but certainly helpful, 

Would be great if you took over this step.

maknapp: > I feel that dropping the streaming capabilities altogether wouldn't be a good idea though

I'm not saying to get rid of streaming, just to allow an optional end time. If an end time is given, evaluation would still be streamed, but optimized for ending as early as possible.

> A calendaring library is fundamentally expected to answer "when is the next alarm?"

What is the purpose of an unbounded query like that though? The real world question would be more like "is there an upcoming alarm?" Does anyone care if their next alarm is 3 years from now? I would only care to know about alarms within hours - weeks at most. For implementing the actual alarm to notify a user, it only needs to know within a short time period: "is there an alarm between now and the next time data is reloaded?"

axunonb: > What is the purpose of an unbounded query like that though?

I think there are examples in real-world scheduling: reminders for annual events, insurance renewals, subscription renewals, birthdays - a user does care if their next alarm is months away, because missing it means a missed payment or a forgotten anniversary. Sparse recurrences are real.

maknapp: I'm not seeing a need for unbounded queries with those examples. A calendar always displays items within a start/end range, so any real use of an unbounded query would not be for display on a calendar. Having no end date implies the need to get the next event even if it occurs tomorrow or 9000 years from now, and having that occurrence be relevant.

> Sparse recurrences are real.

This is _why_ it is an issue that having no end date during evaluation is a problem. In everyday use, evaluating to 3026 is useless because we'll all be dead. There is almost no reason to check for an alarm or event in year 9999, yet sparse recurrences can result in running evaluation until an out of range exception is thrown, which is pointless yet unavoidable right now.

axunonb: > A calendar always displays items within a start/end range, so any real use of an unbounded query would not be for display on a calendar.

Alarms aren't always processed in real-time. A common pattern is to store evaluated alarms in a database and have a scheduler poll for due ones - similar to how CI/CD pipelines schedule jobs, or how email systems queue delayed notifications. In those cases, the caller's "window" is determined by their polling interval or batch strategy, not by anything the library should know about.

> There is almost no reason to check for an alarm or event in year 9999,

The upper limit as well as sparse recurrences is the callers concern.

>  running evaluation until an out of range exception is thrown, which is pointless

There are 2 alternatives to handle it:
a) Throwing an exception is the way to go, when we think the caller should be informed about an overflow.
b) With a graceful overflow stop, the natural DateTime boundary is the limit. No configuration or EndTime is needed. Evaluation simply stops when it would exceed DateTime.MaxValue and returns what was found so far.

Callers apply their own EndTime via TakeWhile.

minichma: > There is almost no reason to check for an alarm or event in year 9999,

But we're not talking about real life scenarios there. Sparse recurrences might be real, but probably not in that extent. Any real-life rule would most likely either have a COUNT, a UNTIL or a reasonable number of recurrences. Its hard to think of a reasonable rule that would run into an overflow exception that would not be a usage error.

Anyhow, having an endTime certainly has certain performance benefits but given the additional complexity it imposes, I feel it would not be worth it. It would not make a _fundamental_ difference regarding the computational complexity. And there were good reasons for getting rid of it. I remember several bugs and discussions we had related to that functionality. So, as nobody is complaining about performance related to that topic and it also doesn't seem to help with that alarm issue, I feel 'endTime' should not be a priority.

maknapp: Some real life scenarios are user mistake or malicious behavior. Similar to #903, using `Calendar.GetOccurrences(...)` is not an option if it is easy to make an event that throws an exception, breaking the entire calendar.

axunonb:
Closed in favor of #944

PR #946, opened 2026-05-10
Reduce use of EvaluationPeriod

maknapp:
Note that this is based off of #934, so really only 2 commits.

`RecurrenceRuleEvaluator` was just wrapping `ZonedDateTime` in an `EvaluationPeriod` with no end time, which meant another `EvaluationPeriod` needed to be made to add an end time. `RecurrenceRuleEvaluator` now yields `ZonedDateTime` directly.

`RecurrenceUtil.GetOccurrences(...)` no longer filters by start/end range because it is a duplicate filter to `RecurringEvaluator.Evaluate()`.

EXDATE date time exclusions were switched to `HashSet` instead of sorting and merging. There is no benefit to merging if the list needs to be sorted.

To help reduce allocations and unnecessary work, RDATE merging and EXDATE filtering is skipped if empty.


minichma (commented):
I'm not the right person to review the whole PR, this is just a comment on using HashSet for EXDATEs. The original idea of using OrderedExclude was probably to be close to the implementation of EXRULEs, where it was more or less required. Once the set has been constructed, the computational cost per recurrence are also somewhat lower in case of OrderedExclude compared to a HashSet lookup, but that's neglectable. Now that it seems that support for EXRULEs has been dropped, there's no strong argument for OrderedExclude over a HashSet lookup. I assume the reason for removing it is to make the code easier to read and somewhat less complex to reason about, right? No concerns from my side.

maknapp:
> I'm not the right person to review the whole PR, this is just a comment on using HashSet for EXDATEs.

Yes, sorry. That specific part was the reason I included you to review - probably should have just asked that directly @ you instead of including to review.

> The original idea of using OrderedExclude was probably to be close to the implementation of EXRULEs, where it was more or less required. Once the set has been constructed, the computational cost per recurrence are also somewhat lower in case of OrderedExclude compared to a HashSet lookup, but that's neglectable. Now that it seems that support for EXRULEs has been dropped, there's no strong argument for OrderedExclude over a HashSet lookup. I assume the reason for removing it is to make the code easier to read and somewhat less complex to reason about, right? No concerns from my side.

Yes, since it was the only remaining use of `OrderedExclude` left and `HashSet` was good enough for this case. I just wanted to double-check that it was not doing something I was missing.

axunonb (approved):
Sorry for the delay, LGTM, thanks

Issue #909, opened 2026-01-18
Remove deprecated EXRULE and others


maknapp:
Are all items marked with `Obsolete` since v5 ready to be removed? Removing `EXRULE` will simplify evaluation. `RecurrenceId` and `AlarmAction.Procedure` are the others.

axunonb:
Yes, they are ready to be removed

PR #947, opened 2026-05-25
Remove obsolete RecurrenceId

maknapp:
Remove obsolete `RecurrenceId`.

Added `RecurrenceIdentifier` to `IRecurrable` - this probably should have been done in v5. Any reason to do this in v5 now? Using `IRecurrable` directly is probably not very common.

This still ignores THISANDFUTURE in the same way as before.

Related #909, #455

### DIFF
--- a/Ical.Net.Tests/RecurrenceIdentifierTests.cs
+++ b/Ical.Net.Tests/RecurrenceIdentifierTests.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // Copyright ical.net project maintainers and contributors.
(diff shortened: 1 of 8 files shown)


axunonb (approved):
Consequent to rename, although the shorter `RecurrenceId` was closer to the RFC `RECURRENCE-ID`

axunonb:
> This still ignores THISANDFUTURE in the same way as before.

Yes. Should we follow-up in a new issue?

maknapp:
There is already an issue for this - #455

PR #958, opened 2026-07-03
Match recurrence ID by Instant

maknapp:
The RECURRENCE-ID property was being converted to DateTime to match against occurrences, which would only compare local time. If the RECURRENCE-ID was specified in a different time zone, it would match with the wrong occurrence or nothing at all.

This converts the RECURRENCE-ID to Instant using the specified time zone or the evaluation time zone. This could be a breaking change for events in a time zone with a floating RECURRENCE-ID.

Building the dictionary was also changed to remove the need to use GroupBy and sorting.

### DIFF
--- a/Ical.Net/Calendar.cs
+++ b/Ical.Net/Calendar.cs
@@ -198,7 +198,7 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
     public virtual IEnumerable<Occurrence> GetOccurrences<T>(DateTimeZone timeZone, Instant? startTime = null, EvaluationOptions? options = null) where T : IRecurringComponent
     {
         // Get UID/RECURRENCE-ID combinations that replace occurrences
-        var recurrenceIdsAndUids = GetRecurrenceIdsAndUids(Children);
+        var recurrenceIdsAndUids = GetRecurrenceIdsAndUids(timeZone, Children);
 
         var occurrences = RecurringItems
             .OfType<T>()
@@ -206,7 +206,7 @@ public class Calendar : CalendarComponent, IGetOccurrencesTyped, IGetFreeBusy, I
                 // Exclude occurrences that are overridden by other components with the same UID and RECURRENCE-ID.
                 // This must happen before .OrderedDistinct() because that method would remove duplicates
                 // based on the occurrence time, and we need to remove them based on UID + RECURRENCE-ID.
-                .Where(r => IsUnmodifiedOccurrence(r, recurrenceIdsAndUids)))
+                .Where(r => IsUnmodifiedOccurrence(timeZone, r, recurrenceIdsAndUids)))
 
             // Enumerate the list of occurrences (not the occurrences themselves) now to ensure
             // the initialization code is run, including validation and error handling.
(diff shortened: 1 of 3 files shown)


Ical.Net/DataTypes/Occurrence.cs
```diff
@@ -33,30 +31,6 @@ public Occurrence(IRecurrable recurrable, ZonedDateTime start, ZonedDateTime end
         End = end;
     }
 
-    public CalDateTime? DtStart
```
axunonb, line 36: Removing the property is fine, but must be considered a breaking change. Maybe add to PR comment?

Ical.Net/Calendar.cs
```diff
@@ -233,53 +233,96 @@ public virtual IEnumerable<Occurrence> GetOccurrences<T>(DateTimeZone timeZone,
...
+            // Ignore components without a UID
+            if (r is not IUniqueComponent { Uid: not null } uc)
+            {
+                continue;
+            }
+
+            var key = GetOccurrenceKey(timeZone, uc.Uid, rid);
+
+            // Use the maximum SEQUENCE if present, otherwise use the latest
+            if (uc is CalendarEvent calEvent
```
axunonb, line 259: Nice(r)

axunonb:
> This could be a breaking change for events in a time zone with a floating RECURRENCE-ID.

Isn't this case anyway undefined according to RFC5545?

Mandatory Alignment: 
The RECURRENCE-ID property **MUST** be specified as a date with local time (floating) **if and only if** the DTSTART property of the master recurring component is also specified as a date with local time.

Master with Time Zone: 
If the master events DTSTART is specified using "Form #3: Date with Local Time and Time Zone Reference". the RECURRENCE-ID **MUST NOT** be floating; it must also include a time zone reference or be specified in UTC.

When introducing the Logger to ical.net, the idea was to output warnings with the logger.
In this case: The combination is undefined according to RFC5545, and we handle it [description].
Might lead to more transparency.

axunonb (approved):
Nice!

Ical.Net/DataTypes/Occurrence.cs
maknapp: This is not a breaking change because it was not there in v5. This PR removes its only usage within ical.net. The only other purpose for it would be to provide users with an easy way to determine the original date(-time) type of the source relative to the occurrence. I figured it would be better to just remove it and add it back in later if that is a common user issue.

maknapp:
> > This could be a breaking change for events in a time zone with a floating RECURRENCE-ID.
> 
> Isn't this case anyway undefined according to RFC5545?
> 
> Mandatory Alignment: The RECURRENCE-ID property **MUST** be specified as a date with local time (floating) **if and only if** the DTSTART property of the master recurring component is also specified as a date with local time.

I missed the "and only if" part. That simplifies things because there is no breaking change then. Should I remove the test for that then? Or add a comment clearly stating that this is outside of the RFC and just a test to document behavior? Or would it be better to ignore RECURRENCE-IDs that are "invalid" relative to DTSTART?

> When introducing the Logger to ical.net, the idea was to output warnings with the logger.

I have seen the logger stuff and have been mostly confused by it. What does it log to? How would users see this warning? I assumed the logger does not end up in the user-facing library.

axunonb:
> > > This could be a breaking change for events in a time zone with a floating RECURRENCE-ID.
> > 
> > 
> > Isn't this case anyway undefined according to RFC5545?
> > Mandatory Alignment: The RECURRENCE-ID property **MUST** be specified as a date with local time (floating) **if and only if** the DTSTART property of the master recurring component is also specified as a date with local time.
> 
> I missed the "and only if" part. That simplifies things because there is no breaking change then. Should I remove the test for that then? Or add a comment clearly stating that this is outside of the RFC and just a test to document behavior? Or would it be better to ignore RECURRENCE-IDs that are "invalid" relative to DTSTART?
> >
> > _Keeping the test can help to keep the present behavior. So I'd keep it._
> >


> > When introducing the Logger to ical.net, the idea was to output warnings with the logger.
>
> I have seen the logger stuff and have been mostly confused by it. What does it log to? How would users see this warning? I assumed the logger does not end up in the user-facing library.
> >
> > _That's understandable. The logger didn't make into production use (yet). It's just used in unit tests. But still: I think it could be useful for ical.net. How users could subscribe: see `internal static class LoggingProvider` and `SetLoggerFactory`. We went this way to avoid another dependency for logging._

Issue #648, opened 2024-11-21
Introduce and throw more specific exception types.


minichma:
Today many exceptions that are raised throughout the lib are quite unspecific. E.g. in `RecurrencePatternSerializer`, mostly `ArgumentExceptions` are raised, while something like a `CalendarFormatException` or similar would be more appropriate.

Introduce and throw more specific exception types.

PR #959, opened 2026-07-11
Calendar time zone provider

### LINKED ISSUE #749: Unrecognized time zone id Customized Time Zone

maknapp:
This adds a `Calendar.TimeZoneProvider` property to allow users to specify the `NodaTime.IDateTimeZoneProvider` to use during calendar/event evaluation. The default time zone provider for a calendar is `CalendarTimeZoneProviders.TzdbWithAliases` with the calendar defined time zones as the fallback. The default time zone provider for components without a calendar parent is `CalendarTimeZoneProviders.TzdbWithAliases`.

### User choice
Users can choose the time zone provider:
* Just use the default
* Change the fallback(s) by using `CalendarTimeZoneProviders.Combine()` to create their own `NodaTime.IDateTimeZoneProvider` that searches the specified providers in the given order for a matching time zone
* Be more strict by using `NodaTime`'s `DateTimeZoneProviders.Tzdb` directly - with or without the calendar's VTIMEZONE components
* Use only the calendar's VTIMEZONE components without any other provider (`cal.TimeZoneProvider = cal.CreateTimeZoneProvider()`)

### VTIMEZONE
VTIMEZONE changes are not automatically handled. VTIMEZONE components are converted to `NodaTime.DateTimeZone` values on demand and cached. If a VTIMEZONE component is added/removed/changed AFTER being used by the time zone provider, the `Calendar.TimeZoneProvider` will need to be recreated to be up-to-date with the calendar VTIMEZONE components.

One (unlikely?) issue users might run into is that `NodaTime.ZonedDateTime` checks equality by `ReferenceEquals` of its `DateTimeZone`, so recreating the `Calendar.TimeZoneProvider` could cause issues if the user tries to compare a `ZonedDateTime` with another created from a different time zone provider because they would have two different `DateTimeZone` objects even though they represent the same VTIMEZONE.

### Exceptions
Missing time zone errors will now throw NodaTime's DateTimeZoneNotFoundException instead of ArgumentException. I'm not sure if there are downsides vs catching and rethrowing with our own `TimeZoneNotFoundException`. I do like having a more specific exception instead of `ArgumentException`. Related #648.

No more time zone exceptions in Calendar.Load(). As stated above, exceptions due to missing time zones are now thrown during evaluation.

### Time zone searching
The `Calendar.TimeZoneProvider` is created on demand with a default value if none is specifically set. The default is NOT an exact copy of the previous `TimeZoneResolver`. Since this brings more time zone control to the user, it seemed reasonable to remove the more non-standard searching by default: case-insensitive and `string.Contains` searching. I did add a case-insensitive time zone provider in case that is actually needed - I would like to remove that commit though unless users actually need it. I ignored adding a `string.Contains` replacement because that seems very inaccurate to me.

Searching `XmlSerializationSettings` was also removed completely because it is, by NodaTime default, just the TZDB. What is the purpose of searching `NodaTime.Xml.XmlSerializationSettings.DateTimeZoneProvider`? The user can add this provider if they actually use this, so I think it is safe to ignore it completely.

### Global time zone handling removed
`DefaultTimeZoneResolver` and `TimeZoneResolvers` were removed. Having a default time zone provider on the calendar is essentially the same thing for users who currently ignore time zones.

### CalDateTime
This now requires a time zone provider to convert to a `ZonedDateTime`. @axunonb mentioned in #749 that there should be a resolver-less overload still, but a user converting to a `ZonedDateTime` should always be paying attention to the time zone provider anyway. If a user _really_ wants to ignore it, they can just use extensions to use a default time zone provider.

`ToDateTimeUtc()` is also removed just because I did not want to require a `NodaTime` time zone provider to create a `DateTime`. Users avoiding `NodaTime` can still use `ToDateTimeUnspecified()` to get just the local time.  Users can also use extensions to create a `ToDateTimeUtc()` however they want.

None of these changes are breaking changes for `CalDateTime` because they are not in v5.

Fixes #749


axunonb:
That sounds pretty nice. I'll need some time to go through changes.

Ical.Net/CalendarComponents/VTimeZone.cs
```diff
@@ -367,4 +394,123 @@ public string? Location
...
+            }
+
+            if (info == null)
+            {
+                throw new Exception("Could not find zone interval");
+            }
+
+            var name = info.TimeZoneName ?? "Unknown";
+
+            var wallOffset = Offset.FromTimeSpan(info.OffsetTo!.Offset);
```
axunonb, line 548: Should this rather be this way? (subtract OffsetFrom)
Not covered by tests though. See suggested additional tests in VTimeZoneTest
```cs
var wallOffset = Offset.FromTimeSpan(info.OffsetTo!.Offset);
var savings = info.Name == Components.Standard
    ? Offset.Zero
    : Offset.FromTimeSpan(info.OffsetTo!.Offset) - Offset.FromTimeSpan(info.OffsetFrom!.Offset);
```

Ical.Net/CalendarComponents/VTimeZone.cs
```diff
@@ -367,4 +394,123 @@ public string? Location
...
+
+    private sealed class CalendarDateTimeZone : DateTimeZone
+    {
+        private readonly List<VTimeZoneInfo> intervals;
+
+        public static CalendarDateTimeZone From(VTimeZone data)
+        {
+            if (data.TzId == null)
+            {
+                throw new Exception("Time zone ID must be set");
```
axunonb: Maybe introduce a specific VTimeZoneException?
If not use InvalidOperationException?

Ical.Net/CalendarComponents/VTimeZone.cs
```diff
@@ -367,4 +394,123 @@ public string? Location
...
+                    {
+                        maxStart = tzStart;
+                        info = tz;
+                    }
+                }
+            }
+
+            if (info == null)
+            {
+                throw new Exception("Could not find zone interval");
```
axunonb: See above - VTimeZoneException?

Ical.Net.Tests/VTimeZoneTest.cs
```diff
@@ -264,6 +267,208 @@ public void VTimeZoneAmericaDetroitShouldSerializeProperly()
...
+        {
+            // America/New_York is not defined in the calendar, but it
+            // should still be able to evaluate.
+            results = cal.GetOccurrences(start).ToList();
+        });
+
+        Assert.That(results, Has.Count.EqualTo(1));
+    }
+
     private static Calendar CreateTestCalendar(string tzId, DateTime? earliestTime = null, bool includeHistoricalData = true)
```
axunonb, line 472: These additional tests fail with the original code in `VTimeZone` line 507-511

```cs
    [Test, Category("VTimeZone")]
    public void VTimeZone_CalendarDateTimeZone_GetZoneInterval_Should_Calculate_Savings_Correctly()
    {
        // Arrange: Create a calendar with a custom VTIMEZONE that has daylight saving time
        var icalString = @"BEGIN:VCALENDAR
VERSION:2.0
PRODID:ical.net
BEGIN:VTIMEZONE
TZID:Test/DST
BEGIN:STANDARD
DTSTART:19701101T020000
RRULE:FREQ=YEARLY;BYMONTH=11;BYDAY=1SU
TZOFFSETFROM:-0400
TZOFFSETTO:-0500
TZNAME:EST
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:19700308T020000
RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=2SU
TZOFFSETFROM:-0500
TZOFFSETTO:-0400
TZNAME:EDT
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
DTSTART;TZID=Test/DST:20250615T120000
DTEND;TZID=Test/DST:20250615T130000
SUMMARY:Test Event
UID:test@example.com
END:VEVENT
END:VCALENDAR";

        var calendar = Calendar.Load(icalString);
        var vtz = calendar.TimeZones.First();

        // Act: Get the DateTimeZone from the VTIMEZONE
        var dateTimeZone = vtz.ToDateTimeZone();

        // Get a zone interval during daylight saving time (June 2025)
        var instantInDst = Instant.FromUtc(2025, 6, 15, 16, 0); // 12:00 EDT = 16:00 UTC
        var dstInterval = dateTimeZone.GetZoneInterval(instantInDst);

        // Get a zone interval during standard time (December 2025)
        var instantInStd = Instant.FromUtc(2025, 12, 15, 17, 0); // 12:00 EST = 17:00 UTC
        var stdInterval = dateTimeZone.GetZoneInterval(instantInStd);

        // Assert: Daylight saving time should have non-zero savings
        using (Assert.EnterMultipleScope())
        {
            // During daylight time (EDT), wall offset is -04:00
            Assert.That(dstInterval.WallOffset.ToTimeSpan(), Is.EqualTo(TimeSpan.FromHours(-4)),
                "Wall offset during DST should be -04:00");

            // Savings should be +01:00 (the difference from standard time)
            Assert.That(dstInterval.Savings.ToTimeSpan(), Is.EqualTo(TimeSpan.FromHours(1)),
                "Savings during DST should be +01:00");

            // Standard offset should be -05:00 (wall offset - savings)
            Assert.That(dstInterval.StandardOffset.ToTimeSpan(), Is.EqualTo(TimeSpan.FromHours(-5)),
                "Standard offset during DST should be -05:00");

            // During standard time (EST), wall offset is -05:00
            Assert.That(stdInterval.WallOffset.ToTimeSpan(), Is.EqualTo(TimeSpan.FromHours(-5)),
                "Wall offset during standard time should be -05:00");

            // Savings should be zero during standard time
            Assert.That(stdInterval.Savings.ToTimeSpan(), Is.EqualTo(TimeSpan.Zero),
                "Savings during standard time should be zero");

            // Standard offset equals wall offset during standard time
            Assert.That(stdInterval.StandardOffset.ToTimeSpan(), Is.EqualTo(TimeSpan.FromHours(-5)),
                "Standard offset during standard time should be -05:00");
        }
    }

    [Test, Category("VTimeZone")]
    public void VTimeZone_CalendarDateTimeZone_GetZoneInterval_Savings_Should_Match_NodaTime_Tzdb()
    {
        // Arrange: Create a VTIMEZONE from a known TZDB zone with daylight saving time
        var tzId = "America/New_York";
        var vtz = VTimeZone.FromDateTimeZone(tzId);
        var calendar = new Calendar();
        calendar.AddChild(vtz);

        var calendarDateTimeZone = vtz.ToDateTimeZone();
        var tzdbZone = CalendarTimeZoneProviders.TzdbWithAliases[tzId];

        // Act & Assert: Compare savings during both DST and standard time
        var testInstants = new[]
        {
        Instant.FromUtc(2025, 7, 15, 12, 0),  // Mid-summer (DST active)
        Instant.FromUtc(2025, 1, 15, 12, 0)   // Mid-winter (standard time)
    };

        foreach (var instant in testInstants)
        {
            var tzdbInterval = tzdbZone.GetZoneInterval(instant);
            var calendarInterval = calendarDateTimeZone.GetZoneInterval(instant);

            using (Assert.EnterMultipleScope())
            {
                Assert.That(calendarInterval.WallOffset, Is.EqualTo(tzdbInterval.WallOffset),
                    $"Wall offset mismatch at {instant}");

                Assert.That(calendarInterval.Savings, Is.EqualTo(tzdbInterval.Savings),
                    $"Savings mismatch at {instant}: Expected {tzdbInterval.Savings}, but got {calendarInterval.Savings}");

                Assert.That(calendarInterval.StandardOffset, Is.EqualTo(tzdbInterval.StandardOffset),
                    $"Standard offset mismatch at {instant}");
            }
        }
    }
```

Ical.Net/CalendarTimeZoneProviders.cs
```diff
@@ -0,0 +1,223 @@
...
+                    ?? throw new DateTimeZoneNotFoundException($"Time zone ID {id} is unknown");
+
+                return zone;
+            }
+        }
+
+        public string VersionId => field ??= "combined: " + string.Join(", ", providers.Select(x => x.VersionId));
+
+
+        public DateTimeZone GetSystemDefault() => throw new NotImplementedException();
```
axunonb: Maybe implement e.g. like this?
```cs
public DateTimeZone GetSystemDefault()
{
    return providers.FirstOrDefault()?.GetSystemDefault()
        ?? DateTimeZoneProviders.Tzdb.GetSystemDefault();
}
```

Ical.Net/Evaluation/TimeZoneInfoEvaluator.cs
```diff
@@ -18,6 +18,9 @@ protected VTimeZoneInfo TimeZoneInfo
     }
 
     protected override NodaTime.ZonedDateTime GetEnd(NodaTime.ZonedDateTime start) => start;
+
+    protected override NodaTime.IDateTimeZoneProvider TimeZoneProvider => CalendarTimeZoneProviders.TzdbWithAliases;
+
```
axunonb, line 23: Would it make sense to implement like this?

```cs
protected override EvaluationPeriod EvaluateRDate(Period rdate, NodaTime.DateTimeZone referenceTimeZone)
{
    var start = rdate.StartTime.ToZonedOrDefault(referenceTimeZone, TimeZoneProvider);
    var end = GetEnd(start);
    return new EvaluationPeriod(start, end);
}
```
VTimeZoneInfo is used to represent STANDARD and DAYLIGHT components within a VTIMEZONE
(Time zone transitions are instants, not periods with duration.)

axunonb (commented):
This looks very well designed to me, great!
Nice fix to  ` RECURRENCE-ID`.
Few potential issues I came across. See the comments.
Sonar cloud comments: left to your discretion.

Ical.Net/CalendarComponents/VTimeZone.cs
maknapp: Yes, fixed.

Ical.Net.Tests/VTimeZoneTest.cs
maknapp: I added the above tests with one changed to directly compare NodaTime intervals.

Ical.Net/CalendarTimeZoneProviders.cs
maknapp: It seems wrong to use TZDB if it is not in the list of providers here. I'm not sure if there is a use for this, which is why I had skipped it. I changed it to loop through providers to get the first system default.

Ical.Net/Evaluation/TimeZoneInfoEvaluator.cs
maknapp: The start value should always be used with the "offset from" value. I added RDATE support using the offset directly.

maknapp:
Tests fail when converting to and from `VTimeZone` when you compare `NodaTime` zone intervals directly because `VTimeZone.FromDateTimeZone` produces historical info using RDATE instead of UNTIL. Intervals that span multiple years are also not grouped together.

I have been working on making it use RRULEs always while making round-trip converting stay equal to `NodaTime` time zones, but it is still not quite correct for every time zone. I'm also not sure how useful this is anyway. The real way to do this would be to read the time zone source data directly instead of using `NodaTime`'s intervals.

Does any major calendar even produce complete VTIMEZONE data?

axunonb (approved):
LGTM

axunonb:
> Does any major calendar even produce complete VTIMEZONE data?

Many, like Outlook, just add the VTIMEZONE in a very simplified way, never the whole history (if I understood "complete" correctly). Look at the DTSTART ;-)
```ics
BEGIN:VCALENDAR
PRODID:-//Microsoft Corporation//Outlook 16.0 MIMEDIR//EN
VERSION:2.0
METHOD:PUBLISH
X-MS-OLK-FORCEINSPECTOROPEN:TRUE
BEGIN:VTIMEZONE
TZID:W. Europe Standard Time
BEGIN:STANDARD
DTSTART:16011028T030000
RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=10
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:16010325T020000
RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=3
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
```

PR #964, opened 2026-07-22
Readme: fix duplicate text

JimmyCushnie:
The duplicate text makes for a clunky read. I just noticed this and figured I'd open a PR to fix it.


axunonb (approved):
Good catch, thanks

PR #965, opened 2026-07-24
Escape backslash for TEXT values

### LINKED ISSUE #963: Literal backslashes aren't escaped correctly in text properties
**Describe the bug**
When a text property such as CalendarEvent.Description or CalendarEvent.Location contains a literal backslash `\`, the serializer doesn't escape it to `\\`, causing it to be treated as part of an escape sequence or dropped by consumers. Other special characters like commas are automatically escaped, so I would expect the same behavior here instead of needing to pre-escape the backslashes myself.

**To Reproduce**
Serialize a calendar with an event where the Description field has a backslash in it.

Unit tests:
```cs
        [TestCase(@"Lab, Training", @"Lab\, Training")] // Passes
        [TestCase(@"Lab; Training", @"Lab\; Training")] // Passes
        [TestCase(@"Lab \ Training", @"Lab \\ Training")] // Fails, consumer sees "Lab  Training"
        [TestCase(@"Lab \\ Training", @"Lab \\\\ Training")] // Fails, consumer sees "Lab \ Training"
        [TestCase(@"Orientation\New Hire", @"Orientation\\New Hire")] // Fails, consumer sees "Orientation<newline>ew Hire"
        public void Verify_serialization_escapes_text_special_characters(string description, string serializedDescription)
        {
            var calendar = new Calendar();
            var calEvent = new CalendarEvent { Description = description };
            calendar.Events.Add(calEvent);

            var result = new CalendarSerializer().SerializeToString(calendar);
            Assert.That(result, Does.Contain($"DESCRIPTION:{serializedDescription}"));

            var resultCal = Calendar.Load(result);
            Assert.That(resultCal.Events[0].Description, Is.EqualTo(description));
        }
```

**Expected behavior**
The serializer should automatically escape literal backslashes as `\\`.

- OS: Windows 11
- .NET version: .NET Framework 4.8
- ical.net version: 5.2.3

maknapp:
This also improves escape/unescape performance and reduces allocations.

I added net48 to Benchmarks target frameworks so that we can benchmark for netstandard2.0 target.

Fixes #963

### DIFF
--- a/Ical.Net.Benchmarks/Ical.Net.Benchmarks.csproj
+++ b/Ical.Net.Benchmarks/Ical.Net.Benchmarks.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFrameworks>net10.0</TargetFrameworks>
+        <TargetFrameworks>net10.0;net48</TargetFrameworks>
         <OutputType>Exe</OutputType>
         <nullable>enable</nullable>
         <LangVersion>latest</LangVersion>
--- a/Ical.Net.Benchmarks/Runner.cs
+++ b/Ical.Net.Benchmarks/Runner.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // Copyright ical.net project maintainers and contributors.
 // Licensed under the MIT license.
 //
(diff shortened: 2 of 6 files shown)


axunonb (approved):
LGTM, tnx.
Maybe consider Sonar Cloud comments for benchmarks.

maknapp:
Sonar is wrong, nullable is enabled there.

PR #968, opened 2026-07-28
Split multi-value TEXT on unescaped separator commas

gaoflow:
### Problem

Multi-valued TEXT properties (CATEGORIES, RESOURCES, ...) are split on separator commas with a single-character negative lookbehind:

```csharp
internal static readonly Regex UnescapedCommas = new Regex(@"(?<!\\),", ...);
```

The lookbehind only inspects the one character before the comma, so it cannot count backslash parity. A value that ends in a backslash is escaped to end in `\\` (an escaped backslash), so the serialized stream contains `...\\,` where the comma is a genuine value separator. The lookbehind sees the trailing `\` and refuses to split.

This is a direct sibling of the backslash-escaping fix merged in #965: the encoder now correctly emits `\\` before a list separator, but the splitter was never updated to cope with it, so a list whose values contain backslashes no longer round-trips.

Concretely, `[a\, b]` serializes to `CATEGORIES:a\\,b` (correct) but deserializes back to the single value `a\,b` instead of the two values `a\` and `b`.

### Truth table

A comma is a separator iff preceded by an **even** number (incl. 0) of backslashes; the old regex splits only when preceded by **zero**:

| encoded | values | old | fixed |
|---|---|---|---|
| `a,b` | `a`, `b` | 2 ✓ | 2 ✓ |
| `a\,b` | `a,b` | 1 ✓ | 1 ✓ |
| `a\\,b` | `a\`, `b` | **1 ✗** | 2 ✓ |
| `a\\\,b` | `a\,b` | 1 ✓ | 1 ✓ |
| `a\\\\,b` | `a\\`, `b` | **1 ✗** | 2 ✓ |

### Fix

Replace the lookbehind regex with an escape-aware left-to-right scan that counts backslash parity, mirroring the consumption already used by `Unescape`: a backslash escapes the following character, so a comma is a separator only when it is not escaped.

The single fix site governs every multi-valued TEXT property (`GetPropertyAllowsMultipleValues`), so this repairs the whole class, not just CATEGORIES. The other `Split(',')`/`Split(';')` calls in the serializers operate on structured non-TEXT grammars with no backslash escaping and are unaffected; the lookbehind splitter was the only one of its kind.

### Tests

Added a round-trip test over CATEGORIES covering the full truth table above (plain separator, escaped comma inside a value, escaped backslash before a separator, escaped backslash + escaped comma, two escaped backslashes, trailing backslash on the last value, and several separators in one line). The three escaped-backslash cases fail on `main` and pass with the fix; the escaped-comma cases are the opposite-direction controls that must not over-split. Full suite green (2469 tests, net10.0).

### DIFF
--- a/Ical.Net/Serialization/DataTypes/StringSerializer.cs
+++ b/Ical.Net/Serialization/DataTypes/StringSerializer.cs
@@ -10,7 +10,6 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Text.RegularExpressions;
 using Ical.Net.DataTypes;
 
 namespace Ical.Net.Serialization.DataTypes;
@@ -342,7 +341,45 @@ public class StringSerializer : EncodableDataTypeSerializer
         return string.Join(",", values);
     }
 
-    internal static readonly Regex UnescapedCommas = new Regex(@"(?<!\\),", RegexOptions.Compiled, RegexDefaults.Timeout);
+    // A value scan stops at an unescaped comma (a separator) or a backslash
+    // (which escapes the following character).
+    private static readonly char[] _valueSeparators = ['\\', ','];
+
+    /// <summary>
+    /// Splits a serialized multi-value TEXT property on its separator commas. A comma
+    /// separates values only when it is not escaped; a backslash escapes the following
+    /// character, so a comma preceded by an odd number of backslashes belongs to a value
+    /// (RFC 5545 3.1.1). Mirrors the left-to-right consumption used by <see cref="Unescape"/>.
+    /// </summary>
+    internal static IEnumerable<string> SplitOnUnescapedCommas(string value)
+    {
+        var start = 0;
+        var pos = 0;
+
+        while (pos < value.Length)
+        {
+            var next = value.IndexOfAny(_valueSeparators, pos);
+            if (next < 0)
+            {
+                break;
+            }
+
+            if (value[next] == '\\')
+            {
+                // Backslash escapes the next character; skip the pair.
+                pos = next + 2;
+            }
+            else
+            {
+                yield return value.Substring(start, next - start);
+                start = next + 1;
+                pos = next + 1;
+            }
+        }
+
+        yield return value.Substring(start);
+    }
+
     public override object? Deserialize(TextReader? tr)
     {
         if (tr == null)
@@ -367,9 +404,8 @@ public class StringSerializer : EncodableDataTypeSerializer
             AssociatedObject = context.Peek() as ICalendarObject
         };
 
-        var encodedValues = serializeAsList ? UnescapedCommas.Split(value) : new[] { value };
-        var escapedValues = encodedValues.Select(v => Decode(dt, v)).ToList();
-        var values = escapedValues.Select(Unescape).ToList();
+        var encodedValues = serializeAsList ? SplitOnUnescapedCommas(value) : Enumerable.Repeat(value, 1);
+        var values = encodedValues.Select(v => Unescape(Decode(dt, v))).ToList();
 
         if (values.Count == 1)
         {
--- a/Ical.Net/Serialization/DataTypes/StringSerializer.cs
+++ b/Ical.Net/Serialization/DataTypes/StringSerializer.cs
@@ -404,13 +404,14 @@ public class StringSerializer : EncodableDataTypeSerializer
             AssociatedObject = context.Peek() as ICalendarObject
         };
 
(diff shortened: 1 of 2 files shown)


Ical.Net/Serialization/DataTypes/StringSerializer.cs
```diff
@@ -342,7 +341,41 @@ private static string EscapeWithSearchValues(string valueToEscape)
...
+    /// character, so a comma preceded by an odd number of backslashes belongs to a value
+    /// (RFC 5545 3.1.1). Mirrors the left-to-right consumption used by <see cref="Unescape"/>.
+    /// </summary>
+    internal static List<string> SplitOnUnescapedCommas(string value)
+    {
+        var values = new List<string>();
+        var start = 0;
+        var pos = 0;
+
+        while (pos < value.Length)
```
maknapp, line 359: Looping character by character is slow for medium to large strings. Can you use `IndexOfAny`?

Ical.Net/Serialization/DataTypes/StringSerializer.cs
```diff
@@ -367,7 +400,7 @@ private static string EscapeWithSearchValues(string valueToEscape)
             AssociatedObject = context.Peek() as ICalendarObject
         };
 
-        var encodedValues = serializeAsList ? UnescapedCommas.Split(value) : new[] { value };
+        var encodedValues = serializeAsList ? SplitOnUnescapedCommas(value) : new List<string> { value };
```
maknapp: `SplitOnUnescapedCommas(...)` could be changed to an iterator method and yield return each value directly to `Decode(...)` and `Unescape(...)`. This would prevent several unnecessary lists from being created.

maknapp (commented):
This does fix the issue, but I have some suggestions if you would like to work on this more.

binggao1230:
Applied both — the split now uses `IndexOfAny` to jump between `\` and `,` instead of scanning char by char, and `SplitOnUnescapedCommas` is now an iterator whose values flow straight into `Decode`/`Unescape`, so the intermediate lists are gone. Full test suite green.

Ical.Net/Serialization/DataTypes/StringSerializer.cs
```diff
@@ -367,9 +404,8 @@ private static string EscapeWithSearchValues(string valueToEscape)
             AssociatedObject = context.Peek() as ICalendarObject
         };
 
-        var encodedValues = serializeAsList ? UnescapedCommas.Split(value) : new[] { value };
-        var escapedValues = encodedValues.Select(v => Decode(dt, v)).ToList();
-        var values = escapedValues.Select(Unescape).ToList();
+        var encodedValues = serializeAsList ? SplitOnUnescapedCommas(value) : Enumerable.Repeat(value, 1);
```
maknapp: If we know `value` is not a list, it would be better to just return `Unescape(Decode(dt, value))` directly.

Ical.Net/Serialization/DataTypes/StringSerializer.cs
```diff
@@ -342,7 +341,45 @@ private static string EscapeWithSearchValues(string valueToEscape)
...
+    // (which escapes the following character).
+    private static readonly char[] _valueSeparators = ['\\', ','];
+
+    /// <summary>
+    /// Splits a serialized multi-value TEXT property on its separator commas. A comma
+    /// separates values only when it is not escaped; a backslash escapes the following
+    /// character, so a comma preceded by an odd number of backslashes belongs to a value
+    /// (RFC 5545 3.1.1). Mirrors the left-to-right consumption used by <see cref="Unescape"/>.
+    /// </summary>
+    internal static IEnumerable<string> SplitOnUnescapedCommas(string value)
```
maknapp, line 354: Could make a span-based version using `SearchValues` for net8.0+

Ical.Net/Serialization/DataTypes/StringSerializer.cs
gaoflow: Done in 63c9cff8 — the non-list path now returns `Unescape(Decode(dt, value))` directly, no list allocation. Full suite 2469/2469.

Ical.Net/Serialization/DataTypes/StringSerializer.cs
gaoflow: Left as-is for now — the iterator already uses `IndexOfAny`; a `SearchValues` span variant can be a follow-up if you want it.

maknapp (approved):

PR #969, opened 2026-07-28
Implement RFC 6868 caret encoding for parameter values

gaoflow:
A newline in a property parameter value makes `Calendar.Load(Serialize(cal))` throw: the raw newline terminates the content line, so ical.net emits output its own parser rejects.

    att.Parameters.Set("CN", "line1\nline2");
    // Load(Serialize(cal)) -> SerializationException: missing colon

RFC 6868 is the standard remedy and was unimplemented. This adds the symmetric codec: encode `^`->`^^`, DQUOTE->`^'`, newline->`^n` on serialization (§3.1) and the exact inverse when parsing parameter values (§3.2), each a single left-to-right pass.

DQUOTE was previously stripped, under a comment that parameter values MUST NOT contain DQUOTE. RFC 6868 `^'` keeps that guarantee (no literal DQUOTE reaches the wire) while round-tripping the character instead of dropping it, so I implemented that half too.

Tests cover `decode(encode(v)) == v` for `^`, newline, DQUOTE, a literal `^n`/`^^` and mixed values, the `Load(Serialize())` round-trip through the public API, and that plain values stay byte-identical. Full suite passes on net10.0; net8.0/net48 build (no net8 runtime on my machine to run them).

### DIFF
--- a/Ical.Net.Tests/Rfc6868Tests.cs
+++ b/Ical.Net.Tests/Rfc6868Tests.cs
@@ -1,0 +1,123 @@
+﻿//
+// Copyright ical.net project maintainers and contributors.
+// Licensed under the MIT license.
+//
+
+using System.Collections.Generic;
+using Ical.Net.CalendarComponents;
+using Ical.Net.DataTypes;
+using Ical.Net.Serialization;
+using NUnit.Framework;
+
+namespace Ical.Net.Tests;
+
+[TestFixture]
+public class Rfc6868Tests
+{
+    // ---- Codec unit tests (RFC 6868 §3.1 / §3.2) ----
+
+    [TestCase("plain text", "plain text")]
+    [TestCase("a^b", "a^^b")]                 // caret -> ^^
+    [TestCase("say \"hi\"", "say ^'hi^'")]    // DQUOTE -> ^'
+    [TestCase("line1\nline2", "line1^nline2")] // LF -> ^n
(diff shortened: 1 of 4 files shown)


maknapp:
@gaoflow, you appear to be mass committing to a bunch of repos. Does this issue directly affect you in any way? Do you know which major calendar services implement this?

binggao1230:
Hi @maknapp — fair question. Yes, I've been sending correctness fixes across a number of libraries; the account is AI-assisted (an agent finds and drafts the fix) but run under my direction, and I'm accountable for everything that goes in, so I'm glad to drop anything that isn't wanted.

I don't have a production dependency on Ical.Net here — I found this while auditing the serializer against RFC 6868/5545, where a parameter value containing a double-quote or newline currently can't round-trip. I don't have a survey of which calendar clients decode 6868, so if the real-world payoff feels too thin, I'm fine with you closing it — no argument.

maknapp:
While this looks to be backwards compatible, it might not be a good idea to support this if no major calendar providers support this. I tried to add an ATTENDEE to a Google Calendar event as shown in the RFC example, but Google just silently ignored the ATTENDEE. I did not test any other calendars. Throwing an exception in ical.net would at least notify the user, although the exception could be better than "missing colon."

Could also have serializing options to enable support if it is not too complicated.

@axunonb Any opinions on this?

PR #982, opened 2026-09-06
Use BenchmarkSwitcher

maknapp:
Using `BenchmarkSwitcher` would make it easier to run specific benchmarks (instead of temporarily commenting out code).

The manual config seems to just be for allowing Debug mode, but `Debugger.IsAttached` means debugger takes a different route anyway.


axunonb (approved):
LGTM, thanks

Issue #702, opened 2025-01-21
Optimize `IStringSerializer.SerializeToString` for less GC pressure


ssratkevich:
Exploring current implementation of `ComponentSerializer.SerializeToString` and it's siblings in other serialization classes I noticed, that instead of drilling down to callers `StringBuilder` children always create their owns.  
I suggest to have some way to use common (single) `StringBuilder` all way down in the serialization.  
This can reduce memory pressure and allocation of objects that are essentially unnecessary.  

It also opens a way to customization of calendar objects and their serialization for granular calendar building that could pay attention to some external caching strategy of, say, events and others.

angelaki:
Am I mistaken or is a direct serialization to a stream not yet possible? I'd recommend to replace the StringBuilder by a StreamWriter when touching this part of the lib so a direct serialization to a stream might be possible?

axunonb:
@angelaki Does your recommendation go back to lower memory footprint or other benefits of `StreamWriter` that you're currently missing?

angelaki:
> [@angelaki](https://github.com/angelaki) Does your recommendation go back to lower memory footprint or other benefits of `StreamWriter` that you're currently missing?

It's actually more about being able to stream the ICal via an Http endpoint.

axunonb:
Maybe use abstraction like `ICalWriter` with implementations for `StringBuilderWriter` and `StreamWriterWriter` to decouple the serialization logic from the output medium.

Issue #978, opened 2026-09-04
new GeographicLocation(string) Returns 0.0,0.0 co-ordinates


CraigChamberlain:
It looks like the parsing on the Geographic location constructor that takes a string uses a GeographicLocationSerializer object to attempt to parse the input string. Debugging the process on line 21 of GeographicLocation.cs the process seems to work but perhaps the object is not returned and then the naked constructor triggers?

I can't see any tests or examples for valid strings.  I've tried the following, all fail with actual being 0.

            var loc = new GeographicLocation("37.386013;-122.082932");
            Assert.AreEqual(37.386013, loc.Latitude);
            Assert.AreEqual(-122.082932, loc.Longitude);

            var loc2 = new GeographicLocation("GEO:37.386013;-122.082932");
            Assert.AreEqual(37.386013, loc2.Latitude);
            Assert.AreEqual(-122.082932, loc2.Longitude);

However, the following works fine

            var loc = new GeographicLocationSerializer().Deserialize("37.386013;-122.082932");
            Assert.AreEqual(37.386013, loc.Latitude);
            Assert.AreEqual(-122.082932, loc.Longitude);

maknapp:
The string constructor is deserializing but then does not set the properties of the current object. This will be fixed in #973.

Issue #937, opened 2026-04-02
Use static constructors for string parsing


maknapp:
All types with a constructor that parses a string value should change to Parse or TryParse methods.

The `public CalDateTime(string value, string? tzId = null);` should be a static constructor or just removed entirely. Right now it is creating an extra CalDateTime object and copying the values.

It would also be helpful if DateTimeSerializer had a static default instance instead of creating one every time. This constructor is only used in tests right now.

PR #983, opened 2026-09-12
UTF-8 Serializer

### LINKED ISSUE #973: Serialize to UTF-8 byte stream
The current serializer uses nested `StringBuilder`s to serialize to string, even when serializing to `Stream`.

I propose having a single `CalendarSerializer` with converter classes that read/write from/to a UTF-8 byte stream similar to `System.Text.Json`. `CalendarSerializer.SerializeToString()` will serialize to a byte array, then convert to string.

Things to be removed (list likely to change):
* Remove all `SerializeToString` methods. `ICalendarComponents` can be serialized via `CalendarSerializer` but `DataTypes` will just not have a standalone way to serialize.
* Remove all string constructors that current use `_____Serializer` to parse.
* Remove `Calendar.Load(Stream)` and `Calendar.Load(String)` - these will be in `CalendarSerializer` only.
* ~~Remove Serialize events (`OnDeserialized`, `OnDeserializing`, etc.)~~
* Remove support for serializing to other encodings - only support UTF-8. Supporting other encodings would change how lines are folded for each encoding.

This would fix #702 and #937.

Some of #908 could be fixed depending on changes.

---

@axunonb, I have been testing various changes to the current serializer vs a new one as described above. I'm leaning towards this route, so I wanted to write down a draft plan. Any comments/opinions on the above would be helpful. If you are working on anything overlapping let me know.

maknapp:
This is a mostly functional serializer with some parts completely untested. I did change a few tests because they seemed like mistakes to me.

I think deprecating in v6 is the way to go here. The public surface area of the current serializer methods is too large and won't allow for minor breaking changes to be made with the new serializer (like changing exception types). For example, I tried changing the existing `SerializeToString` methods to use the new `TryParse` methods, but many `SerializeToString` methods differ how they throw exceptions vs returning `null`.

Some of the `SerializeToString` methods are copied directly into the `TryParse` or converter methods just to be 100% sure not to break anything with the old serializer. Some `TryParse` methods are different to test reducing allocations.

Some tests are changed to test only the new `TryParse` methods instead of continuing to test the corresponding `SerializeToString` methods - this was just me being lazy for now. The latest commit is testing out how to test both serializers at the same time.

Type mapping is simplified. The new serializer matches only by property name. A `CalendarPropertyConverter` is expected to handle all VALUE types for its property. `Activator` is not used at all, so this should avoid any issues mentioned in #977 too.

Undefined properties use the default converter, which assumes all values are strings even if VALUE is defined. This is the same behavior as the old serializer.

Backslash encoding is applied only to TEXT value types. Most/all(?) other value types specifically say no other encoding (backslash) applies.

There is no encoding option. It currently writes without BOM by default, same as old serializer. Do we need an option to include BOM?

### Todo
- Custom exception type with properties like `LineNumber` to point directly to the error
- Ensure all name/property/value types contain strictly valid characters as specified in the RFC
- Option or callback to control how exceptions are handled
- Remove logging
- Items listed in #973
- Improve copied parsing methods
- Improve test coverage
- Make sure all `TryParse` methods have good parameter names and will never throw exceptions
- Check if `OnDeserializing` and `OnDeserialized` events are required
- Mark all old serializer methods as obsolete
- `SerializeAsync`
- `PipeReader` and `PipeWriter` (might make `SerializeAsync` easier too)

<details>
<summary>Benchmarks</summary>

| Method                      | Runtime            | Mean       | Gen0    | Gen1   | Allocated |
|---------------------------- |------------------- |-----------:|--------:|-------:|----------:|
| Deserialize                 | .NET 10.0          |  44.281 us | 11.0474 | 1.6479 |  90.52 KB |
| Deserialize                 | .NET 8.0           |  51.969 us | 11.3525 | 1.2817 |  92.94 KB |
| Deserialize                 | .NET Framework 4.8 | 125.079 us | 16.8457 | 2.4414 |  104.2 KB |
|                             |                    |            |         |        |           |
| Deserialize2                | .NET 10.0          |  21.894 us |  6.9580 | 0.9460 |  56.96 KB |
| Deserialize2                | .NET 8.0           |  27.703 us |  7.0190 | 0.9460 |  57.35 KB |
| Deserialize2                | .NET Framework 4.8 |  70.102 us | 10.6201 | 1.4648 |  65.82 KB |
|                             |                    |            |         |        |           |
| BenchmarkSerializeCalendar  | .NET 10.0          |  10.407 us |  3.1433 | 0.0610 |  25.76 KB |
| BenchmarkSerializeCalendar  | .NET 8.0           |  12.658 us |  3.1891 | 0.0610 |  26.14 KB |
| BenchmarkSerializeCalendar  | .NET Framework 4.8 |  26.722 us |  5.2185 | 0.0916 |   32.2 KB |
|                             |                    |            |         |        |           |
| BenchmarkSerializeCalendar2 | .NET 10.0          |   3.187 us |  0.6638 |      - |   5.42 KB |
| BenchmarkSerializeCalendar2 | .NET 8.0           |   4.422 us |  0.6866 |      - |   5.66 KB |
| BenchmarkSerializeCalendar2 | .NET Framework 4.8 |  11.868 us |  1.0986 |      - |   6.77 KB |
|                             |                    |            |         |        |           |
| SerializeMultibyte          | .NET 10.0          |  12.816 us |  4.4708 | 0.1068 |  36.57 KB |
| SerializeMultibyte          | .NET 8.0           |  14.896 us |  4.5013 | 0.1068 |  36.82 KB |
| SerializeMultibyte          | .NET Framework 4.8 |  27.128 us |  7.2327 | 0.1831 |  44.62 KB |
|                             |                    |            |         |        |           |
| SerializeMultibyte2         | .NET 10.0          |   3.524 us |  0.6104 |      - |      5 KB |
| SerializeMultibyte2         | .NET 8.0           |   4.544 us |  0.6256 |      - |   5.14 KB |
| SerializeMultibyte2         | .NET Framework 4.8 |  10.519 us |  0.9308 |      - |   5.79 KB |
|                             |                    |            |         |        |           |
| DeserializeCalendar         | .NET 10.0          |  14.748 us |  3.9215 | 0.1984 |  32.11 KB |
| DeserializeCalendar         | .NET 8.0           |  19.620 us |  3.9978 | 0.1831 |  32.82 KB |
| DeserializeCalendar         | .NET Framework 4.8 |  47.291 us |  6.0425 | 0.3052 |  37.24 KB |
|                             |                    |            |         |        |           |
| DeserializeCalendar2        | .NET 10.0          |   5.820 us |  1.9379 | 0.0610 |  15.88 KB |
| DeserializeCalendar2        | .NET 8.0           |   8.423 us |  1.9531 | 0.0610 |  16.01 KB |
| DeserializeCalendar2        | .NET Framework 4.8 |  18.050 us |  3.0518 |      - |  18.85 KB |
</details>

Fixes #973, #978, #702, #937
Partial #908
Partial #977


maknapp:
> Failed DeserializesSingleEventCalendar [8 ms]
  Error Message:
     Assert.That(output, Is.EqualTo(ics))
  Expected string length 867 but was 903. Strings differ at index 15.
  Expected: "BEGIN:VCALENDAR\nVERSION:2.0\nPRODID:-//Mozilla.org/NONSGML Moz..."
  But was:  "BEGIN:VCALENDAR\r\nVERSION:2.0\r\nPRODID:-//Mozilla.org/NONSGML M..."

Not sure why the test failed... expected should be using `\r\n` line endings.

axunonb:
> Not sure why the test failed... expected should be using `\r\n` line endings.

`IcsFiles.Attachment4` reads the raw embedded .ics file with StreamReader.ReadToEnd(), so the string retains whatever line endings are actually stored in the embedded resource file — in this case, I think the Attachment4.ics file on disk has LF-only line endings.

The fix is either
`var ics = IcsFiles.Attachment4.ReplaceLineEndings("\r\n");`
or to save the file with CR LF line ending.

We experienced this issue in the past already. Is there something wrong with the line ending definition for the repo?
[edit]
Yes, see #984 that should fix the issue

axunonb:
> There is no encoding option. It currently writes without BOM by default, same as old serializer. Do we need an option to include BOM?

If I remember correctly the BOM option had compatibility reasons. The BOM is clearly excluded in RFC, so I vote for not allowing uncompliant output. A related question might arise with line endings, which are also well defined as CR LF, not LF-only. Different output is up to users.

PR #984, opened 2026-09-13
chore: Force CRLF line endings / no BOM for .ics test fixtures per RFC 5545

axunonb:
Fixes the  `.gitattributes` / `core.autocrlf` normalization issue

Run ` git add --renormalize -- .` locally after checkout of the branch

Commits with `*.ics` files to test project are checked for "no BOM".


maknapp:
Would it be better for ics files be excluded from all auto-formatting so that testing bad line endings and BOM is possible? I don't see any tests specifically for deserializing with BOM, so removing this one Language2.ics file removes the only test for it.

axunonb:
Good point - I'd lean the other way here. Rather than excluding `.ics` files from formatting/normalization (which is fragile: a contributor using a plain text editor, or a future rebase/checkout, could still silently strip a BOM we're relying on), I think it's more robust to:

1. Remove the BOM from `Language2.ics` so all fixtures stay consistent and protected by the existing formatting/BOM-check gate.
2. Add a dedicated, self-contained unit test for BOM handling that doesn't depend on a fixture file's byte-for-byte preservation at all. E.g.:

```csharp
[Test]
public void Deserialize_WithUtf8Bom_ParsesSuccessfully()
{
    var icsText = "BEGIN:VCALENDAR\r\nVERSION:2.0\r\n...\r\nEND:VCALENDAR\r\n";
    var bom = new byte[] { 0xEF, 0xBB, 0xBF };
    var bytesWithBom = bom.Concat(Encoding.UTF8.GetBytes(icsText)).ToArray();

    using var stream = new MemoryStream(bytesWithBom);
    var calendar = Calendar.Load(stream);

    Assert.That(calendar, Is.Not.Null);
}
```

This way:
- The BOM is generated in-memory at test time, so it can never be accidentally "fixed" by an editor, formatter, or Git checkout - the test is fully self-documenting and immune to file-encoding drift.
- We keep BOM detection/rejection or -tolerance coverage without needing special-cased fixture files or `.gitattributes` exceptions.
- All `.ics` fixtures remain subject to the same formatting rules and the `EnsureIcsFixturesWithoutBOM.sh` gate, so accidental BOM contamination elsewhere is still caught.

If there's a reason it needs to come from an actual file on disk (e.g., testing a specific file-loading code path rather than `Calendar.Load(Stream)`), we could instead write the BOM-prefixed bytes to a temp file at test-time rather than committing a permanently BOM-prefixed fixture to the repo. The test cases "with BOM" are very rare anyway.

maknapp:
Yes, that sounds like a better plan. Do you want to add a BOM deserializing test to this PR?

axunonb:
> Yes, that sounds like a better plan. Do you want to add a BOM deserializing test to this PR?

Done.

Ical.Net.Tests/SimpleDeserializationTests.cs
```diff
@@ -432,9 +432,9 @@ public void EventStatus()
     }
 
     [Test, Category("Deserialization")]
-    public void Language4()
+    public void Language1()
     {
-        var iCal = SimpleDeserializer.Default.Deserialize(new StringReader(IcsFiles.Language4)).Cast<Calendar>().Single();
+        var iCal = SimpleDeserializer.Default.Deserialize(new StringReader(IcsFiles.Language3)).Cast<Calendar>().Single();
```
maknapp: Is this supposed to be using `IcsFiles.Language1`?

axunonb: Indeed. Now corrected

maknapp (approved):

